PR #16, opened 2020-12-04
CI: Add a GitHub Action to build on Windows

Thog:
This build in any commits and pull requests.


Thog:
For some reasons the job failed with some weird error huh

@WerWolv would be nice to restart it

WerWolv:
Thanks a lot :)
I restarted it but it still seems to fail with the same error, do you know what's up with that?

PR #40, opened 2020-12-06
cmake: add install target

umarcor:
Ref #31.

This PR adds an install target so that `make install` works.


Blackoo429:
How do I start

PR #47, opened 2020-12-06
Add a list of packages to install for building on Fedora 33

Calinou:
Tested in a fresh Fedora 33 Docker container.


WerWolv:
Hi, thanks a lot!
In the future when more distros are going to get added, I think it would be better to move all of this to a separate document.
Merging this for now though

PR #53, opened 2020-12-07
create subdir 'dist', update compiling guidelines

umarcor:
This PR moves the all in one lists of dependencies to scripts that users can execute. The README is updated accordingly.


WerWolv:
Thanks a lot :)

PR #57, opened 2020-12-07
Fix segmentation fault when searching for an empty hex string

### LINKED ISSUE #56: CTRL + F, CTRL + G don't work in "Hex Editor", crash after entering an empty string in Search->Hex
Hello. I found several bugs in ImHex.
1. CTRL + F, CTRL + G don't work in "Hex Editor"
2. If open Edit->Search->Hex and enter (null string) - ImHex crashed
GIF 07 12 2020 17-06-23

Sven-C:
Close #56.

This PR fixes the crash caused by a segfault when searching for an empty hex string. The crash gets fixed by appending a null byte to the bytes to search for. When searching for an empty hex string, the currently highlighted sequences get deselected, similar to how string searching behaves.

### DIFF
--- a/source/views/view_hexeditor.cpp
+++ b/source/views/view_hexeditor.cpp
@@ -775,7 +775,8 @@ R"(
             string = "0" + string;
 
         std::vector<u8> hex;
-        hex.reserve(string.size() / 2);
+        hex.reserve(string.size() / 2 + 1);
+        hex[hex.size()-1] = '\0';
 
         for (u32 i = 0; i < string.size(); i += 2) {
             char byte[3] = { string[i], string[i + 1], 0 };


WerWolv:
Hi thanks for the PR
However, this doesn't actually fix the issue properly. Increasing the size of the hex vector by 1 causes it to not be zero sized anymore which later on causes `hex[0]` to not segfault anymore. It would be better to simply check if `string` is empty and in that case bail out instead of searching in the first place.

Sven-C:
I have thought of that as well, but just returning an empty list as a result of the search will cause existing highlights to persist, which is inconsistent behavior with the string searching functionality (where searching an empty string causes all existing highlights to disappear). I'll look a bit more into it.

IAMTH3 (approved):

PR #80, opened 2020-12-09
Build ImHex on MacOS with xmake 

dodola:
1. Auto install build deps
2. No need other build config

### DIFF
--- a/source/views/view_hexeditor.cpp
+++ b/source/views/view_hexeditor.cpp
@@ -493,7 +493,7 @@ namespace hex {
             return false;
 
         fseek(file, 0, SEEK_END);
-        size_t size = ftello64(file);
+        size_t size = ftello(file);
         rewind(file);
 
         data.resize(size);
--- a/source/views/view_strings.cpp
+++ b/source/views/view_strings.cpp
@@ -4,10 +4,12 @@
 #include "helpers/utils.hpp"
 
 #include <cstring>
-
+#if !defined(__GNUC__)
 #include <llvm/Demangle/Demangle.h>
+#endif
 
 using namespace std::literals::string_literals;
+extern "C" char* __cxa_demangle(const char*, char*, size_t*, int*);
 
 namespace hex {
 
@@ -37,7 +39,11 @@ namespace hex {
             }
             ImGui::Separator();
             if (ImGui::MenuItem("Demangle")) {
+                #if !defined(__GNUC__)
                 this->m_demangledName = llvm::demangle(this->m_selectedString);
+                #else
+                this->m_demangledName  = __cxa_demangle(this->m_selectedString.c_str(), nullptr, nullptr, nullptr);
+                #endif
                 if (!this->m_demangledName.empty())
                     View::doLater([]{ ImGui::OpenPopup("Demangled Name"); });
             }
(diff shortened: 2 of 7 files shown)


source/views/view_hexeditor.cpp
```diff
@@ -493,7 +493,7 @@ namespace hex {
             return false;
 
         fseek(file, 0, SEEK_END);
-        size_t size = ftello64(file);
+        size_t size = ftello(file);
```
WerWolv, line 496: This breaks proper reading of huge files on other platforms

source/views/view_strings.cpp
```diff
@@ -4,10 +4,12 @@
 #include "helpers/utils.hpp"
 
 #include <cstring>
-
+#if !defined(__GNUC__)
 #include <llvm/Demangle/Demangle.h>
+#endif
 
 using namespace std::literals::string_literals;
+extern "C" char* __cxa_demangle(const char*, char*, size_t*, int*);
```
WerWolv, line 12: `__cxa_demangle` and `llvm::Demangle` are not the same thing! llvm supports both Itanium and MSVC name mangling while `__cxa_demangle` only supports Itanium. This doesn't fix the whole problematic, it just removes it.

source/views/view_strings.cpp
```diff
@@ -37,7 +39,11 @@ namespace hex {
             }
             ImGui::Separator();
             if (ImGui::MenuItem("Demangle")) {
+                #if !defined(__GNUC__)
                 this->m_demangledName = llvm::demangle(this->m_selectedString);
+                #else
+                this->m_demangledName  = __cxa_demangle(this->m_selectedString.c_str(), nullptr, nullptr, nullptr);
```
WerWolv, line 45: Same as above

source/views/view_tools.cpp
```diff
@@ -7,8 +7,11 @@
 #include "providers/provider.hpp"
 #include "helpers/utils.hpp"
 
+#if !defined(__GNUC__)
 #include <llvm/Demangle/Demangle.h>
+#endif
 
+extern "C" char* __cxa_demangle(const char*, char*, size_t*, int*);
```
WerWolv, line 14: Same as above

source/views/view_tools.cpp
```diff
@@ -76,7 +79,11 @@ namespace hex {
     void ViewTools::drawDemangler() {
         if (ImGui::CollapsingHeader("Itanium/MSVC demangler")) {
             if (ImGui::InputText("Mangled name", this->m_mangledBuffer, 0xF'FFFF)) {
+                #if !defined(__GNUC__)
                 this->m_demangledName = llvm::demangle(this->m_mangledBuffer);
+                #else
+                this->m_demangledName  = __cxa_demangle(this->m_mangledBuffer, nullptr, nullptr, nullptr);
```
WerWolv, line 85: Same as above

xmake.lua
```diff
@@ -0,0 +1,76 @@
+
```
WerWolv, line 1: Is there any reason to use another build system than cmake? cmake works totally fine on MacOS as well so I don't see the need to use another one which I manually have to update as well, every time I make changes to the build process

Issue #37, opened 2020-12-05
libLLVMDemangle.so


Mailaender:
Is it possible to make the dependency optional? I am trying to package this for @openSUSE and I can't find the dependency. As I ran into many broken links for other distributions, I think it may be deprecated.

WerWolv:
Hi!
It's absolutely not deprecated, in fact it's still being further improved as seen here https://github.com/llvm/llvm-project/tree/master/llvm/include/llvm/Demangle.

ImHex in itself does have a lot of dependencies already. And I know not all of them can be available on absolutely every distribution. Therefore, if I start making one dependency optional, lots of other people will come asking for capstone or python or libmagic to be made optional. I feel like ImHex is going to lose lots of what it is when having different features available on different architectures because prebuilt libraries aren't available there. In my opinion you should rather try to get these libraries be available on those distros as well instead of ripping them out of projects. That will help other people too that use these libraries.

Sijisu:
I have exactly the same issue as @Mailaender (I'm on openSUSE as well), can't find the dependency anywhere.

Mailaender:
I understand. https://build.opensuse.org/package/show/devel:tools:compiler/llvm somehow has it disabled somehow.

It isn't packaged for https://www.archlinux.org/packages/extra/x86_64/llvm-libs/ as you linked in the ReadMe and https://www.archlinux.org/packages/extra/x86_64/llvm/ only contains headers and a statically linked library. Is it even available for @ArchLinux?

Mailaender:
We believe it is part of libLLVM itself already.

WerWolv:
```
➜  / find /usr | grep libLLVM
...
/usr/lib/libLLVMDemangle.a
```

Definitely available on my Arch Linux setup.
I have `llvm 11.0.0-1` and `llvm-libs 11.0.0-1` installed.

WerWolv:
Oh I just noticed the title, sorry. Me writing libLLVMDemangle.so in the cmake script was actually an error. It was fixed today to use the correct file. It's very much possible that that's the reason nothing came up for you. Sorry about that

Mailaender:
Even with https://github.com/WerWolv/ImHex/pull/35 applied it can't find the library. 

https://build.opensuse.org/package/view_file/devel:tools:compiler/llvm11/llvm-do-not-install-static-libraries.patch is probably the culprit here so nothing on your end.

Mailaender:
[...]

PR #82, opened 2020-12-10
Add a Gentoo ebuild

abrasive:
Gentoo doesn't have a community package set like the AUR, so it'd be handy to make this available here if you like.

### DIFF
--- a/dist/ImHex-9999.ebuild
+++ b/dist/ImHex-9999.ebuild
@@ -1,0 +1,34 @@
+# app-editors/ImHex
+# Copyright 2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+DESCRIPTION="A hex editor for reverse engineers, programmers, and eyesight"
+HOMEPAGE="https://github.com/WerWolv/ImHex"
+SRC_URI=""
+EGIT_REPO_URI="https://github.com/WerWolv/ImHex.git"
+
+PYTHON_COMPAT=( python3_{6,7,8,9} )
+
+inherit git-r3 python-single-r1 cmake
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64"
+IUSE=""
+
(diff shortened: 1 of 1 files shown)


tuxuser:
In case you are interested in integrating the stable release too: https://github.com/WerWolv/ImHex/issues/37#issuecomment-740230248

abrasive:
Ah, your ebuild is much nicer, @tuxuser :-) I'll come back to this when I have some time.

Thog:
It should be noted that this libLLVM issue should be fixed @tuxuser as of https://github.com/WerWolv/ImHex/commit/9b853350f1cfe16f0ff8265c7fbddedbc44bbd94

VPaulV:
Hi Guys! Trying to build imhex-1.17 with https://github.com/WerWolv/ImHex/issues/37#issuecomment-740230248 ebuild. But getting an error 
```
CMake Error at cmake/build_helpers.cmake:280 (message):
  External dependency
  /var/tmp/portage/app-editors/imhex-1.17.0/work/ImHex-1.17.0/lib/external/capstone
  is empty!

  Make sure to correctly clone ImHex using the --recurse-submodules git
  option or initialize the submodules manually.
  ```
  I have capstone already installed into the system, is there anyway to disable this check?

WerWolv:
@VPaulV I have no idea about Gentoo but the fix is to properly clone the repository using `git clone https://github.com/WerWolv/ImHex --recurse-submodules`. If you just disable the check, ImHex will fail to build completely. Capstone isn't the only dependency that will be missing, it's just displaying the first one it found that's missing.

VPaulV:
@WerWolv  The thing that I already have all dependencies installed into the system, but build ignores -DUSE_SYSTEM_CAPSTONE=1

WerWolv:
@VPaulV You don't have all dependencies installed.

If you don't use --recurse-submodules, you'll also be missing the pattern language, libromfs and a few others. These won't be on any package manager you're using.

PR #83, opened 2020-12-10
Add support for macOS and cxxabi demangler to decouple from LLVM

mikr:
First commit:
On macOS ImHex must be build with g++-10 which clashes with the homebrew llvm build that uses libc++ intead of libstdc++.
I added a cxxabi demangler so LLVM is no longer needed to demangle but LLVM demangling can still be used by `cmake -DLLVM_DEMANGLE=1` anyway if needed.

Second commit:
The remaining compiler fixes for macOS, added cmake directives to find the necessary Homebrew libraries and set the rpath for the dynamic linking of the Python libraries.

I have built and run the program successfully on macOS 10.14, macOS 10.15 and Ubuntu 20.04.1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@ pkg_search_module(GLFW REQUIRED glfw3)
 pkg_search_module(GLM REQUIRED glm)
 pkg_search_module(CRYPTO REQUIRED libcrypto)
 pkg_search_module(CAPSTONE REQUIRED capstone)
+if (APPLE)
+    pkg_search_module(OPENSSL REQUIRED OpenSSL)
+endif (APPLE)
+
 if (NOT LLVM_DEMANGLE)
     include(CheckCXXSymbolExists)
     check_cxx_symbol_exists(abi::__cxa_demangle "cxxabi.h" HAVE_CXXABI)
@@ -23,6 +27,14 @@ if (LLVM_DEMANGLE OR NOT HAVE_CXXABI)
 endif (LLVM_DEMANGLE OR NOT HAVE_CXXABI)
 find_package(nlohmann_json REQUIRED)
 find_package(Python COMPONENTS Interpreter Development)
+if (APPLE)
+    find_package(Python3 REQUIRED)
+    find_package(glfw3 REQUIRED)
+    find_package(OpenSSL REQUIRED)
+    find_library(CAPSTONE_LIBRARY capstone)
+    find_library(MAGIC_LIBRARY magic)
+endif (APPLE)
+
 
 if(Python_VERSION LESS 3)
     message(STATUS ${PYTHON_VERSION_MAJOR_MINOR})
@@ -37,6 +49,10 @@ if (LLVM_FOUND)
 endif (LLVM_FOUND)
 
 include_directories(include ${GLFW_INCLUDE_DIRS} ${CAPSTONE_INCLUDE_DIRS} libs/ImGui/include libs/glad/include ${Python_INCLUDE_DIRS})
+if (APPLE)
+    include_directories(include ${OPENSSL_INCLUDE_DIRS} ${CAPSTONE_INCLUDE_DIRS})
+endif (APPLE)
+
 
 # Get Python major and minor
 string(REPLACE "." ";" PYTHON_VERSION_MAJOR_MINOR ${Python_VERSION})
@@ -54,6 +70,11 @@ SET(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DRELEASE")
 SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
 
 add_definitions(-DLLVM_DEMANGLE=${LLVM_DEMANGLE})
+if (APPLE)
+    set(VAR "Python_LINK_OPTIONS" CACHE STRING "LINKER:-rpath,/Library/Developer/CommandLineTools/Library/Frameworks")
+    add_link_options(${Python_LINK_OPTIONS})
+endif (APPLE)
+
 add_executable(ImHex
         source/main.cpp
         source/window.cpp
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -12,20 +12,47 @@ pkg_search_module(GLFW REQUIRED glfw3)
 pkg_search_module(GLM REQUIRED glm)
 pkg_search_module(CRYPTO REQUIRED libcrypto)
 pkg_search_module(CAPSTONE REQUIRED capstone)
+if (APPLE)
+    pkg_search_module(OPENSSL REQUIRED OpenSSL)
```
filcab, line 16: This is probably required on all platforms *but* we hit the lucky path where this library is in the same place as other libraries and Linux/Windows end up not noticing the lack of searching for openssl.

CMakeLists.txt
```diff
@@ -12,20 +12,47 @@ pkg_search_module(GLFW REQUIRED glfw3)
...
 find_package(OpenGL REQUIRED)
-find_package(LLVM REQUIRED CONFIG)
+if (LLVM_DEMANGLE OR NOT HAVE_CXXABI)
+    find_package(LLVM REQUIRED CONFIG)
+endif (LLVM_DEMANGLE OR NOT HAVE_CXXABI)
 find_package(nlohmann_json REQUIRED)
 find_package(Python COMPONENTS Interpreter Development)
+if (APPLE)
+    find_package(Python3 REQUIRED)
+    find_package(glfw3 REQUIRED)
```
filcab, line 32: Why do we need to search for the libraries again? On my system, the `pkg_search_module` calls are enough

CMakeLists.txt
```diff
@@ -12,20 +12,47 @@ pkg_search_module(GLFW REQUIRED glfw3)
...
 endif()
 
-llvm_map_components_to_libnames(_llvm_libs demangle)
-llvm_expand_dependencies(llvm_libs ${_llvm_libs})
 
-include_directories(include ${GLFW_INCLUDE_DIRS} ${CAPSTONE_INCLUDE_DIRS} ${LLVM_INCLUDE_DIRS} libs/ImGui/include libs/glad/include ${Python_INCLUDE_DIRS})
+if (LLVM_FOUND)
+    llvm_map_components_to_libnames(_llvm_libs demangle)
+    llvm_expand_dependencies(llvm_libs ${_llvm_libs})
+    include_directories(include ${LLVM_INCLUDE_DIRS})
```
filcab, line 48: no need to have `include` here, IIUC.

CMakeLists.txt
```diff
@@ -12,20 +12,47 @@ pkg_search_module(GLFW REQUIRED glfw3)
...
+endif (LLVM_FOUND)
+
+include_directories(include ${GLFW_INCLUDE_DIRS} ${CAPSTONE_INCLUDE_DIRS} libs/ImGui/include libs/glad/include ${Python_INCLUDE_DIRS})
+if (APPLE)
+    include_directories(include ${OPENSSL_INCLUDE_DIRS} ${CAPSTONE_INCLUDE_DIRS})
```
filcab, line 53: Just like before, this should be needed on all systems, but we end up not noticing we don't have these directories on other platforms because all includes are in the same dirs.

CMakeLists.txt
```diff
@@ -88,14 +121,20 @@ add_executable(ImHex
         resource.rc
         )
 
-target_link_directories(ImHex PRIVATE ${LLVM_LIBRARY_DIR})
+if (LLVM_FOUND)
+    target_link_directories(ImHex PRIVATE ${LLVM_LIBRARY_DIR})
+endif (LLVM_FOUND)
+
+if (APPLE)
+    target_link_libraries(ImHex PRIVATE ${GLFW_LIBRARIES} ${MAGIC_LIBRARY} ${OPENSSL_CRYPTO_LIBRARY} ${CAPSTONE_LIBRARY} ${Python_LIBRARIES})
```
filcab, line 129: ah cool. This is probably the best way to do it in a cross-platform way, but in order to be able to merge the `target_link_libraries` calls we'd need to know if Windows would still use static libs. Good idea.

source/helpers/utils.cpp
```diff
@@ -3,6 +3,13 @@
 #include <cstdio>
 #include <codecvt>
 #include <locale>
+#include <iostream>
+
+#if !(LLVM_DEMANGLE + 0) && (defined(__clang__) || defined(__GNUG__))
```
filcab, line 8: Probably best to fix this definition in cmake (just set the define to 0 or 1), so we don't need to convert to numeric in the preprocessor.

filcab: We can also define an easy to read variable/preprocessor constant so we don't need to repeat this weird condition again later.

source/lang/evaluator.cpp
```diff
@@ -189,7 +189,7 @@ namespace hex::lang {
             }
             else {
                 members.push_back(pattern);
-                unionSize = std::max(memberSize, unionSize);
+                unionSize = std::max(memberSize, (u64)unionSize);
```
filcab, line 192: Best to fix `memberSize`'s definition to be `size_t` given that it only gets values from `size_t` things.

source/helpers/utils.cpp
```diff
@@ -90,4 +97,23 @@ namespace hex {
         return result;
     }
 
+#if !(LLVM_DEMANGLE + 0) && (defined(__clang__) || defined(__GNUG__))
+    std::string demangle(const std::string &mangled_name) {
+        int status = 0;
+        int skip_underscore = mangled_name.find("__") == 0;
```
filcab, line 103: Shouldn't we also make sure there's a `Z` just after the `_`?

filcab (commented):
Looks cool. I ended up working on bringing it up on macOS too, but figured I should comment on this PR with a few suggestions. Using libcxxabi instead of llvm is a good idea given the issues with the C++ standard libraries. Won't scale if we end up getting demanglers for non-native platforms (e.g: demangle Itanium symbols on Windows, or Windows symbols on Titanium), but right now it's a good thing and gets us a working ImHex!

Thanks!

Thog:
Closing as of https://github.com/WerWolv/ImHex/commit/310059f2740ccc85333dc77380993631a5d91b49

PR #85, opened 2020-12-10
Proper DPI scaling and basic custom font

averne:
For now the font is loaded from a hardcoded path, but in the future it would be better to ask the user directly (in a settings panel), as well as expose some kind of customization for the ui/font scaling.

~~Also some stuff scales badly because of hardcoded window sizes, which need to be weighted by the scaling factor.~~ EDIT: I fixed the issues I saw.


averne:
@ocornut It appears that on @WerWolv's setup these changes (especially the first commit) make fonts a bit blurry, as opposed to previously with the bitmap font:


Full window on [low dpi](https://media.discordapp.net/attachments/478162551381819392/786860511802818580/unknown.png) (windows), and on [high dpi][https://media.discordapp.net/attachments/573472192587759626/786867375701229568/Screenshot_20201211_084905.png) (linux/xorg).

ocornut:
Io.FontGlobalScale should never used. You should scale the font when Adding it to the font atlas and rebuild the font atlas.

averne:
That's what I'm doing [here](https://github.com/WerWolv/ImHex/pull/85/files#diff-90da325c23da2f8860d4d6c79af12527fd2d0cb909e172edf84daf2e2d1c9724R86-R109). If we have a custom font, reset FontGlobalScale to 1.0 and build the atlas scaled at the right DPI.

ocornut:
Hmm... Your screenshot above shows the default font but your code seems to not be performing that scaling on the default font? I am missing something?

Also you need to round the resulting scaled size for now, as not rounding it currently can result to misalignment in shape rendering.

averne:
The screenshots compare the bitmap font with no rescaling (low dpi) with a ttf font scaled for 15 pixels height.
I don't rebuild the atlas fot the bitmap font yet indeed, I'll fix that later.

ocornut:
Oh I see. I need then to clarify that for most fonts its difficult to get a sharp output at ~14-18 pixels with stb_truetype at it doesn’t support hinting (and we don’t have subpixel rendering either).
When you get toward 4K-suitable font sizes eg >20-30 it starts to be not a problem.

If you switch to using Freetype + misc/imgui_freetype.cpp you can get better small size output at it supports hinting.

averne:
Ah I see, thanks a lot for your help.

filcab:
Hi all!

Neat project! Can I just ask how I would be able to configure the size of the font being used?
After this commit, the text becomes *huge* (see pics) on a mac (I know it's not supported yet, I have a few patches which I'll upload soon).
I did see there's a reference to a font.ttf, but I don't think there's one in the repo. Am I missing something?

Thank you,
 Filipe

Pics:
Same configuration/resolution/etc. I'm screenshotting the same area of the screen (top left and go right up to the first menuitem)

Before:
Screenshot 2020-12-12 at 20 35 33

After:
Screenshot 2020-12-12 at 20 39 37

averne:
Hi, this is probably because on Retina displays the framebuffer size is different from the window size. Can you try playing with this? https://github.com/ocornut/imgui/blob/master/imgui.h#L1725 (ImGui mentions using 2, 2).

filcab:
@averne 
I've tried that, but it didn't work too well.
Changing font scale "works" but the UI is still drawn as if the font is large. I *can*, though, hack it like this:
```
diff --git a/source/window.cpp b/source/window.cpp
index f30f1ee..ed99407 100644
--- a/source/window.cpp
+++ b/source/window.cpp
@@ -223,7 +223,8 @@ namespace hex {
             glfwGetMonitorContentScale(monitor, &xscale, &yscale);
 
             // In case the horizontal and vertical scale are different, fall b
ack on the average
-            this->m_globalScale = this->m_fontScale = std::midpoint(xscale, ys
cale);
+            //this->m_globalScale = this->m_fontScale = std::midpoint(xscale, 
yscale);
+            this->m_globalScale = 1.0;
```

This is... obviously wrong. But I'm not familiar with the codebase, maybe someone else can figure out what the correct fix is with this info.

Thank you,
 Filipe

averne:
What values do `glfwGetWindowSize` and `glwfGetFramebufferSize` report?

filcab:
At the end of `static bool ImGui_ImplGlfw_Init(GLFWwindow* window, bool install_callbacks, GlfwClientApi client_api)`:
```
init: glfwGetWindowSize: w: 1280, h:720
init: glfwGetFramebufferSize: w: 2560, h:1440
```

averne:
So DisplayFramebufferScale does nothing? That's odd. Maybe it's being reset back somewhere later?

filcab:
I have this patch, printing `2 2` every frame:
```
diff --git a/libs/ImGui/source/imgui_impl_glfw.cpp b/libs/ImGui/source/imgui_impl_glfw.cpp
diff --git a/libs/ImGui/source/imgui_impl_glfw.cpp b/libs/ImGui/source/imgui_impl_glfw.cpp
index b2fb427..457f7c9 100644
--- a/libs/ImGui/source/imgui_impl_glfw.cpp
+++ b/libs/ImGui/source/imgui_impl_glfw.cpp
@@ -45,6 +45,8 @@
 #include "imgui.h"
 #include "imgui_impl_glfw.h"
 
+#include <cstdio>
+
 // GLFW
 #include <GLFW/glfw3.h>
 #ifdef _WIN32
@@ -477,6 +479,7 @@ void ImGui_ImplGlfw_NewFrame()
     io.DisplaySize = ImVec2((float)w, (float)h);
     if (w > 0 && h > 0)
         io.DisplayFramebufferScale = ImVec2((float)display_w / w, (float)display_h / h);
+    printf("%g %g\n", io.DisplayFramebufferScale[0], io.DisplayFramebufferScale[1]);
     if (g_WantUpdateMonitors)
         ImGui_ImplGlfw_UpdateMonitors();
```

*with* the bug

filcab:
Before you ask, I've added more dumps:
```
w: 1440, h: 855, display_w: 2880, display_h: 1710
```

adriweb:
Any more news? This is making it almost unusable on retina Macs :(
Let me know if more testing/feedback if needed, I can help.

WerWolv:
I'm really depending on PRs for this one. I have no way myself to test or even compile anything for macOS

PR #112, opened 2021-01-04
m1 build

krocki:
builds using provided instructions on Apple M1 macbook pro

### DIFF
--- a/plugins/libimhex/include/helpers/utils.hpp
+++ b/plugins/libimhex/include/helpers/utils.hpp
@@ -23,12 +23,6 @@
     #define ftello64 ftell
 #endif
 
-template<>
-struct std::is_integral<u128> : public std::true_type { };
-template<>
-struct std::is_integral<s128> : public std::true_type { };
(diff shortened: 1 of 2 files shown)


WerWolv:
Hi! The unused ranges include is already gone now. However the four other lines are really required for is_integral to properly pick up s128 and u128. You can't just remove them.
If you're still interested in this PR, please fix building by moving these lines to an appropriate place instead of removing them. 

I don't have a mac so the only real way for me to test compilation there is with the GitHub Actions which right now takes up to 3 hours sometimes to start...

WerWolv:
More information about this, it builds absolutely fine with clang 11.0.0 on Arch Linux. I'm really not sure why it should be failing on Mac

WerWolv:
The issue seems to have been that libc++ already defines is_integral and is_signed for 128 bit types. Only including them when not using it fixed the issue.

krocki:
thanks, I'm going to give it another try

PR #120, opened 2021-01-08
fix(lang): fixes occurrence index in findSequence built-in function

kaero:
Hi! A little fix to make `findSequence` works as described :)


WerWolv:
Oops, thank you :)

PR #126, opened 2021-01-10
Fix build on FreeBSD

nmantani:
Hello,

This pull request fixes build error on FreeBSD. The added lines are conditional and they do not affect build on other operating systems.

On FreeBSD, libcrypto is installed by default but pkg-config related files are not installed.
So it is necessary to use find_library instead of pkg_search_module in CMakeLists.txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,11 @@ set(Python_FIND_FRAMEWORK NEVER)
 
 # Find packages
 find_package(PkgConfig REQUIRED)
-pkg_search_module(CRYPTO REQUIRED libcrypto)
+if (CMAKE_SYSTEM_NAME MATCHES FreeBSD)
+    find_library(CRYPTO crypto REQUIRED)
+else()
+    pkg_search_module(CRYPTO REQUIRED libcrypto)
+endif()
 pkg_search_module(CAPSTONE REQUIRED capstone)
 find_package(OpenGL REQUIRED)
 find_package(nlohmann_json REQUIRED)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,10 +23,9 @@ set(Python_FIND_FRAMEWORK NEVER)
 
 # Find packages
 find_package(PkgConfig REQUIRED)
-if (CMAKE_SYSTEM_NAME MATCHES FreeBSD)
+pkg_search_module(CRYPTO libcrypto)
+if(NOT CRYPTO_FOUND)
     find_library(CRYPTO crypto REQUIRED)
-else()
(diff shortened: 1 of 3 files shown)


WerWolv:
Thanks for the PR!

CMakeLists.txt
```diff
@@ -23,7 +23,11 @@ set(Python_FIND_FRAMEWORK NEVER)
 
 # Find packages
 find_package(PkgConfig REQUIRED)
-pkg_search_module(CRYPTO REQUIRED libcrypto)
+if (CMAKE_SYSTEM_NAME MATCHES FreeBSD)
```
WerWolv, line 26: Could this maybe be made more generic so it checks if it can find something with pkgconfig and if it doesn't, it uses find_library?

source/views/view_hexeditor.cpp
```diff
@@ -13,6 +13,10 @@
 #undef __STRICT_ANSI__
 #include <cstdio>
 
+#if defined(__FreeBSD__)
```
WerWolv, line 16: I have a similar block of code in `/plugins/libimhex/helpers/utils.hpp` already for apple devices. Those two could maybe be combined

CMakeLists.txt
nmantani: Thank you for reviewing! I updated CMakeLists.txt to use a more generic way.

https://github.com/WerWolv/ImHex/pull/126/commits/e058dd80ac5cdc67af703ebc9b25463f47e26a0c

source/views/view_hexeditor.cpp
nmantani: Yes, combining these macros is preferable.  However, source files under ```/source``` directory include .hpp files in ```/include``` directory and source files under ```/plugins/libimhex/source``` directory include .hpp files in ```/plugins/libimhex/include``` directory. So it is not possible to combine them without drastically changing this structure.

WerWolv: libimhex is also included in the main application, it acts as an interface between the two. So just adding it there and including utils.hpp in view_hexeditor.cpp should be alright

nmantani: Thank you for explanation. I have integrated the macro in ```source/views/view_hexeditor.cpp``` into ```/plugins/libimhex/helpers/utils.hpp``` and build is successful on FreeBSD.

https://github.com/WerWolv/ImHex/pull/126/commits/da2c8805fb0afe481b64d30b37ccb8350d878a3c

WerWolv: Perfect, thank you!

PR #132, opened 2021-01-10
Show commit on about page only if available

george-hopkins:
If the _.git_ folder is not present during the build, the about page will contain an uninformative line because `GIT_BRANCH` and `GIT_COMMIT_HASH` are empty. This may happen if the source code is downloaded from the releases page or is exported with `git archive`.


WerWolv:
Thank you for the PR :)

PR #138, opened 2021-01-18
Enable ImGui keyboard navigation

csinkers:
Small change to enable the built-in imgui keyboard navigation for things like moving around with the arrow keys, opening menus with alt etc.


WerWolv:
Thanks a lot for the PR!

PR #158, opened 2021-02-14
windows installation script

clovistricon:


bvernoux:
I confirm this sh script work fine win mingw64 (just need to launch it with `sh winbuilder.sh`)
- Build tested with success with Windows7 SP1 Pro/Entreprise and Windows 10
- Note with Windows7 SP1 Pro/Entreprise there is an issue to launch ImHex see Windows7 SP1 Pro/Entreprise

clovistricon:
I will check it out.
Also, should I add a shebang and some error handling then ?

bvernoux:
I'm not sure MSYS2/Mingw64 can manage correctly shebang like on real GNU/Linux system it is why I launched it with sh explicitely which is not a big problem.

clovistricon:
I looked with the commands ldd and strings and it seems that the usp10 lib wasn't linked at all. I think it has to be either usp10.dll/Usp10.dll or libusp10.a.
I put this one like in this link: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=28493

bvernoux:
I have tried your latest modification and there is the same issue with Windows7 Pro/Entreprise when I try to execute imhex.exe:
image

PR #179, opened 2021-02-26
dont dlclose nullptrs in plugins

ydnax:
ImHex segfaulted for me at startup on nixos. adding the check fixed it for me


WerWolv:
Thanks!

PR #183, opened 2021-02-28
Install files in their proper paths on linux

roblabla:
This makes sure everything goes in their proper FHS paths on linux, while (hopefully) keeping the current paths on windows and macos.


plugins/libimhex/CMakeLists.txt
```diff
@@ -56,6 +55,7 @@ endif ()
 
 add_library(libimhex SHARED ${LIBIMHEX_SOURCES})
 
+target_include_directories(libimhex PUBLIC include ${XDGPP_INCLUDE_DIRS})
```
WerWolv, line 58: `${XDGPP_INCLUDE_DIRS}` should be added to the line below instead of having two `target_include_directories` directives

plugins/libimhex/source/helpers/utils.cpp
```diff
@@ -208,27 +209,47 @@ namespace hex {
...
+
+            configDirs.insert(configDirs.begin(), xdg::ConfigHomeDir());
+            dataDirs.insert(dataDirs.begin(), xdg::DataHomeDir());
+
+            std::vector<std::string> result;
 
             switch (path) {
                 case ImHexPath::Patterns:
-                    return { std::string(homeDirectory) + "/.imhex/patterns", IMHEX_INSTALL_PREFIX "/share/imhex/patterns" };
+                    transform(dataDirs.begin(), dataDirs.end(), back_inserter(result),
```
WerWolv, line 222: This should be `std::transform` and `std::back_inserter`. Where's the `using namespace std` leak coming from? xdg?

plugins/libimhex/source/helpers/utils.cpp
```diff
@@ -13,6 +13,7 @@
     #include <unistd.h>
```
WerWolv, line 13: Those three includes can be removed now. They were only there for the old home directory detection

WerWolv (changes requested):

plugins/libimhex/source/helpers/utils.cpp
roblabla: got no clue, I checked the xdgpp.hpp file out, it doesn't contain using namespace std.

roblabla:
All fixed. Dunno where the `using namespace std;` comes from, but that doesn't seem to be related to this PR?

WerWolv:
Yeah I also can't find it anywhere. 
Fixes look good though. Thank you :)

PR #185, opened 2021-03-01
Fix new linux file locations in README

roblabla:
Just realized I forgot to update the README with the fixed location of configuration files.


WerWolv:
Hah didn't notice either. Thank you

PR #186, opened 2021-03-01
Fix various crashes on bad settings data.

roblabla:
getSetting now returns a straight nlohmann::json instead of an optional.
If the data isn't present, it will return a json null.

All accesses to the settings will first check that the data has the
expected type.

### DIFF
--- a/plugins/builtin/source/content/settings_entries.cpp
+++ b/plugins/builtin/source/content/settings_entries.cpp
@@ -5,7 +5,7 @@ namespace hex::plugin::builtin {
     void registerSettings() {
 
         ContentRegistry::Settings::add("hex.builtin.setting.interface", "hex.builtin.setting.interface.color", 0, [](auto name, nlohmann::json &setting) {
-            static int selection = setting;
+            static int selection = setting.is_number() ? static_cast<int>(setting) : 0;
 
             const char* themes[] = {
                     "hex.builtin.setting.interface.color.dark"_lang,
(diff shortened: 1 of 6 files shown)


plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -5,7 +5,7 @@ namespace hex::plugin::builtin {
     void registerSettings() {
 
         ContentRegistry::Settings::add("hex.builtin.setting.interface", "hex.builtin.setting.interface.color", 0, [](auto name, nlohmann::json &setting) {
-            static int selection = setting;
+            static int selection = setting.is_null() ? 0 : static_cast<int>(setting);
```
WerWolv, line 8: It's kinda awkward having to do this check and hardcode the default value here and in the function call before for every Setting you add. Instead it might be better to just store the default value together with the rest of the registered setting and have a function `ContentRegistry::Settings::getInteger(setting)` that does that and returns the default value from before if the settings value has an incorrect type

WerWolv: Another issue also is that if you change the setting manually and you put a string there instead of an int, it will still fail because it's not null but it's also not an int

roblabla: I at least fixed the setting to be an int. For adding the default value, I'm not comfortable enough with the code or C++ to do this properly.

adriweb:
I have a similar problem on a fresh build, where the settings don't exist (or the file is an empty json object).
It just crashes at launch : `[json.exception.type_error.302] type must be number, but is null`.

Debugging that lead me to view_hexeditor.cpp: 

```cpp
        EventManager::subscribe<EventSettingsChanged>(this, [this] {
            auto alpha = ContentRegistry::Settings::getSetting("hex.builtin.setting.interface", "hex.builtin.setting.interface.highlight_alpha");
```

Initially from `EventManager::post<EventSettingsChanged>();`

PR #187, opened 2021-03-01
Fix passing path through argv

roblabla:
This is necessary to fix the bad any_cast error that occurs otherwise.


WerWolv:
Lol wtf thanks

Issue #171, opened 2021-02-22
Can not build in ubuntu 18 with Anaconda installed


alex-bender:
```
~ cmake -DCMAKE_BUILD_TYPE=Release ..
-- Checking for one of the modules 'libmagic'
-- Configuring done
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
/home/alex/src/ImHex/plugins/libimhex/MBEDTLS_INCLUDE_DIR
   used as include directory in directory /home/alex/src/ImHex/plugins/libimhex

CMake Warning at CMakeLists.txt:37 (add_executable):
  Cannot generate a safe runtime search path for target imhex because files
  in some directories may conflict with libraries in implicit directories:

    runtime library [libfreetype.so.6] in /usr/lib/x86_64-linux-gnu may be hidden by files in:
      /home/alex/anaconda3/lib

  Some of these libraries may not be found correctly.


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.

```
How can I troubleshoot this better?

theKidOfArcrania:
First, I suggest you check the `dist/get_deps_debian.sh` file that installs all the dependencies (this file is used in the automatic workflows), and make sure you have all of them installed. In this case if you look at the first error, that is actually just having the lack of libmbedtls-dev installed. If you look at the second note, that's just a warning, so it has nothing to do with anaconda

dragonxtek:
I can't compile it in Ubuntu 20.10.
After running ./get_deps_debian.sh I get the following:
`cmake -DCMAKE_BUILD_TYPE=Release ..

-- Checking for one of the modules 'libmagic'
CMake Error: File /tmp/ImHex/external/yara/yara/libyara/yara.pc.in does not exist.
CMake Error at external/yara/CMakeLists.txt:124 (configure_file):
  configure_file Problem configuring file


CMake Error at plugins/libimhex/CMakeLists.txt:8 (add_subdirectory):
  The source directory

    /tmp/ImHex/external/nativefiledialog

  does not contain a CMakeLists.txt file.


CMake Error at plugins/libimhex/CMakeLists.txt:9 (add_subdirectory):
  The source directory

    /tmp/ImHex/external/fmt

  does not contain a CMakeLists.txt file.


-- Configuring incomplete, errors occurred!
See also "/tmp/ImHex/build/CMakeFiles/CMakeOutput.log".

`

[...]

Issue #113, opened 2021-01-05
Failed to build on Ubuntu 20.04


Corfucinas:
```
└─> …/ImHex/build on  master via 喝 v3.16.3 
at 11:27:46 [💣 127] ✗  cmake -DCMAKE_BUILD_TYPE=Release ..
-- Checking for one of the modules 'libmagic'
CMake Error at CMakeLists.txt:47 (list):
  list index: 2 out of range (-2, 1)


-- Configuring incomplete, errors occurred!
See also "/home/the-box/Github/ImHex/build/CMakeFiles/CMakeOutput.log".
```

File is attached, thanks for taking a look! 

(I ran the bash script under build before running cmake)

[CMakeOutput.log](https://github.com/WerWolv/ImHex/files/5767944/CMakeOutput.log)

cyang812:
I met the same issue. plz let me know how to fix it. Thanks

aeroc7:
I just commented out line 47 and it builds successfully, though this should by no means be a permanent solution.

Corfucinas:
I tried that now I'm having this issue

```
In file included from /home/the-box/Github/ImHex/plugins/libimhex/source/helpers/utils.cpp:1:
/home/the-box/Github/ImHex/plugins/libimhex/include/helpers/utils.hpp:62:10: fatal error: concepts: No such file or directory
   62 | #include <concepts>
      |          ^~~~~~~~~~
compilation terminated.
make[2]: *** [plugins/libimhex/CMakeFiles/libimhex.dir/build.make:76: plugins/libimhex/CMakeFiles/libimhex.dir/source/helpers/utils.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:215: plugins/libimhex/CMakeFiles/libimhex.dir/all] Error 2
make: *** [Makefile:130: all] Error 2

```

aeroc7:
> 
> 
> I tried that now I'm having this issue
> 
> ```
> In file included from /home/the-box/Github/ImHex/plugins/libimhex/source/helpers/utils.cpp:1:
> /home/the-box/Github/ImHex/plugins/libimhex/include/helpers/utils.hpp:62:10: fatal error: concepts: No such file or directory
>    62 | #include <concepts>
>       |          ^~~~~~~~~~
> compilation terminated.
[...]

PR #191, opened 2021-03-02
Update deps in README

theKidOfArcrania:
This change is to reflect the commit made in 785ecb8a7858edfb88747623b7437da0c02b5317, which changes openssl libcrypto to libmbedtls, making changes everywhere except in the README documentation.

This seems to be the root of the issue in #171 and #113 (see the very last comment made)

May also be nice to update the CMake files to make a more friendlier error message (not added here)


WerWolv:
Thanks!

The CMake scripts are more of a "Hey it works" than a nice implementation since I'm not that experienced with them. If you have any suggestions, I'd be happy to improve them though

PR #193, opened 2021-03-03
[Plugin/Lang] Small corrections to de_DE

tumGER:
As is tradition whenever you localize your projects.

With <3,

Tom


plugins/builtin/source/lang/de_DE.cpp
```diff
@@ -52,7 +52,7 @@ namespace hex::plugin::builtin {
                 { "hex.common.address", "Adresse" },
                 { "hex.common.size", "Länge" },
                 { "hex.common.region", "Region" },
-                { "hex.common.match_selection", "Benutz selektion" },
+                { "hex.common.match_selection", "Benutzselektion" },
```
WerWolv, line 55: This looks wrong but the original wasn't great either. It's a toggle for some tools that makes the memory region the tool works with be synchronized with whatever the user selected in the hex editor

tumGER: Arbeitsbereich/Benutzungsbereich synchronisieren maybe?

WerWolv: Arbeitsbereich synchronisieren sounds nice

PR #196, opened 2021-03-04
Added Italian (IT) translation

CrustySean:


plugins/builtin/source/lang/it_IT.cpp
```diff
@@ -0,0 +1,475 @@
+#include <hex/plugin.hpp>
+
+namespace hex::plugin::builtin {
+
+    void registerLanguageItIT() {
+        ContentRegistry::Language::registerLanguage("Italian (IT)", "it-IT");
```
WerWolv, line 6: This should probably be `Italian (Italy)` or just `Italian` alone. Are there different versions?

plugins/builtin/source/lang/it_IT.cpp
```diff
@@ -0,0 +1,475 @@
...
+                { "hex.view.hashes.name", "Hash" },
+                    { "hex.view.hashes.settings", "Impostazioni" },
+                    { "hex.view.hashes.function", "Funzioni di Hash" },
+                    { "hex.view.hashes.iv", "Valore Iniziale" },
+                    { "hex.view.hashes.poly", "Polinomio" },
+                    { "hex.view.hashes.result", "Risultato" },
+
+                { "hex.view.help.name", "Aiuto" },
+                    { "hex.view.help.about.name", "Riguardo ImHex" },
+                        { "hex.view.help.about.translator", "Tradotto da WerWolv" },
```
WerWolv, line 128: It was translated by you, not by me :P

plugins/builtin/source/lang/it_IT.cpp
```diff
@@ -0,0 +1,475 @@
+#include <hex/plugin.hpp>
+
+namespace hex::plugin::builtin {
+
+    void registerLanguageItIT() {
```
WerWolv, line 5: This function needs to be forward declared and called in `plugin_builtin.cpp`, otherwise the language won't be added

WerWolv (changes requested):

PR #199, opened 2021-03-08
Fix path to plugins in readme.

roblabla:
Typing is hard.


WerWolv:
Reviewing is hard.
Thank you!

PR #215, opened 2021-04-03
Fix: ignore non-existent magic directories

ThisALV:
Many of the default directories searched for mgc files do not actually exists into some systems (e.g. `/usr/share/ubuntu/imhex/`), which is setting error to non-zero error code so mgc files end up ignored.


WerWolv:
Thank you!

PR #225, opened 2021-04-20
Fix: patterns popup

ThisALV:
This PR fixes two minor issues:
- Available .hexpat list not cleaned between each scan
- .hexpat list popup not showing up first time it is opened, which is problematic for usage of command-line argument / recent files to open file at startup

As a side effect, pattern editor view is now dismissed while .hexpat list is open, but it show up again as soon as popup has been closed.

### DIFF
--- a/source/views/view_pattern.cpp
+++ b/source/views/view_pattern.cpp
@@ -225,7 +225,7 @@ namespace hex {
     }
 
     void ViewPattern::drawContent() {
-        if (ImGui::Begin(View::toWindowName("hex.view.pattern.name").c_str(), &this->getWindowOpenState(), ImGuiWindowFlags_None | ImGuiWindowFlags_NoCollapse | ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoScrollWithMouse)) {
+        if (ImGui::Begin(View::toWindowName("hex.view.pattern.name").c_str(), &this->getWindowOpenState(), ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoScrollWithMouse)) {
             auto provider = SharedData::currentProvider;
 
             if (provider != nullptr && provider->isAvailable()) {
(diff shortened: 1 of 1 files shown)


source/views/view_pattern.cpp
```diff
@@ -144,6 +144,7 @@ namespace hex {
             preprocessor.addDefaultPragmaHandlers();
 
 
+            m_possiblePatternFiles.clear();
```
WerWolv, line 147: Use `this->` to match style used elsewhere

source/views/view_pattern.cpp
```diff
@@ -224,22 +225,23 @@ namespace hex {
...
-                this->m_textEditor.Render("hex.view.pattern.name"_lang, textEditorSize, true);
-
-                auto consoleSize = ImGui::GetContentRegionAvail();
-                consoleSize.y -= ImGui::GetTextLineHeightWithSpacing();
-
-                ImGui::PushStyleColor(ImGuiCol_ChildBg, this->m_textEditor.GetPalette()[u32(TextEditor::PaletteIndex::Background)]);
-                if (ImGui::BeginChild("##console", consoleSize, true, ImGuiWindowFlags_AlwaysVerticalScrollbar)) {
-                    for (auto &[level, message] : this->m_console) {
-                        switch (level) {
+        if (!ImGui::IsPopupOpen("hex.view.pattern.accept_pattern"_lang)) {
```
WerWolv, line 228: I'm not sure I understand what you're doing here. This will cause the pattern source window to disappear completely while the "accept pattern" popup is shown.

WerWolv (changes requested):
Thanks a lot for the PR, I didn't even realize it was broken

source/views/view_pattern.cpp
ThisALV: Yes, and if it was open, it will show up again as soon as "Accept pattern" popup is gone.
It's the only clear way I found to address that issue, and as "Accept pattern" was a modal popup anyways, I don't think it should matter.

ThisALV:
Hi. As this PR didn't have any activity for a while, I was wondering if it is okay and if it can be merged into master.

PR #232, opened 2021-05-13
Fix typo in Debian/Ubuntu dependencies script

DrivebyCoding:
The "!" parameter to if needs to be free-standing. As is the script always outputs this on Ubuntu 20.04:

./get_deps_debian.sh: 8: !which: not found


WerWolv:
Thanks!

PR #241, opened 2021-05-23
Fix yara system lib detection and add support for fmt and curl

Mailaender:
This is a followup of https://github.com/WerWolv/ImHex/pull/210.


plugins/libimhex/CMakeLists.txt
```diff
@@ -13,14 +13,24 @@ else()
 endif()
 
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../../external/nativefiledialog ${CMAKE_CURRENT_BINARY_DIR}/external/nativefiledialog EXCLUDE_FROM_ALL)
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../../external/fmt ${CMAKE_CURRENT_BINARY_DIR}/external/fmt)
+if(USE_SYSTEM_FMT)
```
WerWolv, line 16: Some places use `if (NOT USE_SYSTEM_XXX)` and some use `if (USE_SYSTEM_XXX)` and have the if and else body swapped. This should be a bit more unified

WerWolv (changes requested):

WerWolv:
Thank you!

plugins/libimhex/CMakeLists.txt
Mailaender: Updated.

PR #255, opened 2021-06-16
Setting an env var for macOS 10.15 support

fosterbrereton:
CMake [seems to have trouble](https://stackoverflow.com/questions/34208360/cmake-seems-to-ignore-cmake-osx-deployment-target) setting the minimum macOS target unless it's done just right. This change sets an environment variable so it will get picked up directly, instead of through cmake.


WerWolv:
Thank you! :)

PR #258, opened 2021-06-17
Add rpath to binary executable on macOS

fosterbrereton:
The loading of `builtin.hexplug` was causing the runtime also load `libimhex.dylib` as specified by the path `@rpath/libimhex.dylib`. The macOS runtime to looks up the array of rpaths (runtime paths) within the application binary itself in an attempt to discover and load the `libimhex.dylib`. Unfortunately the binary had no rpaths specified, and the runtime would either find no `libimhex.dylib` (causing the plugin to fail to load) or a _totally different_ `libimhex.dylib` than the one inside the application bundle (causing ODR violations and all kinds of other strange behaviors.)

The fix here is to inject an rpath into the imhex binary that causes `@rpath/libimhex.dylib` to resolve to the `libimhex.dylib` embedded within the application bundle. This will cause the runtime to realize that dylib is already loaded, and the `builtin.hexplug` is then able to load correctly.

Note that recent macOS versions now include Gatekeeper, which may block the loading of dylibs to applications that do not have appropriate permissions. As a result, users may have to grant `imhex.app` full disk access in order to load plugins:

imhex_access

If many users hit this issue, a note should be added to the README.

There are a couple other minor changes that come along with this PR, hopefully none of which are too surprising.

### DIFF
--- a/source/helpers/plugin_manager.cpp
+++ b/source/helpers/plugin_manager.cpp
@@ -15,8 +15,10 @@ namespace hex {
     Plugin::Plugin(std::string_view path) {
         this->m_handle = dlopen(path.data(), RTLD_LAZY);
 
-        if (this->m_handle == nullptr)
+        if (this->m_handle == nullptr) {
+            fprintf(stderr, "dlopen failed: %s", dlerror());
             return;
+        }
 
         auto pluginName = fs::path(path).stem().string();
 
--- a/source/helpers/plugin_manager.cpp
+++ b/source/helpers/plugin_manager.cpp
@@ -16,7 +16,7 @@ namespace hex {
         this->m_handle = dlopen(path.data(), RTLD_LAZY);
 
         if (this->m_handle == nullptr) {
-            fprintf(stderr, "dlopen failed: %s", dlerror());
+            hex::log::error("dlopen failed: {}", dlerror());
             return;
         }
 
--- a/source/window.cpp
+++ b/source/window.cpp
@@ -100,8 +100,12 @@ namespace hex {
             {
                 auto language = ContentRegistry::Settings::getSetting("hex.builtin.setting.interface", "hex.builtin.setting.interface.language");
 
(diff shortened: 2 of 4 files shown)


source/helpers/plugin_manager.cpp
```diff
@@ -15,8 +15,10 @@ namespace hex {
     Plugin::Plugin(std::string_view path) {
         this->m_handle = dlopen(path.data(), RTLD_LAZY);
 
-        if (this->m_handle == nullptr)
+        if (this->m_handle == nullptr) {
+            fprintf(stderr, "dlopen failed: %s", dlerror());
```
WerWolv, line 19: Instead of using fprintf, you can use `hex::log::error` with libfmt syntax

source/window.cpp
```diff
@@ -100,8 +100,12 @@ namespace hex {
             {
                 auto language = ContentRegistry::Settings::getSetting("hex.builtin.setting.interface", "hex.builtin.setting.interface.language");
 
-                if (language.is_string())
+                if (language.is_string()) {
                     LangEntry::loadLanguage(static_cast<std::string>(language));
+                } else {
+                    // If no language is specified, fall back to English.
+                    LangEntry::loadLanguage("en-US");
```
WerWolv, line 107: That's a step in the right direction, we can go with that for now. 
The real fix would be to reset the setting back to a sensible default if it has the wrong type or an invalid value

WerWolv (changes requested):
Thank you so much!

source/helpers/plugin_manager.cpp
fosterbrereton: Fixed.

source/window.cpp
fosterbrereton: Noted, thanks.

WerWolv:
Thank you for the changes!

PR #280, opened 2021-08-05
Installation procedure on Ubuntu 20.04

packmad:


WerWolv:
Thank you! This procedure should no longer be needed now with the fixed releases packages / flatpak package though

PR #299, opened 2021-09-15
warnings: fix format-security warnings

### LINKED ISSUE #245: Failed to compile with `clang++`
Trying to compile ImHex 1.8 using `clang++`. It fails with the error: `/var/tmp/pamac-build-kibouo/imhex/src/imhex/source/views/view_hexeditor.cpp:305:32: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]`. (See attached log for complete output.)

Info: 
- [AUR package 1.8.1-1](https://aur.archlinux.org/packages/imhex),
- `clang++`: 
image

[compile.log](https://github.com/WerWolv/ImHex/files/6546632/compile.log)

russkel:
Manjaro and presumably other systems have started adding the `-Werror=format-security` flag, causing the build to fail.

It might make sense to add the `-Werror=format-security` flag to the compiler flags given how easy it is to keep it happy.

Fixes #245 and #295

### DIFF
--- a/source/views/view_help.cpp
+++ b/source/views/view_help.cpp
@@ -17,7 +17,7 @@ namespace hex {
 
 
     static void drawTitle(const std::string &title) {
-        ImGui::TextColored(ImVec4(0.6F, 0.6F, 1.0F, 1.0F), title.c_str());
+        ImGui::TextColored(ImVec4(0.6F, 0.6F, 1.0F, 1.0F), "%s", title.c_str());
     }
 
     static void drawBuiltinFunction(
@@ -27,9 +27,9 @@ namespace hex {
         const std::string &description
     ) {
         ImGui::Bullet();
-        ImGui::TextColored(ImVec4(0.3F, 0.7F, 0.2F, 1.0F), return_type.c_str());
+        ImGui::TextColored(ImVec4(0.3F, 0.7F, 0.2F, 1.0F), "%s", return_type.c_str());
         ImGui::SameLine();
-        ImGui::TextColored(ImVec4(0.57F, 0.24F, 0.69F, 1.0F), name.c_str());
+        ImGui::TextColored(ImVec4(0.57F, 0.24F, 0.69F, 1.0F), "%s", name.c_str());
         ImGui::SameLine();
         ImGui::TextColored(ImVec4(0.71F, 0.19F, 0.31F, 1.0F), "(");
         ImGui::SameLine();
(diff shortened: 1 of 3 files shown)


WerWolv:
Thank you!
I assumed the issue was caused by the `%llx` vs. `%lx` warnings I've given up on fixing instead of this.

source/views/view_help.cpp
```diff
@@ -27,15 +27,15 @@ namespace hex {
...
-        ImGui::TextColored(ImVec4(0.57F, 0.24F, 0.69F, 1.0F), name.c_str());
+        ImGui::TextColored(ImVec4(0.57F, 0.24F, 0.69F, 1.0F), "%s", name.c_str());
         ImGui::SameLine();
-        ImGui::TextColored(ImVec4(0.71F, 0.19F, 0.31F, 1.0F), "(");
+        ImGui::TextColored(ImVec4(0.71F, 0.19F, 0.31F, 1.0F), "%s", "(");
         ImGui::SameLine();
         ImGui::Text("%s", arguments.c_str());
         ImGui::SameLine();
-        ImGui::TextColored(ImVec4(0.71F, 0.19F, 0.31F, 1.0F), ")");
+        ImGui::TextColored(ImVec4(0.71F, 0.19F, 0.31F, 1.0F), "%s", ")");
```
WerWolv, line 38: Are these two lines really giving off warnings too?

WerWolv: I meant the ones with only `"("` and  `")"` as the string. Mobile might have messed up something sorry

russkel: Oh sorry, it doesn't look like it actually. I can revert those lines.

WerWolv: Great thanks :)

russkel:
I don't think "%s" is working for the `LangEntry` lines.

The warning is correct:
```
/home/russ/aur/imhex/src/imhex/source/views/view_help.cpp: In member function ‘void hex::ViewHelp::drawAboutPopup()’:
/home/russ/aur/imhex/src/imhex/source/views/view_help.cpp:80:34: warning: format ‘%s’ expects argument of type ‘char*’, but argument 2 has type ‘hex::LangEntry’ [-Wformat=]
   80 |             ImGui::TextWrapped("%s", "hex.view.help.about.thanks"_lang);
      |                                 ~^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                  |   |
      |                                  |   hex::LangEntry
      |                                  char*
```

WerWolv:
I love C generics... You need to explicitly static_cast the value to a const char* for the conversation operator to be called correctly

russkel:
> I love C generics... You need to explicitly static_cast the value to a const char* for the conversation operator to be called correctly

That worked, but damn it's ugly. Should I fix up the others and push changes?

WerWolv:
Yes please.
Yeah I don't think there's really a better way besides wrapping the ImGui calls into proper C++ variadic functions. Either way it's going to be ugly

russkel:
As master isn't compiling for me, I cherry-picked the patch onto `v1.9.0` (https://github.com/russkel/ImHex/tree/v1.9.0) and compiled it on Manjaro. Seems to be working well.

PR #303, opened 2021-09-20
i18n: Chinese(Simplified)

xtexChooser:
As the title


source/window/win_window.cpp
```diff
@@ -162,7 +162,9 @@
 
         void Window::initNative() {
             auto hConsoleWindow = ::GetConsoleWindow();
+            #if !defined(DEBUG)
             ::ShowWindow(hConsoleWindow, FALSE);
```
WerWolv, line 166: This was deliberately done this way to prevent the console window from showing up by default but still output logs if it was opened from the console.

I find the console window to be in the way in almost all cases so this allows only showing it when actually needed.

cmake/build_helpers.cmake
```diff
@@ -185,7 +185,7 @@ macro(createPackage)
             UNRESOLVED_DEPENDENCIES_VAR _u_deps
             CONFLICTING_DEPENDENCIES_PREFIX _c_deps
             DIRECTORIES ${DEP_FOLDERS}
-            POST_EXCLUDE_REGEXES ".*system32/.*\\.dll"
+            POST_EXCLUDE_REGEXES ".*system32/.*\\.dll" ".*SysWow64/.*\\.dll"
```
WerWolv, line 188: What's the reason this had to be added?

WerWolv (changes requested):
Thanks for the PR!
I'm more than happy to merge the Chinese translation you made, however I left some comments about the other things you added

source/window/win_window.cpp
xtexChooser: No, I tried to open ImHex.exe from cmd and cmd closed

cmake/build_helpers.cmake
xtexChooser: Reverted

source/window/win_window.cpp
xtexChooser: Reverted

WerWolv:
Perfect thank you :)

WerWolv (approved):

PR #308, opened 2021-09-23
build: Fix system libraries usage

KokaKiwi:
When compiling using system libraries, the name used for linking are all wrong and result in lots of link errors.
Thus, this PR fixing the lib names used for linking by using the CMake variables when available or correct lib names.


WerWolv:
Hi! Thanks a lot for the PR!

PR #317, opened 2021-10-08
sys: Fix macOS compilation (as of 2dc1886)

Kuruyia:
This PR fixes the compilation of ImHex on macOS.
It doesn't add any QoL improvements about the state of the software on this operating system, but at least it'll lessen the work needed to get ImHex running on macOS.

I tried to modify the code in a way that would only affect macOS builds (using `#if defined(OS_MACOS)` guards, although some changes could affect other platforms.
The main change that could behave in such a way is the removal of uses of `std::views::reverse` in favor of the older, more supported reverse iterators.
This was done because LLVM 13.0.0 (release 4 days ago as of now) is the first LLVM version to support the `ranges` standard library, but its support is incomplete. Support for `std::views::reverse` is, however, already in the source of LLVM and will probably be available in one of the next releases.
I honestly think merging this PR can wait for an LLVM update, as macOS is not officially supported by ImHex anyway.

The external cURL was also updated for its support of mbedtls 3.0, which is now the latest version available on Homebrew.

Tests were done on an M1 Mac, but I believe x86 will work just fine as well.

Thanks for this software :)

### DIFF
--- a/plugins/builtin/source/content/pl_builtin_functions.cpp
+++ b/plugins/builtin/source/content/pl_builtin_functions.cpp
@@ -194,7 +194,7 @@ namespace hex::plugin::builtin {
                 auto index = Token::literalToSigned(params[1]);
 
 #if defined(OS_MACOS)
-                const auto signIndex = index >> (sizeof(index) - 1);
+                const auto signIndex = index >> (sizeof(index) * 8 - 1);
                 const auto absIndex = (index ^ signIndex) - signIndex;
 #else
                 const auto absIndex = std::abs(index);


Kuruyia:
Mmhm the macOS CI doesn't seem to pull LLVM 13 from Homebrew, which is needed for `ranges`.
Homebrew might be out of date on their machine, although I wonder if we can just `brew update` in the Workflow, before installing the deps?

Kuruyia:
Oh, indeed it seems to compile without those includes, I thought they were more dependent on it.
Let's see if it doesn't break anything!

WerWolv:
Hm it can't seem to find libGL now: `ld: library not found for -lGL`
Is that called differently on MacOS or is a package missing?

Kuruyia:
OK, it is (finally) working!
The artifact launches fine, but strings are not localized, behavior that I didn't have on a development build. I'll try to fix that next.

WerWolv:
That would be fine by me.
I don't really know how the bundles on MacOS work though. If you have any improvements for the current system, feel free to change things

WerWolv:
Cool thanks a lot :)

On other platforms, all resources are being checked for in a folder directly next to the executable followed by a folder in some common configuration/resources location. (AppData on Windows, $HOME/.local on Linux). I think this should be the case on MacOS too so you don't have to copy stuff into the app bundle but instead just put the files in e.g the Application Support folder.

Standalone plugins will need some more work anyways before being usable. Most importantly, libimhex needs to move to its own repo first. I'll take care of that when we get there.

Thanks a lot for all your work so far. Let me know when you think you're ready for merging or have any questions

WerWolv:
Thanks a lot for all the work!

Issue #320, opened 2021-10-09
[Bug] Wrong CRC results


raron:
### Operating System

Linux

### What's the issue you encountered?

The results of CRC32 and CRC16 calculation in the window View -> Hashes are wrong.

### How can the issue be reproduced?

Write a pattern with a known CRC into a file. The Zlib's [
A Painless Guide to CRC Error Detection Algorithms](https://zlib.net/crc_v3.txt) uses "123456789" (without quotation mark, without newline).

    echo -ne "123456789" > /tmp/crc

Open the file in ImHex, open Hashes view, at Region check Match Selection, select all the file (9 bytes), select CRC32 as Hash function, use FFFFFFFF as Initial value, use 04C11DB7 as Polynomial.

The ImHex show F99AEA5C as Result when it should be cbf43926 (or at least 649C2FD3 or FC891918 or 1898913F if input / output is reflected).

Used ImHex version: 2176988.

### ImHex Version

1.10.1

### Additional context?

The Zlib's CRC guide specify the CBF43926 as a good value:

    Name   : "CRC-32"
    Width  : 32
    Poly   : 04C11DB7
    Init   : FFFFFFFF
    RefIn  : True
    RefOut : True
    XorOut : FFFFFFFF
    Check  : CBF43926

where the Check field "contains the checksum obtained when the ASCII string "123456789" is fed through the specified algorithm".

This result even match with the online CRC calculator at:
http://www.sunshine2k.de/coding/javascript/crc/crc_js.html

In the ImHex there are no settings for input / output reflection and final xor value. (The final xor value seems to be implicitly hard-coded to be 0xffffffff at `plugins/libimhex/source/helpers/crypto.cpp:110` where the bitwise not of the result is returned.)

Note: the result form ImHex does not match with results from this online calculator with any input / output reflection settings.

Others sources produce the same result, too:

<details>
The crc32 command on linux systems gives the results CBF43926

    $ crc32 /tmp/v
    cbf43926

which does match with the CHECK value from the Zlib's CRC guide

This result is match with the result from `boost::crc`:

    #include <boost/crc.hpp>
    #include <iostream>
    
    int main()
    {
        const char a[] = "123456789";
        boost::crc_optimal<32, 0x4C11DB7, 0xffffffff, 0xffffffff, true, true> crc;
        crc.process_block(std::begin(a), std::end(a)-1);
        std::cout << std::hex << crc.checksum() << std::endl;
    
        return EXIT_SUCCESS;
    }

[...]

PR #321, opened 2021-10-10
Fix CRC and hash calculations

raron:
The results from CRC calculations (and in some cases hash calculcations) were wrong, as I wrote in issue #320. (Now I suspect it was a reflected algorithm, but I could not get any matching result from any other source anyway.)

This pachset fixes the CRC and the hash calculations and cleans up the hash calculations a bit, but it adds the Boost CRC header only library as a (build) dependency, if you not mind.

There were multiple issues:
First to multiple methods (offset/size, begin/end) was used to represent the range of the data to be hashed, because of this (sometimes) the CRC / hash has been calculated from the wrong (not the selected) data.

Second: it seems at the event `EventRegionSelected` the `size` parameter of the `Range` is smaller by one than the really selected range, I could not find the cause of this.  
Edit: it seem it only less-by-one if the event is generated by selecting a region by mouse, if I click onto a field in Pattern Data, the calue of the range.size is correct. (So the results will be wrong in that case.)

Third: some CRCs uses additional parameters: xorout, reflectIn, reflectOut (these reverses the bit patters and used, because in some situations it is / was easier to implement in hardware in that way). The Boost CRC library supports these parameters, so it was easy to add them to the UI, too.

The check for `(this->m_hashRegion[1] >= this->m_hashRegion[0])` was removed, because it does not make any sense if the `m_hashRegion[1]` is interpreted as size parameter. (I hope I did not mess up the identation too much.)

Added CRC8, the Boost CRC supports it, so why not.

Changed the switch on m_currHashFunction to based on enum values, so in the future it should be easier to add more hashes or reorder them.

### DIFF
--- a/plugins/libimhex/source/helpers/crypto.cpp
+++ b/plugins/libimhex/source/helpers/crypto.cpp
@@ -17,6 +17,9 @@
 #include <span>
 #include <concepts>
 #include <functional>
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
 
 #if MBEDTLS_VERSION_MAJOR <= 2
 
@@ -45,8 +48,8 @@ namespace hex::crypt {
     void processDataByChunks(prv::Provider* data, u64 offset, size_t size, Func func)
     {
         std::array<u8, 512> buffer = { 0 };
-        for (u64 bufferOffset = 0; bufferOffset < size; bufferOffset += buffer.size()) {
-            const u64 readSize = std::min(buffer.size(), size - bufferOffset);
+        for (size_t bufferOffset = 0; bufferOffset < size; bufferOffset += buffer.size()) {
+            const auto readSize = std::min(buffer.size(), size - bufferOffset);
             data->read(offset + bufferOffset, buffer.data(), readSize);
             func(buffer.data(), readSize);
         }


WerWolv:
Hi, Thanks a lot for your PR!
Your changes all look really good, however I'm not sure how fond I am of adding a dependency on boost. Even if it's just referencing the CRC library, it still pulls in a bunch of other boost libraries which feels overkill for what it's used. Instead, figuring out what exactly is wrong with my CRC functions is I think the way to go (doesn't mean you have to do that).

WerWolv:
When I think of Boost I just think of an incredibly powerful but also incredibly bloated library where each part depends on 50 others. The build time of ImHex already isn't great and I'm worried adding another big library like that just for one feature is going to make it even worse. I don't know if this is the case here too though.

I'd love to have support for all the options you've added here.

WerWolv:
> I think even if the algorithm uses the reflected implementation, in the user interface the non-reflected parameters should be entered
Yeah absolutely. That makes the most sense

> What would you prefer, should I make new commit on this branch / PR (to remove a boost dependency and change algorithm back) or close this and create a new branch / PR from master?
Feel free to commit more changes to this PR here. Shouldn't cause any issues right now.

> (I want to check the 16 and 8 bit variants, too, and maybe create some unit-test with a bunch of randomly generated known-to-be-good parameter + data + result triplets.)

Unit tests right now are kinda only made specifically to test the pattern language. We'd probably have to split the `tests` subproject up into multiple different parts then where one is purely for the pattern language and the other(s) are for things like the hash functions or other algorithms.

WerWolv:
Latest commit refactors the whole test structure a little bit to allow for crc and other tests like that using Catch2.

raron:
Hello,

I have changed the algorithm back to more-or-or-less original one and dropped the boost dependency. I have modified it a bit so the additional features (reflexted in/out, xorout) are still supported.

I have compared thousands of runs with the results from boost CRC using randomly generated parameters and data (CRC bit lengths 8, 16-64 (it seems boot CRC does not build eg. with bit length 10)).
I have compared some results by hand with [this online tool](http://www.sunshine2k.de/coding/javascript/crc/crc_js.html), and some values for CRC bit length < 8 with [this other online tool](https://www.ghsi.de/pages/subpages/Online%20CRC%20Calculation/index.php?Polynom=10100110001&Message=f1015a42).

Thank you for the modification of the tests structure, I will write the unit-test for this next. What do you prefer: would you merge this branch and then I would create a new branch / pull request from there or should I rebase this back to the master and make the unit-test commits on this?

PR #325, opened 2021-10-13
Close button in pattern popup

LFriede:
Hi,

There is no close button in the "Load pattern..." popup window. You can't close that popup without loading a pattern.

I just added that button.


WerWolv:
Thanks a lot :)

LFriede:
You're welcome. I just discovered ImHex, I'll do more stuff :)
Thank you for this great project!

PR #335, opened 2021-11-01
Tests for the CRC and hash algorithms

raron:
Hello,

I made the unit tests for the CRC and the hash algorithms. I have found some smaller bugs during this, which have been fixed.

There is a significant question about the `hex::crypt::encode16` in which the function `mbedtls_mpi_read_binary` ignores the initial null bytes from the input and so the result of the `encode16({0x00, 0x2a})` is `"2A"`.

Technically (as mbedtls_mpi_ is a bignum library) this is true (as mathematically writing zeros before a number does not change it's value), but it makes the encoding lossy (the information about the length of the input can be lost), and it could make `decode16(encode16(x)) == x` false, what (I think) is mostly unexpected.

At this moment the test samples affected by this are skipped and a single last case makes the test case to fail (so the other tests will run). (I have not come up with a good fix yet which is not ugly (eg. prepending the output of encode16 before return) without implementing the whole thing by hand.)

I would like to get your opinion about this: is this the expected result from the conversion / should it be checked by the tests / should it be fixed?


WerWolv:
Thanks a lot for all the work first of all, the tests look great!

About the encode16, this is absolutely an oversight and should be fixed. I think as of right now nothing uses the encode function so it never broke so far.
Implementing it manually might just be the best option. Using libfmt for it should be reasonably fast but if not, I can implement a faster version myself.

WerWolv:
Do you think this is ready for merge?

raron:
Yes, I think.

WerWolv:
Great thanks a lot :)

PR #336, opened 2021-11-04
Fix Provider::getBaseAddress

moonshadow565:
Fixes getBaseAddress which also fixes jumping across pages.

What is base address supposed to be anyway?
Is it base address to load file at?
Is it offset in file to load at?
Is it current display address?

### DIFF
--- a/plugins/libimhex/source/providers/provider.cpp
+++ b/plugins/libimhex/source/providers/provider.cpp
@@ -100,7 +100,7 @@ namespace hex::prv {
     }
 
     u64 Provider::getBaseAddress() const {
-        return this->m_baseAddress + PageSize * this->m_currPage;
+        return this->m_baseAddress;
     }
 
     size_t Provider::getSize() const {


WerWolv:
Hi and thank you!
`getBaseAddress` returns the start address of your loaded data. For most files, this should just be 0x00 but if you have e.g a RAM dump you might want the addresses to all start at 0x8000'0000 instead.
When moving to a different page, does it still display the correct address to the left of the data? That was the reason I added that page offset calculation there I think.

moonshadow565:
No, because setPage is called 2 times:

<https://github.com/WerWolv/ImHex/blob/master/source/views/view_hexeditor.cpp#L1243>
```cpp
// explictly here
provider->setCurrentPage(std::floor(double(newOffset - baseAddress) / prv::Provider::PageSize));

// from callback here
EventManager::post<RequestSelectionChange>(Region { newOffset, 1 });
```

And because baseAddress is subtracted in both cases it ends up subtracting the address from itself which results in page 0.

moonshadow565:
> but if you have e.g a RAM dump you might want the addresses to all start at 0x8000'0000 instead.

Provider should probably handle all rebasing internally so maybe it would be better if setBaseAddress was not public?
In that way all external code would allways deal with virtual addresses.

WerWolv:
Yeah but with your changes now, does switching to another page make the addresses on the left start at zero again or does it properly display the correct address of the data now?
Like, if a page is 0x1000'0000 bytes long, the first page should start counting at 0x00, the second one should start counting at 0x1000'0000.

If it doesn't, that calculation needs to be added elsewhere.


> Provider should probably handle all rebasing internally so maybe it would be better if setBaseAddress was not public?
In that way all external code would allways deal with virtual addresses.

It needs to be exposed because you can also manually set the base address through the `File` menu or through a pragma in pattern language source code.

moonshadow565:
No address display is incorrect.

WerWolv:
Okay that'd need to be fixed first then. `getBaseAddress` is used in pretty much every view to display the absolute address of things. In the pattern data view to show the address of placed patterns, in the disassembly view to show instruction addresses, in the string finder view to show string addresses.

I think we should keep `getBaseAddress` as it is but rename it to `getPageStartAddress` and use that everywhere instead of `getBaseAddress`. Then add a new function called `getBaseAddress` with your changes in place and use that for e.g the goto command and the other things that were broken. What do you think about that?

moonshadow565:
Reading is now off by one page as well.
I think that sounds good.

Alternatively you could also split all functions in provider that take address.
So instead of passing getBaseAddress() you would call func_physicall or func_virtual where necessary.

WerWolv:
That's fine by me as well but I'd use the terms `Relative` and `Absolute`

moonshadow565:
That's also good with maybe a comment to note what it is relative to.
Main reason i noticed page jumping is broken is because of this: <https://github.com/moonshadow565/ImHex/tree/hacky-reclass>.
Would be nice to be able to do something like that via plugin.
You can close this PR now.

WerWolv:
Oh nice I wanted to implement something like that for ages. 
This can absolutely be done via plugins right now, it just needs to be built in-tree currently.

I'll see if I can clean up that mess a bit then. I didn't even notice jumping being broken because I rarely look at large files.

PR #338, opened 2021-11-09
Fix infinite loop on signal handling.

KokaKiwi:
ImHex catch the SIGABRT but the signal handler raising the same signal makes the whole thing doing an infinite call loop (SIGABRT -> sighandler -> raise SIGABRT -> SIGABRT -> sighandler -> etc.)

The fix is simply to unregister the signal handler when catching a signal, as it terminates the whole program with SIGABRT anyway (maybe re-raising the same signal instead of SIGABRT?)

### DIFF
--- a/source/window/window.cpp
+++ b/source/window/window.cpp
@@ -242,8 +242,8 @@ namespace hex {
             EventManager::post<EventAbnormalTermination>(signalNumber);
 
             // Let's not loop on this...
-            std::signal(SIGABRT, nullptr);
-            std::raise(SIGABRT);
+            std::signal(signalNumber, nullptr);
+            std::raise(signalNumber);
         };
 
         std::signal(SIGTERM, signalHandler);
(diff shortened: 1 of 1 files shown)


source/window/window.cpp
```diff
@@ -241,6 +241,8 @@ namespace hex {
         auto signalHandler = [](int signalNumber) {
             EventManager::post<EventAbnormalTermination>(signalNumber);
 
+            // Let's not loop on this...
+            std::signal(SIGABRT, nullptr);
             std::raise(SIGABRT);
```
KokaKiwi, line 246: Why not re-raising the original signal here?

```suggestion
            std::signal(signalNumber, nullptr);
            std::raise(signalNumber);
```

WerWolv:
Hi
Thanks a lot! This was reported by a few people but I didn't look into it yet. For some reason this works totally fine on Windows...

source/window/window.cpp
WerWolv: Yeah re-raising is probably smarter

KokaKiwi:
> For some reason this works totally fine on Windows...

Signal handling seems to be kinda platform-defined so it depends on... wibbly-wobbly stuff :smile:

WerWolv:
Wibbly wobbly timey wimey stuff :P
Thank you :)

Issue #328, opened 2021-10-18
[Feature] support regex search in string view


giltabibian:
### What feature would you like to see?

support regex search in string view

### How will this feature be useful to you and others?

Sipport more than exact match on string. 

### 

- [ ] I can provide PoC for that feature or am willing to work on it myself and submit a PR

giltabibian:
Thanks for adding this support!

giltabibian:
@qdlmcfresh thanks for adding regex support!
Is it possible to add also count of the detected pattern?

qdlmcfresh:
Yeah sure np :)

qdlmcfresh:
![grafik](https://user-images.githubusercontent.com/10837173/144105825-59979dd8-f2d8-4701-b99d-88bae7ab9df5.png)
Like this?

giltabibian:
Yes! Great 😊Quick! Thanks!

PR #345, opened 2021-11-17
add regex filter in string view

qdlmcfresh:
Implement #328 
Use C++-Style Regex -Strings to filter the output
grafik


WerWolv:
Thanks a lot :)

PR #352, opened 2021-11-29
ImGui: Nearest-Neighbour Font Upscaling

DavidBuchanan314:
Switch from `GL_LINEAR` to `GL_NEAREST` for font texture upscaling.

This makes the bitmap font crisp on hidpi displays (tested on a retina Macbook Pro 14", 2021)

This needs further testing on other configurations - it's plausible that I've made other usecases worse.

Before:
image

After:
image


DavidBuchanan314:
The font still looks slightly blurry in the second screenshot, but that is just an artefact of the screenshot mechanism...

PR #353, opened 2021-11-30
Add result count to string view

qdlmcfresh:
As requested in #328 the string view now displays the result count

### DIFF
--- a/plugins/builtin/source/lang/de_DE.cpp
+++ b/plugins/builtin/source/lang/de_DE.cpp
@@ -285,6 +285,7 @@ namespace hex::plugin::builtin {
                     { "hex.view.strings.filter", "Filter" },
                     { "hex.view.strings.extract", "Extrahieren" },
                     { "hex.view.strings.regex_error", "Ungültiges Regex" },
+                    { "hex.view.strings.results", "%i Ergebnisse" },
                     { "hex.view.strings.searching", "Suchen..." },
                     { "hex.view.strings.offset", "Offset" },
                     { "hex.view.strings.size", "Grösse" },
--- a/plugins/builtin/source/lang/de_DE.cpp
+++ b/plugins/builtin/source/lang/de_DE.cpp
@@ -285,7 +285,7 @@ namespace hex::plugin::builtin {
                     { "hex.view.strings.filter", "Filter" },
                     { "hex.view.strings.extract", "Extrahieren" },
                     { "hex.view.strings.regex_error", "Ungültiges Regex" },
-                    { "hex.view.strings.results", "%i Ergebnisse" },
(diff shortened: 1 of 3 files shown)


plugins/builtin/source/lang/de_DE.cpp
```diff
@@ -285,6 +285,7 @@ namespace hex::plugin::builtin {
                     { "hex.view.strings.filter", "Filter" },
                     { "hex.view.strings.extract", "Extrahieren" },
                     { "hex.view.strings.regex_error", "Ungültiges Regex" },
+                    { "hex.view.strings.results", "%i Ergebnisse" },
```
WerWolv, line 288: You should use `{0}` here instead and use `hex::format` and `ImGui::TextUnformatted` to format and display the string. In cases with multiple format placeholders, this form can be used in languages where the placeholders might not be in the same order as in english. Doesn't matter for only one specifier but it's nice to have it all in the same form.

plugins/builtin/source/lang/en_US.cpp
```diff
@@ -288,6 +288,7 @@ namespace hex::plugin::builtin {
                     { "hex.view.strings.filter", "Filter" },
                     { "hex.view.strings.extract", "Extract" },
                     { "hex.view.strings.regex_error", "Invalid regex" },
+                    { "hex.view.strings.results", "Found %i occurrences" },
```
WerWolv, line 291: Same as above

source/views/view_strings.cpp
```diff
@@ -139,6 +139,11 @@ namespace hex {
                     ImGui::TextSpinner("hex.view.strings.searching"_lang);
                 }
 
+                if(this->m_foundStrings.size() > 0){
+                    ImGui::SameLine();
+                    ImGui::Text("hex.view.strings.results"_lang, this->m_filterIndices.size());
```
WerWolv, line 144: As mentioned above, `ImGui::TextUnformatted` and `hex::format` are better.

source/views/view_strings.cpp
```diff
@@ -139,6 +139,11 @@ namespace hex {
                     ImGui::TextSpinner("hex.view.strings.searching"_lang);
                 }
 
+                if(this->m_foundStrings.size() > 0){
+                    ImGui::SameLine();
```
WerWolv, line 143: This puts the number of results string on the same line as the TextSpinner if it's there. I think results start appearing as soon as they are found so this might clash. Consider making this an `else if` of the above condition? Also please put a space between if and the round bracket as well as between the closing round bracket and the curly one :)

WerWolv (changes requested):

WerWolv:
Thanks a lot for the PR :)

WerWolv (approved):

PR #354, opened 2021-12-01
ci/win: use option 'pacboy' of setup-msys2 to install dependencies

umarcor:
This PR uses option 'pacboy' of setup-msys2 in order to install build dependencies, and take advantage of setup-msys2's caching feature.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       with:
         fetch-depth: 0
         submodules: recursive
-        
+
     - name: 📜 Restore cache
       uses: actions/cache@v2
       with:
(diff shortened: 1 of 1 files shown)


WerWolv:
Cool thanks a lot for the PR!

Could you please restore `dist/get_deps_msys2.sh` though? This is also useful for people that would like to build ImHex themselves on Windows.

umarcor:
Since there is no build script for Windows, IMHO users who want to built ImHex locally should use the recipe: https://github.com/WerWolv/ImHex/blob/master/dist/msys2/PKGBUILD. 

Anyway, I will revert dist/get_deps_msys2.sh, since that is out of the scope of this PR.

umarcor:
Done!

Issue #81, opened 2020-12-09
AppImage image releases


its0x08:
Hello,
Can we have a single file binary that we can just chmod and run?
Nothing worked for me on Kali (latest), which is sad to be honest..
Thanks

Thog:
Never did an AppImage but it could be a nice thing to have I agree! ^^

Tcll:
I'd like to back this
would like to be able to simply download and run ;)

WerWolv:
PRs for this are absolutely welcome! I have no idea how they work (the same as with the dmg files on mac) which will cause them to become the same mess as the dmgs are right now

Tcll:
I've never actually compiled one, but I kinda know what I'm supposed to do...
I'm not the best to rely on for this though... but I can absolutely inform ;)

there are 2 ways to build appimage binaries:

the first is just a simple pack all relative binaries into the local `/bin`, `/lib`, and other directories
but this doesn't exactly guarantee compatibility unless the binaries being packed are guaranteed portable...

thus the second and more complex method of building and packing a squashfs virtual environment comes into play
this method will induce migraines, but guarantees no reliance on kernel-specific libraries that would otherwise prevent the appimage from running.

I've never actually built a squashfs environment though and wouldn't know the first thing about doing so I'm afraid :(

Tcll:
well, as a continuation of my previous post, since I've actually been looking into these things for some time...

I've recently been in contact with some of the unofficial packagers over at the minetest community, and it looks like they use a docker instance to get the squashfs environment following a template folder...
but it looks like the key part of appimage functionality is in static linking when compiling...

or in other words, the program must already be portable (capable of running from a flash drive on any distro)
like Blender for example...

at least this is the general vibe I'm getting from talking with numerous devs about portability...
uTox for example is looking to go full single-file executable instead of an appimage.

eadmaster:
as alternative, [statically linking the Linux elf](https://www.systutorials.com/how-to-statically-link-c-and-c-programs-on-linux-with-gcc/) would also work for me, but it looks a tricky process.

its0x08:
Thats what I meant by AppImage.. statically linking the ELF to make it portable.
Not dockerizing it!

follower:
An AppImage would be great--I tried the Flatpak but it complained about needing `org.freedesktop.Platform/x86_64/20.08`...

[...]

PR #355, opened 2021-12-01
Appimage builder

wardwouts:
This creates a setup to fulfill https://github.com/WerWolv/ImHex/issues/81

I've successfully created AppImages for both v1.11.1 and the master branch using this.


WerWolv:
Wow thanks a lot! I've been meaning to do this for ages!

Could this somehow be integrated into the existing github actions build process so new commits are automatically bundled? That would make distributing future releases as AppImages a lot easier for me as well

wardwouts:
> Wow thanks a lot! I've been meaning to do this for ages!
> 
> Could this somehow be integrated into the existing github actions build process so new commits are automatically bundled? That would make distributing future releases as AppImages a lot easier for me as well

I don't know anything about github actions. But if creating and using docker containers there is possible, then running the following:
```
cd docker
export TAG=v1.2.3
build.sh
extract.sh
```
should put a tagged AppImage like `ImHex-v1.2.3-x86_64.AppImage` in that docker directory. It could be that the security options in `build.sh` need a little tweaking for github actions though. I've tried to keep the scripts straightforward, so they should be easy to adapt.

WerWolv:
Hm it's probably best to separate it. ImHex is already built twice on Linux, once for the portable version and once for the flatpak because I don't think it lets you use a pre-built binary. In the docker image ImHex would then be compiled _again_ to package it into an AppImage.

This is out of scope of this PR so I'd merge this and later on try to integrate it into the github actions without docker.

Could you perhaps move everything you put into the `docker` folder into `/dist/AppImage` or something compareable? Since the `/dist` folder is where everything else related to building ImHex resides in

wardwouts:
Done.

For github actions this is likely a showstopper: https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#supported-linux-capabilities
FUSE, which is needed to build the AppImage, really needs those additional capabilities.

WerWolv:
Thanks :)

The AppImage documentation seems to suggest using this action here: https://github.com/marketplace/actions/build-appimage but that uses some sort of yml recipe file

wardwouts:
As far as I understand appimage-builder recipes depend on the software being put in the AppImage being available as .deb files. Which is why I used appimage-tool for packaging.

WerWolv:
Ah unfortunate. I'll see if I can figure out some other way to do it then.
Anyways, thanks a lot for your contribution :)

wardwouts:
I just realized one thing. As you already do a build in the github actions, that part of the Dockerfile can be skipped if you want to bring all this to github actions. After the build of the source code the only things being done are:

1. copying files into the correct directory structure (the `ldd` line is key for all the dependencies)
2. packing it up

(I got the information on how to do this here https://appiomatic.com/blog/creating-appimage-binary-manually-for-linux-from-your-app/ )

Copying files should be easy.  Packaging using appimage-tool may be tricky, as that uses fuse. But there is another way. Here https://github.com/AppImage/AppImageKit the README.md describes:
```
If you want to generate an AppImage manually, you can:

mksquashfs Your.AppDir Your.squashfs -root-owned -noappend
cat runtime >> Your.AppImage
cat Your.squashfs >> Your.AppImage
chmod a+x Your.AppImage
```

I'll try to get this working and wrap this up in a shellscript to get all the actual steps clear. With that it should be pretty straightforward to move everything into github actions.

wardwouts:
Would something like this work for you?

https://github.com/wardwouts/ImHex/blob/simplify_appimage_build/dist/AppImage/package.sh

If so, I'll cleanup the no longer needed bits and create another PR.

WerWolv:
Yeah that would be perfect, thanks a lot!

PR #356, opened 2021-12-02
Simplify appimage build

wardwouts:


WerWolv:
Thanks! 

You can leave the files from your previous PR around so people that want to package it themselves can do so easily.
Also, I honestly rather not have pre-built binaries in the repo. Could you download `runtime-x86_64` and `AppRun-x86_64` from inside the script with e.g `wget`?

wardwouts:
On it. I'm going to push an update so I can properly test against my own repo. Will let you know when I've tested it. Changed docker to also use this `package.sh` script, making that end easier too.

wardwouts:
And with that final tweak it should be fine now.

WerWolv:
Thanks a lot :)

Issue #358, opened 2021-12-03
[Bug] Crash when frame rate is low


wardwouts:
### Operating System

Linux

### What's the issue you encountered?

When the framerate is low (e.g. when using a remote connection) and there is no popup open ImHex crashes with:
```
$ ImHex
[INFO]  Welcome to ImHex!
imhex: /build/glfw3-kgzjId/glfw3-3.3.2/src/window.c:1085: glfwWaitEventsTimeout: Assertion `timeout >= 0.0' failed.
zsh: abort      ImHex
```

As far as I can see this is due to this line (#287) https://github.com/WerWolv/ImHex/blob/5a2e2b1773bce6a867cf1dd10b27bd9fa092f762/source/window/window.cpp#L287

The last part `(this->m_lastFrameTime - glfwGetTime() + 1 / 5.0))` can go negative when the framerate drops below 5.

### How can the issue be reproduced?

Try using ImHex over an ssh connection and open a file.

### ImHex Version

1.11.1

### 

- [ ] Nightly or built from sources

### Additional context?

- Additional information about your environment.
- If possible and useful, please upload the binary you've been editing when the bug occured.

PR #359, opened 2021-12-03
Fix timeout

wardwouts:
This simply makes sure the glfwWaitEventsTimeout value is never negative, this so glfw does not crash ImHex. As far as I can see this fixes https://github.com/WerWolv/ImHex/issues/358


WerWolv:
Thank you!
Question though, how on earth did this ever work? `this->m_lastFrameTime` should **always** be smaller `glfwGetTime()`. Does it work if you swap the two around? e.g `glfwGetTime() - this->m_lastFrameTime` instead of `this->m_lastFrameTime - glfwGetTime()`

wardwouts:
With a framerate higher than 5fps the addition of 0.2 pulls it back to positive.

From what I understand this is done to keep cpu usage down, but giving it a bit of a sleep between looking for interactions.

wardwouts:
Ok, I've tested 4 versions of setting this timeout now:

1) simply swapping (and taking out that +0.2):
```
glfwWaitEventsTimeout(ImGui::IsPopupOpen(ImGuiID(0), ImGuiPopupFlags_AnyPopupId) ? 0 : (glfwGetTime() - this->m_lastFrameTime));
```
This works, but when a popup is open CPU usage spikes, for what feels like no good reason.

2) no extra code for pop-ups:
```
glfwWaitEventsTimeout(glfwGetTime() - this->m_lastFrameTime);
```
This seems to work fine. Especially then the frame rate is low. When the frame rate is high this a significant amount of CPU. Again for what feels like no good reason.

3) fixed timeout of 0.1
```
glfwWaitEventsTimeout(0.1);
```
This works fine on both high and low frame rates. Lower cpu usage everywhere.
4) fixed timeout of 0.2
```
glfwWaitEventsTimeout(0.2);
```
Slightly lower CPU usage than 0.1. I don't feel any difference in UI responsiveness.

Given all this I think #4 makes most sense.

WerWolv:
Thank you for looking into this. This extra calculation is done to limit the frame rate to exactly 5 FPS. Since it takes some time to draw the frame itself, you can't always just wait 200ms to get to 5 FPS, you need to wait 200ms - frameDrawingTime.

For me, hardcoding 0.2 results in an idle frame rate of about 4.2FPS which doesn't really matter but 5.0 FPS just looks nicer. 
Also the extra code to check if a popup is open is necessary to not make the popup opening animations super laggy. This should probably be handled in a more portable way but as of right now, your code makes those popup animations lag again.

For this reason, it should be the following:
```cpp
glfwWaitEventsTimeout(
    ImGui::IsPopupOpen(ImGuiID(0), ImGuiPopupFlags_AnyPopupId) 
        ? 0 : (1.0 / 5.0) - (glfwGetTime() - this->m_lastFrameTime));
```
`(1.0 / 5.0)` being the idle frame time for 5 FPS which we then subtract the time the frame took to render from.

wardwouts:
I can understand this when the frame rate is 5 fps. But what if it's (much)
higher? Which seems to be most of the time.

WerWolv:
Oh yeah it still needs to make sure that the value is greater than 0 for instances where the frame rendering takes more than 200ms

PR #378, opened 2021-12-21
Remove semicolons after macros in resources

qxxxb:


WerWolv:
How does that change anything? In my opinion there should never be any semicolons at the end of the macro definitions themselves but only when defining them.

qxxxb:
This PR was a bit of a nitpick, but there was an extra semicolon with `RESOURCE_EXPORT`:
```c
extern "C" unsigned char splash[]; extern "C" int splash_size;;
```
I moved semicolons in `resources.cpp` for consistency. In any case, this is superseded by `libromfs`, though you might want to double check if you have extra semicolons there: https://github.com/WerWolv/libromfs/blob/master/generator/source/main.cpp#L33

WerWolv:
You're right, there's indeed an extra one there. Thanks :)

Issue #373, opened 2021-12-17
[Bug] [Debian 11] ImHex exits/crashes when trying to change scaling


kittennbfive:
Hi,

when changing the scaling (because by default the text is quite small with a screen resolution of 2560x1600) ImHex just exits (crashes?). If started from the terminal there is no output except the welcome-message. If started again the scaling is still the default, so something is wrong...

Debian 11.1

release v1.12.1 as .DEB installed normally using apt.

I had the same problem with the current AppImage, so it seems not specific to the .DEB.

WerWolv:
ImHex doesn't crash there, it's supposed to restart itself in order for the scaling to come into effect. It does correctly restart on Windows so the real issue on Linux is that it doesn't restart apparently?

lilymara-onesignal:
I'm also seeing scaling settings fail to apply on macOS aarch64 (though in my case the text is too large initially).

lilymara-onesignal:
I was able to resolve my issue by running 1.11.2, changing the setting there, quitting, then opening 1.12.1.

qxxxb:
> I'm also seeing scaling settings fail to apply on macOS aarch64 (though in my case the text is too large initially).

Does ImHex restart after you change the scaling (but with the scaling settings not applied)? Or does it fail to restart entirely?
On Ubuntu 20.04, this issue that ImHex wasn't restarting properly due to missing environment variables (#382)
```
[ERROR] GLFW Error [65544] : X11: The DISPLAY environment variable is missing
[FATAL] Failed to initialize GLFW!
```

lilymara-onesignal:
ImHex was restarting but not applying the setting. After running 1.11.2 and fixing the scaling, it appears that 1.12.1 is able to restart and apply the setting with no problem. Perhaps some file was not being created by 1.12.1, but did get created by 1.11.2.

rubin55:
I'm having this exact issue: when I open the settings screen (which is scaled too large, see #397), Go to Interface -> Scaling, it is selected as `native` by default. When I change it to `2x`, it succesfully automatically restarts, but the scaling is not applied. This is on 1.12.1 master@f45cc3f.

PR #380, opened 2021-12-21
Store environment variables and use on restart

qxxxb:
- Fixes WerWolv/ImHex#373 on Linux

ImHex does not restart properly due to missing environment variables
```
[ERROR] GLFW Error [65544] : X11: The DISPLAY environment variable is missing
[FATAL] Failed to initialize GLFW!
```

#373 seems to have different issues on macOS though


WerWolv:
Thank you! I didn't even know you could do that.

PR #382, opened 2021-12-22
Fix saving files

qxxxb:
The file save functionality is currently broken on Linux. This patch fixes it, but note that saving a file does not set `ProjectFile::s_hasUnsavedChanged` to `true`, so the dirty indicator `(*)` stays in the title bar. I don't know what project files are used for so I'm not sure what the expected behavior here is.


WerWolv:
Thank you!
Project Files are basically bundling a path to the file you're editing and then everything you did in ImHex around that file. E.g patched bytes, pattern language source code, bookmarks and so on. The (*) indicator staying is fine, it should only disappear if all things have been saved, not just changes made to the file.

PR #395, opened 2022-01-02
sys: Add functionality to clear recent files

draftshade:
Added hyperlink on welcome page and a menu item in `File -> Open Recent` to clear recent files.


WerWolv:
Thanks a lot! 

How did you translate the new text to all supported languages? If it's through a Translator, I suggest to just replace them with a placeholder. I always just populate the entries for the languages I know and then copy over the English translation to the other languages and comment out the line.

Besides that, it might look decent to have some icon from the built-in font awesome or vscode icons for the clear button on the welcome screen. What do you think?

draftshade:
For german and english I knew the word and for italian and chinese I used langenscheidt.com.
Regarding the icon, do you mean instead of the "Clear" link or additionally next to it?

WerWolv:
Yeah I'd just leave the translations you don't know out (commented out english version) and leave it to the translators. Otherwise it's easy to accidentally use the wrong words etc.

I thought instead of the word "Clear" on the welcome screen, just display an `X` icon maybe. The link itself looks a bit out of place in my opinion. The link in the context menu is great the way it is.

PR #401, opened 2022-01-18
Fix usage of CMAKE_INSTALL_PREFIX on Linux

manokara:
Typically one would install locally built packages in the `/usr/local` prefix, which is the default value of `CMAKE_INSTALL_PREFIX`, by running `make install`. However, the way the variable is used in the CMake definitions assumes that `CMAKE_INSTALL_PREFIX` is `/`, which would generate incorrect paths (e.g. `/usr/local/usr/bin/imhex` instead of `/usr/local/bin/imhex`). This PR aims to fix that.


WerWolv:
Thanks! This does mess up packaging of all Linux artifacts though. Can you adjust the github actions build script too so it puts the files in the /usr folder there (or fix the paths to point to the non-usr directory, not sure which one is better)

manokara:
Oh indeed, `CMAKE_INSTALL_PREFIX` is overriden for packaging. In that case I should probably add `/usr` directly in the build script so it still respects the prefix in local builds.

WerWolv:
Cool thanks! The DEBIAN folder still gets added to e.g the ELF artifact though. You'd have to copy it over in the workflow script, then package the deb package and then delete the file again because the other upload/package steps act on the same `build/install` folder as the DEB one

manokara:
Should I squash the commits?

WerWolv:
I'll squash them when I merge the PR. Just waiting for the CI to run now before I merge. Thanks a lot :)

PR #404, opened 2022-01-21
add hints to debug build failures

sjossi:
I added a line with a few commands to help sort out issues with building on macOS. `brew doctor` will complain about not updated SDKs etc. I purposely kept it short, because it's just a hint. Let me know if I should refine it (I would then maybe put it in the wiki)


WerWolv:
Thanks!

PR #416, opened 2022-02-06
Fix DEBUG without GIT_* variables defined

roblabla:
When building ImHex in debug mode outside of a git repo, the build would fail due to missing GIT_ variables.


WerWolv:
Thank you!

PR #418, opened 2022-02-07
Add Japanese Translation

gnuhead-chieb:


WerWolv:
Wow thanks a lot!

plugins/builtin/source/lang/ja_JP.cpp
```diff
@@ -0,0 +1,695 @@
+#include <hex/api/content_registry.hpp>
+#include <hex/api/localization.hpp>
+
+namespace hex::plugin::builtin {
+
+    void registerLanguageEnUS() {
```
WerWolv, line 6: This function should be renamed to `registerLanguageJaJP()` and has to be declared and called in `plugin_builtin.cpp` for the language to be loaded.

plugins/builtin/source/lang/ja_JP.cpp
```diff
@@ -0,0 +1,695 @@
...
+        ContentRegistry::Language::registerLanguage("Japanese", "ja-JP");
+        LangEntry::setFallbackLanguage("en-US");
```
WerWolv, line 8: This can be removed as it's already being set in the English translation file

WerWolv (changes requested):

WerWolv (approved):

PR #422, opened 2022-02-07
update the stb_image.h to the latest version!

dianwoshishi:
fix the problem that when call the function ImGui::LoadImageFromMemory, the function always return nullptr!
this problem was fixed in the latest version of stb


WerWolv:
Already fixed the issue in the last commit, sorry ^^

PR #425, opened 2022-02-08
nodes: Validate some buffer operations

manokara:
- NodeBufferSlice would always error because the bounds condition had the wrong operator
- NodeCastBufferToInteger would produce weird values if the source buffer didn't completely fill the integer.
- Also made a small refactor to number constant processing


WerWolv:
Thank you!

PR #426, opened 2022-02-08
fix errors when compiling example_cpp

dianwoshishi:


WerWolv:
Thank you!

WerWolv:
Btw, there's this repo here that can be used as a template too: https://github.com/WerWolv/ImHex-Cpp-Plugin-Template
Need to update it to 1.14.0 but it would allow for a much cleaner build process for your plugin

WerWolv:
Updated it

dianwoshishi:
OK！
why not convert it to a submodule in /plugin/example_cpp？

WerWolv:
Because the point of that template is that you can build Plugins out-of-tree without having to maintain a fork of ImHex itself.

dianwoshishi:
> Because the point of that template is that you can build Plugins out-of-tree without having to maintain a fork of ImHex itself.

got it!

PR #427, opened 2022-02-08
nodes: Allow immediate values on node inputs

manokara:
In numeric inputs, a textbox where you can input a value manually will be shown when it doesn't have a connection. I think it needs a bit of work on UI/UX though.

screenshot

### DIFF
--- a/lib/libimhex/include/hex/data_processor/attribute.hpp
+++ b/lib/libimhex/include/hex/data_processor/attribute.hpp
@@ -20,8 +20,7 @@ namespace hex::dp {
         };
 
         enum class IOType {
-            In,
-            Out
+            In, Out
         };
 
         Attribute(IOType ioType, Type type, std::string unlocalizedName);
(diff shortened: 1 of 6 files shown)


lib/libimhex/include/hex/data_processor/attribute.hpp
```diff
@@ -55,6 +61,13 @@ namespace hex::dp {
         std::map<u32, Attribute *> m_connectedAttributes;
         Node *m_parentNode = nullptr;
 
+        bool m_allowsImmediate;
+
+        union {
```
WerWolv, line 66: This is kinda eh... I think the better approach would be to make the Attribute class virtual and then have a AttributeInteger, AttributeFloat and AttributeBuffer class that contains the logic for the different types.

lib/libimhex/include/hex/data_processor/attribute.hpp
```diff
@@ -33,6 +32,11 @@ namespace hex::dp {
         [[nodiscard]] IOType getIOType() const { return this->m_ioType; }
         [[nodiscard]] Type getType() const { return this->m_type; }
         [[nodiscard]] const std::string &getUnlocalizedName() const { return this->m_unlocalizedName; }
+        [[nodiscard]] bool allowsImmediate() const { return this->m_allowsImmediate; }
```
WerWolv, line 35: Is there a place where you wouldn't want an immedate value if none is connected?

lib/libimhex/include/hex/data_processor/attribute.hpp
```diff
@@ -33,6 +32,11 @@ namespace hex::dp {
...
+        [[nodiscard]] u64 integerImmediate() const { return this->m_integerImmediate; }
+        [[nodiscard]] float floatImmediate() const { return this->m_floatImmediate; }
+        [[nodiscard]] u64* integerImmediatePtr() { return &this->m_integerImmediate; }
```
WerWolv, line 38: These should probably just return a reference. Then you don't need two different functions anymore either

WerWolv (changes requested):
All in all it looks nice but I'd go the OOP way here and make multiple separate Attribute classes, the same as we do with the Node classes you can inherit from.

PR #428, opened 2022-02-08
[feature] Add feature: Load Configuration from Lua file

dianwoshishi:
Lua is a widely used addon script language. Lots of Applications use it to load configuration or plugins! we can benefit from it, such as changing config without recompiling, implementing new features by Calling Lua functions from C++.

`Selene` is a C++11 friendly header-only binding to Lua. It why I chose it;https://github.com/dianwoshishi/Selene

I import selene into Imhex. It will be useful to help improve flexibility.

example：
welcome_screen.cpp
``` cpp
             #include <hex/helpers/LuaConfig.hpp>

            std::string imhex_desc =   LuaConfig::getLuaConfig()->get_key_value<std::string>("welcome_screen", "desc");
            ImGui::TextFormattedWrapped(imhex_desc);
```
config.lua
```lua

welcome_screen = {}
welcome_screen["desc"] = "this is a Lua script test"

```

image

modified config.lua
```lua

welcome_screen = {}
welcome_screen["desc"] = "this is a Lua script test!!!!!!!!!!!!!"

```

image


WerWolv:
Thank you!

While this is a cool feature, what would be some actual use-cases for this? We already have really extensive plugin support for C++ and Rust plugins. Without making a huge interface especially for ImGui, there's no way to create e.g custom interfaces or anything. Also most of the API currently built as a registry. Meaning, you only register things at the start once and then don't call the registry anymore so there's not a lot of room for configuring things there either.

We do currently have Python included as well which was intended for making scripts that could automatically load files, patch them, add bookmarks, data processors and generate pattern language source code. However this was never really fully implemented. Maybe that can be used for some other things? I can't think of any myself right now though.

dianwoshishi:
maybe, language files can be converted into Lua files, and load at startup. Compiling these files into the binary is inconvenient.

WerWolv:
I think if we were to separate languages from the binary itself, a simple JSON file would work well enough. Using a full blown scripting language just for storing data seems kinda overkill

PR #433, opened 2022-02-12
fix: don't drop defines after include

Diadlo:


lib/libimhex/source/pattern_language/preprocessor.cpp
```diff
@@ -84,7 +84,7 @@ namespace hex::pl {
                         preprocessor.m_defines           = this->m_defines;
                         preprocessor.m_onceIncludedFiles = this->m_onceIncludedFiles;
 
-                        auto preprocessedInclude = preprocessor.preprocess(file.readString(), true);
+                        auto preprocessedInclude = preprocessor.preprocess(file.readString(), /*initialRun =*/false);
```
WerWolv, line 87: Since nothing depends on initialRun anymore now, this can just be removed completely. It was originally used for recursive pre-processing of the input but that's now handled better by creating multiple preprocessor instances when doing that

WerWolv (changes requested):

lib/libimhex/source/pattern_language/preprocessor.cpp
Diadlo: Actually, depends. At least now we clear set of defines and pragmas at the beginning of the `preprocess` function if it's "initial" pass. It looks reasonable that we start preprocessing with empty set and don't clear it when process includes

WerWolv (approved):

PR #437, opened 2022-02-12
fix: Prevent segfault when opening a file w/o read access (linux)

bigfoot547:
When opening a file that you do not have write access to, the file provider will first open the file for reading and writing, discover it does not have permission, then attempt to open the file for reading only. This works as intended. Then, mmap is called with the PROT_WRITE flag which is forbidden, as the file was opened for reading (see `mmap(2)`). Finally, the if condition to check if mmap failed checks if mmap returned a null pointer, which it never will (instead returning `MAP_FAILED` or `(void *)-1` if an error occurred).

This PR removes the PROT_WRITE flag from the mmap call if the file is opened for reading, and properly checks if the mmap call failed.

### DIFF
--- a/plugins/builtin/source/content/providers/file_provider.cpp
+++ b/plugins/builtin/source/content/providers/file_provider.cpp
@@ -238,11 +238,13 @@ namespace hex::plugin::builtin::prv {
 
 #else
         const auto &path = this->m_path.native();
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot! I fixed this issue on Windows a while ago exactly the same lol

PR #443, opened 2022-02-16
pattern: add single_color attribute, fixed static array color override

PredatorCZ:
Added a few little things.

Fixed color attributes not working on static arrays/structs.
Added new attribute `single_color`, this attribute sets same color for all elements inside struct, basically color per struct.
`color` and `single_color` attributes can now be set for struct definitions if user wishes so.

### DIFF
--- a/lib/libimhex/include/hex/pattern_language/ast_node.hpp
+++ b/lib/libimhex/include/hex/pattern_language/ast_node.hpp
@@ -732,6 +732,8 @@ namespace hex::pl {
         if (auto value = attributable->getAttributeValue("color"); value) {
             u32 color = strtoul(value->c_str(), nullptr, 16);
             pattern->setColor(hex::changeEndianess(color, std::endian::big) >> 8);
+        } else if (auto value = attributable->hasAttribute("single_color", false); value) {
+            pattern->setColor(ContentRegistry::PatternLanguage::getNextColor());
         }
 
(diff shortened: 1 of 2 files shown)


lib/libimhex/include/hex/pattern_language/ast_node.hpp
```diff
@@ -1408,6 +1409,15 @@ namespace hex::pl {
 
             structCleanup.release();
 
+            if (!pattern->hasOverriddenColor()) {
+                if (auto value = getAttributeValue("color"); value) {
+                    u32 color = strtoul(value->c_str(), nullptr, 16);
+                    pattern->setColor(hex::changeEndianess(color, std::endian::big) >> 8);
+                } else if (auto value = hasAttribute("single_color", false); value) {
```
WerWolv, line 1416: Should this be just an `if` instead of an `else if`? If you set both the `color` and `single_color` attribute for a struct, `single_color` will be silently ignored

WerWolv:
Thanks a lot! That's a really cool feature

lib/libimhex/include/hex/pattern_language/ast_node.hpp
PredatorCZ: `color` and `single_color` do the same thing in the end, and I wanted for `color` to be always a priority, since user explicitly sets color instead or relying on the next color from color palette like `single_color` does.

PR #444, opened 2022-02-17
feature: User now can add custom directories

PredatorCZ:
Helo.
Created new setting category, where user can specify custom directories.
Custom dirs are only used for patterns, pattern includes, python scripts, magics, yara rules, constants and encodings.
Main reason is that anyone might want to use multiple git repos or sources for such files.
Not implemented for macos, because it uses only one path, but can be implemented as well.

Also added ux change, where setting categories are sorted the way they were actually created instead of alphabetical order.

Tested on linux.

### DIFF
--- a/lib/libimhex/source/api/content_registry.cpp
+++ b/lib/libimhex/source/api/content_registry.cpp
@@ -93,6 +93,10 @@ namespace hex {
                 json[unlocalizedCategory][unlocalizedName] = defaultValue;
         }
 
+        void addCategoryDescrition(const std::string &unlocalizedCategory, const std::string &unlocalizedCategoryDescription) {
+            getCategoryDescriptions()[unlocalizedCategory] = unlocalizedCategoryDescription;
+        }
+
         void write(const std::string &unlocalizedCategory, const std::string &unlocalizedName, i64 value) {
             auto &json = getSettingsData();
 
@@ -171,6 +175,12 @@ namespace hex {
             static std::map<Category, std::vector<Entry>> entries;
 
             return entries;
+        }
+
+        std::map<std::string, std::string> &getCategoryDescriptions() {
+            static std::map<std::string, std::string> descriptions;
+
+            return descriptions;
         }
 
         nlohmann::json getSetting(const std::string &unlocalizedCategory, const std::string &unlocalizedName) {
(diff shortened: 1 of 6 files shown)


WerWolv:
Thanks a lot! This looks great

plugins/builtin/source/lang/en_US.cpp
```diff
@@ -693,7 +693,8 @@ namespace hex::plugin::builtin {
                     { "hex.builtin.provider.disk.sector_size", "Sector Size" },
                     { "hex.builtin.provider.disk.reload", "Reload" },
 
-                { "hex.builtin.layouts.default", "Default" }
+                { "hex.builtin.layouts.default", "Default" },
+                { "hex.builtin.setting.folders", "Folders" }
```
WerWolv, line 697: This belongs further up (Line 677) below the other settings related strings I think

plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -219,6 +221,48 @@ namespace hex::plugin::builtin {
 
             return false;
         });
+
+        static const std::string dirsSetting { "hex.builtin.setting.folders" };
+
+        ContentRegistry::Settings::add(dirsSetting, dirsSetting, std::vector<std::string> {}, [](auto name, nlohmann::json &setting) {
```
WerWolv, line 227: This UI might benefit from a small help text explaining what it's actually doing.
Just a short text like "Adding folders here will cause ImHex to consider them as search paths for patterns, includes, magic files, etc."

lib/libimhex/source/api/content_registry.cpp
```diff
@@ -162,6 +188,15 @@ namespace hex {
             return settings;
         }
 
+        std::vector<std::string> getStringVector(const std::string &unlocalizedCategory, const std::string &unlocalizedName) {
```
WerWolv, line 191: I'd call this `getStringArray` because JSON refers to it as array

WerWolv (changes requested):

plugins/builtin/source/content/settings_entries.cpp
PredatorCZ: I added new methods for adding descriptions for categories.
They override header line that is under tabs, so the description doesn't waste space.

WerWolv (approved):

plugins/builtin/source/content/settings_entries.cpp
WerWolv: Thank you!

PR #447, opened 2022-02-18
fix: welcome screen corrupted banner when settings are open

PredatorCZ:
This issue was visible only for Linux. But when welcome screen and settings are opened, banner image gets corrupted.
It was caused by freeing resource before assigning a new.
In addition, now request for change of the theme is sent only when theme index in settings is actually changed.
And load new banner image when different resource than current is requested.

### DIFF
--- a/plugins/builtin/source/content/welcome_screen.cpp
+++ b/plugins/builtin/source/content/welcome_screen.cpp
@@ -27,6 +27,7 @@ namespace hex::plugin::builtin {
 
     static bool s_layoutConfigured = false;
     static ImGui::Texture s_bannerTexture;
+    static std::string s_bannerTextureName;
     static std::list<fs::path> s_recentFilePaths;
 
     static fs::path s_safetyBackupPath;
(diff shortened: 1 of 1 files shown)


pauleric:
Hi，I found a problem in the code below
```
        if (theme.is_number()) {
		static int lastTheme = theme.get<int>();

		if (const int thisTheme = theme.get<int>(); thisTheme != lastTheme) {
			EventManager::post<RequestChangeTheme>(thisTheme);
			lastTheme = thisTheme;
		}
	}
```
request for change of the theme Can't trigger when the program starts, I think the line
```
	static int lastTheme = theme.get<int>();
```
should be changed to
```
	static int lastTheme = -1;
```

birorobi:
Hi ... I don't understand anything..it's system crashed twice today and
yes, it's probably down to the line..Lp.

pet, 18. ožu 2022. u 11:54 paul ***@***.***> napisao je:

PredatorCZ:
I was able to replicate problem, will post fix.

@birorobi have you tried nightly build?

PR #448, opened 2022-02-19
patterns: added find_sequence_in_range function

PredatorCZ:
Hi,
I've added new builtin function to search a sequence within range.

### DIFF
--- a/plugins/builtin/source/content/pl_builtin_functions.cpp
+++ b/plugins/builtin/source/content/pl_builtin_functions.cpp
@@ -106,12 +106,14 @@ namespace hex::plugin::builtin {
                 return u128(ctx->getProvider()->getActualSize());
             });
 
-            /* find_sequence(occurrence_index, bytes...) */
-            ContentRegistry::PatternLanguage::addFunction(nsStdMem, "find_sequence", ContentRegistry::PatternLanguage::MoreParametersThan | 1, [](Evaluator *ctx, auto params) -> std::optional<Token::Literal> {
+            /* find_sequence_in_range(occurrence_index, start_offset, end_offset, bytes...) */
+            ContentRegistry::PatternLanguage::addFunction(nsStdMem, "find_sequence_in_range", ContentRegistry::PatternLanguage::MoreParametersThan | 3, [](Evaluator *ctx, auto params) -> std::optional<Token::Literal> {
                 auto occurrenceIndex = Token::literalToUnsigned(params[0]);
+                auto offsetFrom      = Token::literalToUnsigned(params[1]);
+                auto offsetTo        = Token::literalToUnsigned(params[2]);
 
                 std::vector<u8> sequence;
-                for (u32 i = 1; i < params.size(); i++) {
+                for (u32 i = 3; i < params.size(); i++) {
                     auto byte = Token::literalToUnsigned(params[i]);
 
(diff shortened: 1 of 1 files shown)


WerWolv:
Cool thank you!
I think `builtin::std::mem::find_sequence` can be be removed completely with this. Instead, we can implement `find_sequence` inside the stdlib by using `find_sequence_in_range` and passing `std::mem::base_address()` as the start address and `std::mem::base_address() + std::mem::size()` as end address. 

Also the macOS build seems to be broken right now. It seems to be because of the `paramsSpan` variable. The version of libc++ and clang that Apple uses is sadly kinda outdated and doesn't support all C++20 or even C++17 features yet.

PredatorCZ:
> Cool thank you! I think `builtin::std::mem::find_sequence` can be be removed completely with this. Instead, we can implement `find_sequence` inside the stdlib by using `find_sequence_in_range` and passing `std::mem::base_address()` as the start address and `std::mem::base_address() + std::mem::size()` as end address.
> 
> Also the macOS build seems to be broken right now. It seems to be because of the `paramsSpan` variable. The version of libc++ and clang that Apple uses is sadly kinda outdated and doesn't support all C++20 or even C++17 features yet.

Ah, thank you for clarifying. Thankfully there is a sdk available. Mac span doesn't support iterators (which is against the standard).
I haven't removed `find_sequence` because I didn't want to break backwards compatibility.
But if you're ok with that, I'll remove the old builtin function.

WerWolv:
The API that's exposed to the end user (through libstd) will stay stable so this shouldn't break anything. 

So even if the underlying implementation changes, the user will still just call `std::mem::find_sequence` and will never touch the compiler intrinsics. I made breaking changes in 1.15.0 to allow this so we wouldn't have to break the API in the future anymore

PredatorCZ:
I couldn't find the commit you're talking about.
API is fine, It's more like ABI change. Where users with older stdlib than main app will call deleted builtin function.
I guess we're relying on a good faith that users will keep both main app and stdlib up to date.
Or maybe I see way too much into it.

WerWolv:
As of right now I only make sure that libstd is compatible with the latest version of ImHex.
I know this isn't exactly great for a programming language but at the same time I think it's necessary since the language is still evolving a lot right now.

The change I was talking about was the following thing:
Before 1.15.0, users would directly use the compiler intrinsics when calling e.g `std::print`. The problem with that was, you basically have access to a bunch of random functions that are just there but have never been defined anywhere which is kinda confusing. Also any change to the signature of those function would mean it immediately breaks all code using it.

With 1.15.0, all those intrinsics got moved to the `builtin` namespace. These functions can still be called if you want to but you're not supposed or required to anymore. Instead, all those intrinsics are now wrapped in functions that can be found inside libstd which really is the only place these intrinsics should ever be called.
Changing something with the signature now only requires changes in libstd now (which the user still needs to download) but the user doesn't need to fix their code anymore

PredatorCZ:
Thank you very much for clarifying, I was able to find to commit after 1.13.

WerWolv:
Oh was it 1.14.0 that added the change already? Sorry!

PredatorCZ:
Hi, is it ready to be merged?

WerWolv:
I think so, yes. Sorry.

PR #451, opened 2022-02-20
Rework libimhex-rs to use autocxx

jam1garner:
* Restructure and rename parts of library to better fit idiomatic Rust
* Add documentation to the items in the library
* Move from just cxx to autocxx (while falling back to cxx for a function which autocxx has an issue detecting all the overloads of)
* General cleanup
* Have bookmark API use Rust-native Range type for describing regions

### DIFF
--- a/lib/libimhex-rs/build.rs
+++ b/lib/libimhex-rs/build.rs
@@ -1,11 +1,22 @@
 fn main() {
     println!("cargo:rustc-link-lib=dylib=imhex");
-    println!("cargo:rustc-link-search=all={}", env!("LIBIMHEX_OUTPUT_DIRECTORY"));
+    println!(
+        "cargo:rustc-link-search=all={}",
+        env!("LIBIMHEX_OUTPUT_DIRECTORY")
+    );
 
     println!("cargo:rerun-if-changed=src/lib.rs");
     println!("cargo:rerun-if-changed=src/imhex_api.rs");
 
-    cxx_build::bridge("src/imhex_api.rs")
+    let include = format!("-I{}/include", env!("LIBIMHEX_SOURCE_DIRECTORY"));
+
+    let path = std::path::PathBuf::from("src");
+    let mut build = autocxx_build::Builder::new("src/lib.rs", &[&path])
+        .extra_clang_args(&[&include, "-x", "c++", "-std=gnu++20"])
+        .auto_allowlist(true)
+        .expect_build();
+
+    build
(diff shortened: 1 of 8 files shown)


lib/libimhex-rs/build.rs
```diff
@@ -1,15 +1,26 @@
...
 
     println!("cargo:rerun-if-changed=src/lib.rs");
     println!("cargo:rerun-if-changed=src/imhex_api.rs");
 
-    cxx_build::bridge("src/imhex_api.rs")
+    let include = format!("-I{}/include", env!("LIBIMHEX_SOURCE_DIRECTORY"));
+
+    let path = std::path::PathBuf::from("src");
+    let mut build = autocxx_build::Builder::new("src/lib.rs", &[&path])
+        .extra_clang_args(&[&include, "-x", "c++", "-std=gnu++20"])
```
WerWolv, line 15: Is this fine for all platforms? Only macOS uses clang currently, Windows and Linux use gcc

lib/libimhex-rs/src/bookmarks.rs
```diff
@@ -0,0 +1,37 @@
+use crate::Color;
+use std::ops::Range;
+
+#[cxx::bridge]
+mod ffi {
+    #[namespace = "hex::ImHexApi::Bookmarks"]
+    extern "C++" {
+        include!("hex/api/imhex_api.hpp");
+
+        pub unsafe fn add(addr: u64, size: u64, name: &CxxString, comment: &CxxString, color: u32);
```
WerWolv, line 10: If overloading is a problem here, we can also just rename those functions to something unique and use autocxx instead. I want this to be as easy to do as possible

WerWolv (approved):

lib/libimhex-rs/build.rs
jam1garner: fwiw this is only libclang which is used internally by autocxx for parsing the C++ headers in order to generate bindings. I believe autocxx vendors its own libclang now so this shouldn't be an issue? CI seems to not have an issue at least

lib/libimhex-rs/src/bookmarks.rs
jam1garner: Renaming is probably a good strategy then! Do you have any name suggestions?

WerWolv: Honestly, the best would probably be to just only have the `add` function that takes an address and size separately and move the one that takes a Region to a `impl` namespace. Or literally just remove it. It itsn't used anywhere except to implement the other `Bookmarks::add` function

PR #453, opened 2022-02-21
Fix dynamic arrays till EOF: off-by-one

Dyddye:
I got the error `Evaluator: reached end of file before finding end of unsized array` while searching for `char[]` at end of a file, although it was a zero-terminated string there.

Adding **another** `0x00` byte at end of file fixxed it temporarily. It would be better to adjust the off-by-one range check.

### Minimal sample:
**example.bin** (containg "example\0")
```
65 78 61 6D 70 6C 65 00
```

with **DynamicCharArray.hexpat**
```
char example[] @ 0x00;
```

produces `Evaluator: reached end of file before finding end of unsized array`.


WerWolv:
Thanks a lot! I was in the progress of doing a huge refactor, that's why I didn't merge it. I added the fix myself now though

PR #457, opened 2022-02-26
patterns: Added bitfield_order pragma

PredatorCZ:
Hi,
I added new pragma `bitfield_order`, so user don't need to add attribute to each bitfield.


WerWolv:
Thank you!
I was in the progress of doing a really big refactor of the pattern language so quite a few things have changed now. Would you mind rebasing your PR? The main thing that matters to you is that all ASTNodes now live in a separate file instead of all being in one file

PredatorCZ:
Rebased

WerWolv:
Thanks a lot!

PR #472, opened 2022-03-12
Move logic to draw patterns into separate class

Diadlo:


WerWolv:
Thank you!
Could you explain a bit what advantages this brings over the way I did it so far?

lib/libimhex/include/hex/pattern_language/patterns/pattern_array_dynamic.hpp
```diff
@@ -151,6 +110,22 @@ namespace hex::pl {
             Pattern::setEndian(endian);
         }
 
+        void accept(PatternVisitor &v) override {
```
WerWolv, line 113: I think it might be nicer to make these functions all private and add the PatternVisitor class as a friend class to the various classes so they are only accessible by the visitor.

That way we also wouldn't need to add various getters to the Pattern classes that are only ever used by the visitor

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
```diff
@@ -0,0 +1,443 @@
...
+namespace hex::pl {
+
+    class ImGuiDrawer : public PatternVisitor
+    {
+    public:
+        ImGuiDrawer(prv::Provider *provider)
+            : m_provider{provider}
+        { }
+
+        void visit(PatternArrayDynamic& pattern) override {
```
WerWolv, line 31: These visit functions should all be able to be marked as `const` and with that also take the pattern either as a `const PatternXYZ *` or a `const Pattern &`

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
```diff
@@ -0,0 +1,443 @@
...
+#include <hex/pattern_language/patterns/pattern_union.hpp>
+#include <hex/pattern_language/patterns/pattern_unsigned.hpp>
+#include <hex/pattern_language/patterns/pattern_wide_character.hpp>
+#include <hex/pattern_language/patterns/pattern_wide_string.hpp>
+
+#include <nlohmann/json.hpp>
+
+namespace hex::pl {
+
+    class ImGuiDrawer : public PatternVisitor
```
WerWolv, line 24: I'd rather call this PatternDrawer or PatternRenderer since it doesn't really have much to do with ImGui

WerWolv (changes requested):

Diadlo:
> Could you explain a bit what advantages this brings over the way I did it so far?

Some time ago I started implementing JSON builder and it should went over the patterns to do its job. But here we lead to a situation where we places all possible pattern handlers in patterns itself. It makes classes overloaded with logic (provide data about concrete pattern + draw GUI + build JSON). Formally we can leave drawing as is and only add visitor for JSON builder, but I think it's better to make it in a generic way.

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
Diadlo: I don't understand you. I'm going to rename it, but why do you think it doesn't have much to do with ImGui? This class is full of ImGui calls. It can draw GUI only for ImGui

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
Diadlo: I thought about it. Formally `visit` not necessary doesn't modify incoming patterns (here it's part of an abstract interface).
Furthermore, I've tried to make it const and faced some issues since some internal functions wasn't const. I can retry and come to you with concrete cases.

lib/libimhex/include/hex/pattern_language/patterns/pattern_array_dynamic.hpp
Diadlo: For me **now** `Pattern` is more like an information provider, so I don't see any issues with adding getters. Now we decrease coupling between drawing logic and providing an information, but `friend` will increase it back. Are you sure, that it's necessary? What advantages it will provide? Less code?

I think rule of thumb is using functions even to access data because with it we can easily refactor internal logic. For example we can add value caching or vice versa calculate value instead of storing it

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
WerWolv: I just like to name things after the thing they're fullfilling, not after things that are more or less implementation details.
I could replace ImGui with some other renderer in there if I wanted and for everything outside of that class, nothing changes.

lib/libimhex/include/hex/pattern_language/patterns/pattern_array_dynamic.hpp
WerWolv: My thought was that there's a lot of weird functions in some of the Patterns now that don't serve much purpose but to expose implementation details. But yeah, you're right with the decoupling, it would make it worse as friends.

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
Diadlo: Done

lib/libimhex/include/hex/pattern_language/imgui_pattern_drawer.hpp
Diadlo: Yep. The issue is changing `displayEnd` for arrays

Diadlo: We can make it `mutable`, but should we? As I said it's ok to modify value inside a visitor

Diadlo: Hm. But acually it's part of display logic. Should we store it in Pattern?

WerWolv: Since you've moved all drawing logic out, I'd also move out all variables that are only needed for displaying if possible. 
It might not be possible with displayEnd though since that needs to be handled per-pattern

Diadlo: I will try to handle it with `map` that will match Pattern -> displayEnd in `PatternDrawer`

Diadlo: `displayEnd` is moved. But `const` still can't be used because `formatDisplayValue` accepts `Token::Literal` which is `std::variuant` with non-const `Pattern` pointer

WerWolv:
Had to fix some smaller issues with the pattern drawing and that created a bunch of conflicts here, sorry. It's not super big changes luckily

Diadlo:
I rebased the branch. Hope I didn't miss anything

WerWolv:
Thank you!
I just tried it out a bit and it seems to work really good. Two things though. 

- Patterns can't really be selected / clicked on anymore. Hovering over the lines makes them highlight for one frame and then disappear again.
- Since we have a nice interface now that we can expand upon, it might be smart to move all the pattern drawing files to the builtin plugin so the pattern language implementation is completely imgui-free. What do you think about that?

Diadlo:
Is `u8 arr[100]` static array?

If so, I can't reproduce the issue

https://user-images.githubusercontent.com/6306608/159113688-f61325e7-99c3-4802-9129-9930027fe90a.mp4

Diadlo:
Is `u8 arr[100]` static array?

If so, I can't reproduce. Can you share an example of pattern with issue?

https://user-images.githubusercontent.com/6306608/159113756-6dc4595f-cde6-4745-ac7c-6ed56c962116.mp4

Diadlo:
> Since we have a nice interface now that we can expand upon, it might be smart to move all the pattern drawing files to the builtin plugin so the pattern language implementation is completely imgui-free. What do you think about that?

Actually, I don't really understand the logic of this division right now, if you think it's better, let's move it. If you don't mind, I'd prefer to do it in a separate PR to reduce potential conflicts

WerWolv:
> Is u8 arr[100] static array?
> If so, I can't reproduce. Can you share an example of pattern with issue?

I triggered it with 
```cpp
struct Test {
    u32 x;
};

Test test[10] @ 0x00;
```

> Actually, I don't really understand the logic of this division right now, if you think it's better, let's move it. If you don't mind, I'd prefer to do it in a separate PR to reduce potential conflicts

I thought about this before since it felt weird to have ImHex built-in functionality specific stuff in libimhex which should be a common thing to use by all plugins. 
But we / I can tackle that later, that's okay.

Diadlo:
> I triggered it with

Still can't reproduce. Please, make sure that you fetched the latest version to test, since there was issue like this in one of the previous versions

WerWolv:
You're right, I tested it again now and it didn't happen anymore

WerWolv (changes requested):
Some more style nit-picks. Other than that it's looking great. Thank you!

lib/libimhex/include/hex/pattern_language/pattern_drawer.hpp
```diff
@@ -0,0 +1,469 @@
...
+        }
+
+        void draw(Pattern& pattern) {
+            if (pattern.isHidden())
+                return;
+
+            pattern.accept(*this);
+        }
+
+        template<class SomePattern>
```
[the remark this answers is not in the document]
Diadlo: Unfortunatly, we can't use this. This function accepts both `PatternArrayDynamic` and `PatternArrayStatic` and only them. Because they have `forEachArrayEntry` method. I can create an interface `PatternArray` with this method, if you want.

For now renamed to `ArrayPattern` to make the intention more clear.

lib/libimhex/include/hex/pattern_language/pattern_drawer.hpp
```diff
@@ -0,0 +1,469 @@
...
+#include <hex/pattern_language/patterns/pattern_struct.hpp>
+#include <hex/pattern_language/patterns/pattern_union.hpp>
+#include <hex/pattern_language/patterns/pattern_unsigned.hpp>
+#include <hex/pattern_language/patterns/pattern_wide_character.hpp>
+#include <hex/pattern_language/patterns/pattern_wide_string.hpp>
+
+#include <nlohmann/json.hpp>
+
+namespace {
+    const u64 DISPLAY_END_DEFAULT = 50;
```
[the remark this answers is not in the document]
Diadlo: Done, but can you clarify why is `static` needed here?

lib/libimhex/include/hex/pattern_language/patterns/pattern_float.hpp
```diff
@@ -13,19 +13,19 @@ namespace hex::pl {
...
+                return *reinterpret_cast<float *>(&data);
             } else if (this->getSize() == 8) {
                 u64 data = 0;
                 provider->read(this->getOffset(), &data, 8);
                 data = hex::changeEndianess(data, 8, this->getEndian());
-
-                this->createDefaultEntry(hex::format("{:e} (0x{:0{}X})", *reinterpret_cast<double *>(&data), data, this->getSize() * 2), *reinterpret_cast<double *>(&data));
+                return *reinterpret_cast<double *>(&data);
+            } else {
+                assert(false);
```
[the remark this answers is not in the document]
Diadlo: Good point, but I'm not completely agree. Do you use release build type?

[`assert`](https://en.cppreference.com/w/cpp/error/assert) behaviour depends on `NDEBUG` definition. If we have debug build it has to call [`std:abort`](https://en.cppreference.com/w/cpp/utility/program/abort) which has `[[noreturn]]` attribute and compiler should correctly detect that this part of code is unreachable.

But in release it does nothing and this code isn't unreachable. So I think it's wrong to call `__builtin_unreachable()` here. I will return NaN here to suppress the warning.

lib/libimhex/include/hex/pattern_language/patterns/pattern_array_dynamic.hpp
```diff
@@ -96,17 +39,31 @@ namespace hex::pl {
...
+        [[nodiscard]] size_t getEntryCount() const {
+            return this->m_entries.size();
+        }
+
         [[nodiscard]] const auto &getEntries() {
             return this->m_entries;
         }
 
+        template<class Fn>
+        void forEachArrayEntry(const Fn& fn) {
```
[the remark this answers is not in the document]
Diadlo: Good point. Thanks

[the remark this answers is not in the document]
Diadlo: Good point. Thanks

lib/libimhex/include/hex/pattern_language/pattern_drawer.hpp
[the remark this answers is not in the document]
WerWolv: Then a new concept that looks like this would work:
```cpp
template<typename T>
concept ArrayPattern = requires(T pattern, const std::function<void(int, Pattern&)>& fn) {
   { pattern.forEachArrayEntry(fn) } -> std::same_as<void>;
};
```

That will ensure that the template will only ever accept a type that has a `forEachArrayEntry` function with the correct parameters / return type

lib/libimhex/include/hex/pattern_language/pattern_drawer.hpp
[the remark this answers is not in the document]
WerWolv: Actually, `static` isn't needed. `constexpr` was the important one

lib/libimhex/include/hex/pattern_language/patterns/pattern_float.hpp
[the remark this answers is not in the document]
WerWolv: I use debug builds when developing and release types for all builds made by the CI.

The problem is, `std::abort` is indeed marked as `[[noreturn]]` but `assert` cannot be because it's dependent on its parameter whether or not it's noreturn.

This is probably a bit of a bad implementation from my side and the pattern should just take an enum in the constructor instead of allowing a size to be passed in. The evaluator will only ever pass 4 or 8 to the constructor, everything else would be a bug. That's why I used `__builtin_unreachable()` elsewhere. Returning `NaN` is fine though, I'll correct this later on

lib/libimhex/include/hex/pattern_language/pattern_drawer.hpp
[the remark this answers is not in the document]
Diadlo: I haven't used concepts before. Look cool :)

WerWolv (approved):

WerWolv:
Thanks a lot for that immense work!

PR #487, opened 2022-04-04
fix: Crash on linux when opened file gets modified

PredatorCZ:
Linux only.
When mapped file gets modified, kernel sends SIGBUS.
Added proper real time sync for file refresh.


WerWolv:
Thank you! Sorry for taking so long

PR #508, opened 2022-05-18
docs: Documentation overhaul

polak-jan:
As previously discussed on Discord, I have taken a look at the documentation structure and tried to make it more consistent and easier to navigate.
So far the changes are mostly limited to reducing the number of pages by combining them into longer ones, grouped by a topic, and by how advanced the functionality described is.

There is space for improvement in the formatting and wording of the pages, for example, the `Attributes` page is quite tightly packed and very hard to read, and other pages have similar albeit less visible issues.

Another area that I think can be quite easily improved upon is making the documentation more accessible by introducing a quickstart/guide section that explains the pattern language on specific examples. This should ideally be split into several chapters slowly introducing the more advanced features of the language.


WerWolv:
Thanks a lot for your work!

I really like the way you restructured and reworded everything. A few things that I found though when looking through it and want your opinion on too:

- Literals are currently under Data Types placed a bit arbitrarily. I'd probably put those under expressions rather since they don't really have anything to do with types.
- The entirety of the expressions page is repeated again at the bottom of the Control flow and conditional parsing page.

polak-jan:
Thanks!

The second thing is definitely a mistake on my end, no idea how I managed to do that.

As for the literals, I wasn't sure about that one. First of all, they were on the data type page already. And while they don't really fit there, they don't really fit in the expressions page either.

I can think of two options:
1) Make a new page, called `Syntax`, that is used purely to describe the basic syntax related stuff with small descriptions that just link to the actual description of the feature.
2) Keep them on the data type page. But as you said, the location of them is pretty tough to get right and the current one is honestly just wherever they ended up after I reordered the other things, only making sure that they are after the built-in types section. The best option I can think of is putting the literals for each type somewhere in the types section, but this has the downside of spreading then out and making them easier to miss, compared to having them all in a single block, and if we're to get their own sub-section to combat this, it would make the TOC pretty crowded which also isn't ideal.

KillyMXI:
I'd probably just put Literals to their own file if they don't belong to anything else.

`pattern_language.rst` seems to be the only summary file. I see no need for an extra file like `Syntax` - it's all about syntax to some extent already.

polak-jan:
`pattern_language.rst` is used by the Sphinx theme to generate the sidebar menu, it isn't really used by itself.

As for a dedicated syntax page, you are correct that it's technically all about the syntax, but with that logic you could also argue that it's all about the pattern language so it should be all in a single file. You have to draw the line somewhere, and that location is something that we need to determine.

I maintain the opinion that a dedicated syntax page would be useful as a sort of quick reference or cheat sheet, instead of needing to go through each separate page for just the couple of lines describing the syntax of the feature.

KillyMXI:
Ok. Is unaware of how the template works.
Some kind of a landing page makes sense, since there is none already in fact.
My words "all about syntax" become irrelevant. Although I have to note I didn't imply it can be merged further. If it were me to write the landing page - it would've been most likely so brief I'd still put literals in their own file.

polak-jan:
No worries, and you are right, a landing page would be a good improvement.

The issue with putting literals in their own file is that it would be at most like 10 or 20 lines. And getting rid of small files like that is one of the main goals of this overhaul. But I am not completely opposed to the idea, if we figure out a way to make it work I won't block it just because.

Issue #511, opened 2022-05-23
Flathub


Mailaender:
### What feature would you like to see?

I created a stable Flatpak from your nightly builds. https://github.com/flathub/flathub/pull/3195 Hope you don't mind.

### How will this feature be useful to you and others?

Get updated on the latest stable version with developer intended dependencies.

### Request Type

- [X] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

WerWolv:
Hey thanks a lot! I don't mind at all, however if possible I'd also like to have flatpak bundling be part of my CI so I can link to it from my nightly page.

I used to have one myself but it started to hang up the entire CI at some point and even when it was working, it basically trippled the build time for the Linux runner because it rebuilt ImHex _again_ from scratch so I removed it.

Mailaender:
Flathub has beta channels, but the infrastructure is not really ready for nightly rebuilds. I also noted that the build is quite slow and due to the many bundled dependencies, the caching can't kick in like it does for other applications.

barthalion:
It can be done with https://github.com/bilelmoussaoui/flatpak-github-actions and having a nightly manifest directly in your repo.

WerWolv:
Looks like the flatpak is public now. Thanks a lot for the help!

PR #513, opened 2022-05-27
Fixed the .desktop file

Mailaender:
For #511 the path is `/app/bin/imhex` so hardcoding is bad. The categories looked like copy and paste. I went the `Development;IDE;`. An alternative would be `Utility;TextEditor;` however I don't think it does the project justice.

### DIFF
--- a/dist/imhex.desktop
+++ b/dist/imhex.desktop
@@ -2,7 +2,7 @@
 Name=ImHex
 Comment=ImHex Hex Editor
 GenericName=Hex Editor
-Exec=/usr/bin/imhex %U
+Exec=imhex %U
 Icon=imhex
 Type=Application
 StartupNotify=true
--- a/dist/imhex.desktop
+++ b/dist/imhex.desktop
@@ -6,6 +6,5 @@ Exec=imhex %U
 Icon=imhex
 Type=Application
 StartupNotify=true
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you! That file is also used with the AppImage and the .deb file I think. Won't that cause issues there?

Mailaender:
That won't cause issues as the AppImage sets the `bin` path accordingly. I just `--appimage-extract`ed my https://github.com/OpenHV/OpenHV/releases/download/20220526/OpenHV-20220526-x86_64.AppImage to confirm.

dist/imhex.desktop
```diff
@@ -2,10 +2,9 @@
 Name=ImHex
 Comment=ImHex Hex Editor
 GenericName=Hex Editor
-Exec=/usr/bin/imhex %U
-Icon=imhex
+Exec=imhex
+Icon=imhex %U
```
WerWolv, line 6: `%U` probably should have been added to `Exec`, not `Icon` :P

WerWolv (changes requested):

dist/imhex.desktop
Mailaender: Well spotted. I added this back in because it seems important for parameterized application starts.

WerWolv (approved):

PR #524, opened 2022-06-04
revert fallback lang

LFriede:
Hi,

wondered why my "Datei"-menu (german language selected in settings) got mixed with brazilian items after updating ImHex. The new brazilian language file overwrites the fallback language from english to brazilian. I don't think this was done on purpose ;)


WerWolv:
Oh damn yeah that's definitely a mistake. Thank you!

PR #537, opened 2022-06-16
Release CI

### LINKED ISSUE #504: request: create source tarballs with submodules included for releases
As a downstream distro packager, it would be really handy if releases had a tarball that included all the submodules as they're pinned in the repo at release time

iTrooz:
Fixes #504

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,7 @@ name: Build
 
 on:
   push:
+    branches: ["*"]
   pull_request:
 
 env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,37 @@
+permissions:
+  contents: write
+
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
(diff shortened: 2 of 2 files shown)


WerWolv:
Thanks a lot!
Would it be possible to do the same thing with the current artifacts as well? That would immensely simplify the whole release creation process

I found this action here that can maybe help: https://github.com/actions/download-artifact

iTrooz:
(I'm going to push some commits to test this, sorry if it causes emails, I'm not sure how else to test the CI)

iTrooz:
Everything should be good !

iTrooz:
Btw, something I forgot to mention, you can simply disable the release workflow, and it will stop uploading artifacts to releases

WerWolv:
Thanks a lot!

classabbyamp:
Thanks for this! One thing I've noticed is that the `.git` directory is in the full sources tarball. it's 36 MB and doesn't need to be there (at least for most distro packaging)

iTrooz:
Oooops. #569

PR #547, opened 2022-06-20
Add information on generated artifacts (.deb and ELF)

iTrooz:
The .deb/ELF generated from the CI only work for Ubuntu 20.04 (because it's the version of the runner, and dependencies break between major releases like theses)

I renamed the CI artifacts to point out the distribution they are made for


iTrooz:
incorporated with #548

PR #548, opened 2022-06-21
CI misc

iTrooz:
This PR aims to fix some small things about the CI :
- do not fetch every branch/commit in checkout, but only the latest commit
- use gcc-10 to build on Linux (gcc-11 is not available by default in Ubuntu 20.04, that causes launch to fail) (see my other PR #547 for information about why Ubuntu 20.04 in particular)

I also added a "workflow_dispatch" trigger to build workflow, because I think it's always a good thing to be able to re-trigger the build manually in case something happens. Tell me if you want me to remove it

Update : this PR do other things. Check the commits name for details

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -17,7 +17,6 @@ jobs:
     - name: 🧰 Checkout
       uses: actions/checkout@v2
       with:
-        fetch-depth: 0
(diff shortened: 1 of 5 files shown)


iTrooz:
All done ! I want reword the commits, if you want me to

PR #553, opened 2022-06-26
Cache

iTrooz:
This PR fixes the cache for Ubuntu 22, and remove `restore-keys`, because according my understanding of the documentation they are redundant

As a bonus (bc from what I saw, you don't mind mixing features in PRs), this fixes a small bug in the release workflow, to always use the artifacts from the master branch


iTrooz:
(Not ready to merge, I still want to experience one thing first)

iTrooz:
Ready to merge !

PR #556, opened 2022-06-27
Update Chinese(Simplified) translation

shirok1:
Fully translated (as I can).


shirok1:
I would like to translate "Regex" in string view window, however it is hard coded.

https://github.com/WerWolv/ImHex/blob/673027c82d24d3d1242530a77fe6cf3c0d9b5401/plugins/builtin/source/content/views/view_strings.cpp#L116

WerWolv:
Hey, thanks a lot!

I'll fix the hardcoded strings in a minute

shirok1:
Pushed.

WerWolv:
Also it looks like you accidentally pushed an old version of the pattern language submodule to your PR. Could you remove that?

shirok1:
Reverted.

WerWolv:
Thank you!

PR #558, opened 2022-06-28
New cache update

iTrooz:
WARNING : unlike the branch name suggests, this PR doesn't only add cache changes for CMake

The complete list of things modified is said in the commit messages


iTrooz:
Note : I couldn't test if restore-keys worked now (as they are supposed to *keep* the CI build fast over time

Issue #557, opened 2022-06-27
[Feature] Network proxy for DLC / Wikipedia


shirok1:
### What feature would you like to see?

Network proxy can be used when downloading from store or else.

### How will this feature be useful to you and others?

For some reason, GitHub is partly blocked in Mainland China, so many Chinese developers use proxies in browsers and `git`. It would be convenient if we can use proxies to access content store in ImHex.

### Request Type

- [X] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

WerWolv:
Hey, I think this would be a nice thing to have, maybe in a separate tab of the settings window. Since there aren't any blocks like this in place where I live, it might be kinda difficult for me to test it.

Looking into other applications that have this feature already, something like this (minus the Auto-detect proxy settings option) would fit nicely into the settings window
![image](https://user-images.githubusercontent.com/10835354/175993994-667c77a8-effe-446e-afe4-fd1d87e8db60.png)

iTrooz:
Hey, simple curiosity, do you know why there isn't there a system wide proxy for github ?
I feel like this would be simpler than seeking support on a project-basis
Just a thought

shirok1:
The environment variable `CURLOPT_PROXY` doesn't work on my machine.

![image](https://user-images.githubusercontent.com/12044683/175996252-b691597d-c99f-46c0-8d90-42a85c3b4036.png)

(the right says "hex.builtin.tools.wiki_explain.invalid_response"

shirok1:
> Hey, simple curiosity, do you know why there isn't there a system wide proxy for github ? I feel like this would be simpler than seeking support on a project-basis Just a thought

The solutions I know without a router are TUN or other virtual NIC, but I failed to configure that.

shirok1:
I have thought of three ways to do this:

1. Make `hex::Net` read the setting registry and set the OPT -- breaks the "Open–Closed Principle"
2. Leave a empty virtual method for subclass to override, and call it in `Net::setCommonSettings` -- every plugin needs to change their code to use the new `NetWithProxy` class

Both are not that clean. Is there an elegant way to achieve this?

WerWolv:
I'd simply expose a static member function in the `hex::Net` class that allows setting the used proxy type, address, port and credentials and then add an interface to the settings page that simply calls that method to set the proxy globally for everybody using the Net class.

WerWolv:
[...]

PR #562, opened 2022-06-29
[Feature] Network proxy for hex::Net

shirok1:
For #557

image

PS. I ended up just reading from setting registry `"hex.builtin.setting.proxy.url"` every time.


lib/libimhex/source/helpers/net.cpp
```diff
@@ -113,6 +115,10 @@ namespace hex {
         curl_easy_setopt(this->m_ctx, CURLOPT_SSLCERTTYPE, "PEM");
         curl_easy_setopt(this->m_ctx, CURLOPT_SSL_CTX_FUNCTION, sslCtxFunction);
 #endif
+
+        std::string proxyUrl = ContentRegistry::Settings::read("hex.builtin.setting.proxy", "hex.builtin.setting.proxy.url", "");
```
WerWolv, line 119: You're accessing things that are conceptually not known by libimhex as they are created in the built-in plugin. Even if the built-in plugin technically always is available, depending on things that get created in a plugin is really bad.

That's why I suggested adding a static member function to set this value from the built-in plugin

WerWolv (changes requested):

lib/libimhex/source/helpers/net.cpp
shirok1: Hmm, I think in this way, if this setting is not set, the proxy will be set to none, which is exactly what is expected to happen without built-in plugin.

WerWolv: Of course it works fine. It just conceptually creates a dependecy of libimhex on the built-in plugin which is wrong. The built-in plugin knows of libimhex and can use it as it wishes, but libimhex knows nothing about the built-in plugin. It doesn't matter to libimhex if any plugin is being loaded or not. 

Because of this reason, libimhex should not be the one querying the information it needs from the built-in plugin but the built-in plugin should send the information to libimhex (inverting that dependency).

shirok1: I also referenced how you handle `"hex.builtin.setting.font.font_size"`, which is read from registry every time. (The opposite is `"hex.builtin.setting.interface.color"`, which is only read when Welcome Screen is init, but I think that is more related to UI refresh requirement)

shirok1: I wonder if there is a convention for initing **the** static field at startup.

WerWolv: Yeah those two settings completely go against these principles as well and I hate it. If there was a nice way to change it, I absolutely would. 

With the Net class there is an easy way to not have a depdency like that so it's best to avoid it by adding a new function to it:

```cpp
static void setProxy(const std::string &url);
```

There's really no issue with having static functions. They are a necessity and absolutely make sense in this case as it's a global setting that should apply to every single instance of the Net class. It doesn't create or expose a global static member variable at all (which I agree with you, would be bad design) but instead creates an interface to configure how the Net class should behave in the whole program.

shirok1: Oh, sorry. I didn't realize `registerSettings` is an init method, not a imgui function, until now...

shirok1:
@WerWolv

lib/libimhex/include/hex/helpers/net.hpp
```diff
@@ -62,6 +64,8 @@ namespace hex {
         std::mutex m_transmissionActive;
         float m_progress    = 0.0F;
         bool m_shouldCancel = false;
+
+        static std::string m_proxyUrl;
```
WerWolv, line 68: The convention I use for these variables is prefixing them with `s_` instead of `m_` since they're static and not a member

WerWolv (changes requested):

lib/libimhex/include/hex/helpers/net.hpp
WerWolv: Also `Net::s_proxyUrl` instead of just `s_proxyUrl`

WerWolv:
Thanks a lot for the changes. Besides that little nitpick it looks great!

shirok1:
Changed.

WerWolv:
Thanks a lot!

PR #564, opened 2022-06-30
install pattern files in the right place

iTrooz:
Currently, the pattern files are installed in /usr on Linux, where they should be in /usr/share/imhex

This PR fix it


cmake/build_helpers.cmake
```diff
@@ -70,21 +70,18 @@ macro(detectOS)
         set(CMAKE_INSTALL_BINDIR ".")
         set(CMAKE_INSTALL_LIBDIR ".")
         set(PLUGINS_INSTALL_LOCATION "plugins")
-        set(MAGIC_INSTALL_LOCATION "magic")
```
iTrooz, line 73: The MAGIC_INSTALL_LOCATION variable wasn't even used, so I removed it

PR #566, opened 2022-07-02
AUR/ArchLinux support

iTrooz:
This PR : 
- adds a version.txt file in the root that will be used by the entire workflow (including cmake) to define project version
- adds a ArchLinux variant (.pkg.tar.zst)
- updates the `imhex-bin` AUR package with this variant on release

As said on Discord, I would like this to be merged at the last moment, but feel free to review it in advance, as I shouldn't change many things now


version.txt
```diff
@@ -0,0 +1 @@
+1.18.2
```
WerWolv, line 1: Please name this file `VERSION`

dist/Arch/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+md5sums=(SKIP)
+validpgpkeys=()
+
+package() {
+    tar -xf imhex-$pkgver-ArchLinux.pkg.tar.zst
+
+
+    install -DT $srcdir/usr/bin/imhex $pkgdir/usr/bin/imhex
+    install -DT $srcdir/usr/lib/libimhex.so $pkgdir/usr/lib/libimhex.so
+	install -DT $srcdir/usr/share/imhex/plugins/builtin.hexplug $pkgdir/usr/share/imhex/plugins/builtin.hexplug
```
WerWolv, line 31: Formatting here looks a bit wrong. Also, is it possible to use a glob here for everything in the `plugins` folder with the `.hexplug` ending? Just in case there will be more plugins in the future

.github/workflows/release.yml
```diff
@@ -54,3 +64,20 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           files: '*'
+
+      - name: Prepare PKGBUILD
```
WerWolv, line 68: Not enough emojis!

.github/workflows/build.yml
```diff
@@ -284,3 +284,86 @@ jobs:
           name: Linux AppImage
           path: '*.AppImage'
 
+  archlinux-build:
+    name: ArchLinux Build
```
WerWolv, line 288: I'd name this something like `🐧 ArchLinux <Version>` so it fits with the rest of the names

.github/workflows/build.yml
```diff
@@ -284,3 +284,86 @@ jobs:
...
+    runs-on: ubuntu-latest
+    
+    container:
+      image: archlinux:base-devel
+
+    steps:
+      - name: Update all packages and install necessary packages
```
WerWolv, line 295: `Install setup dependencies` might be better

.github/workflows/build.yml
```diff
@@ -284,3 +284,86 @@ jobs:
...
+        run: |
+          echo "version=`cat version.txt`" >> $GITHUB_ENV
+      
+      - name: Prepare PKGBUILD
+        run: |
+          cp dist/Arch/PKGBUILD build
+          sed -i 's/%version%/${{env.version}}/g' build/PKGBUILD
+
+    # makepkg doesn't want to run as root, so I had to chmod 777 all over
+      - name: Package imhex-archlinux.pkg.tar.zst
```
WerWolv, line 347: `Package ArchLinux .pkg.tar.zst`

.github/workflows/build.yml
```diff
@@ -284,3 +284,86 @@ jobs:
...
+          
+          chmod -R 777 .
+          
+          sudo -u nobody makepkg
+
+          # Remplace the old file
+          rm imhex-${{env.version}}-ArchLinux.pkg.tar.zst
+          mv *.pkg.tar.zst imhex-${{env.version}}-ArchLinux.pkg.tar.zst
+      
+      - name: Upload imhex-archlinux.pkg.tar.zst
```
WerWolv, line 363: `Upload ArchLinux .pkg.tar.zst`

.github/workflows/build.yml
WerWolv, line 295: Also you need more emojis here :P

WerWolv (changes requested):

.github/workflows/build.yml
iTrooz: I'm not sure Arch has versions, so I'm just going to put `🐧 ArchLinux`

.github/workflows/build.yml
iTrooz: I separated them into two steps (update all packages & install setup dependencies) to make it explicit that all packages are being upgraded

WerWolv (approved):

PR #577, opened 2022-07-08
detect system theme with xdg portal

iTrooz:
The implementation is still awful (popen) but I honestly didn't want to spend much time make a clean implementation, I just wanted to make a quickfix to make it use a standard (org.freedesktop.appearance color-theme)

I understand this is still ugly, and would understand if you rejected the merge


iTrooz:
I tested it with ArchLinux KDE, and I'm going to try with XFCE and Gnome

iTrooz:
Oops, I just realised this was only going to bepresent in newer DE versions
I don't know what to do about this PR, I will probably keep it here for now, and maybe close it later when I have a more clear idea of what to do

iTrooz:
Closed (at least for now) because I realised this would be a bit harder than thought, and honestly I don't want to do it for now

PR #578, opened 2022-07-10
fix: setting folder tab crash

PredatorCZ:
For some reason json treats unicode string as an array.
Not sure if this a best way, but better than crashing.

edit:

Had to do refactor, user folders didn't load at startup

### DIFF
--- a/plugins/builtin/source/content/settings_entries.cpp
+++ b/plugins/builtin/source/content/settings_entries.cpp
@@ -13,6 +13,19 @@
 
 #include <nlohmann/json.hpp>
 
+namespace {
+    static std::vector<std::fs::path> userFolders;
+    static void loadUserFoldersFromSetting(nlohmann::json &setting) {
+        userFolders.clear();
+        std::vector<std::string> paths = setting;
+        for (const auto &path : paths) {
+            // JSON reads char8_t as array, char8_t is not supported as of now
+            std::u8string_view uString(reinterpret_cast<const char8_t *>(&path.front()), reinterpret_cast<const char8_t *>(std::next(&path.back())));
+            userFolders.emplace_back(uString);
+        }
+    }
+};
+
 namespace hex::plugin::builtin {
 
     void registerSettings() {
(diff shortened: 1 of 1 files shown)


plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -14,6 +14,25 @@
...
 
 namespace hex::plugin::builtin {
+    namespace folders {
+        static std::vector<std::fs::path> folders;
+
+        void loadFromSetting(nlohmann::json &setting) {
+            folders.clear();
+            std::vector<std::string> paths = setting;
+            for (const auto &path : paths)
+                folders.emplace_back(path);
```
WerWolv, line 24: This is not going to work. On Windows, paths are all UTF-16. This means you're going to load UTF-8 values from the settings file and std::fs::path will then interpret that UTF-8 data as ASCII which will cause the value to become completely currupted.

What needs to be done instead is:
- Load a std::string from the JSON, this will be UTF-8.
- Copy the data from that std::string into a std::u8string
- Store that value in a std::fs::path. That way conversion will be done correctly.

You cannot convert from a std::fs::path to a std::string and back to a std::fs::path without causing these issues. The in-between string needs to be a std::u8string.

plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -14,6 +14,25 @@
 #include <nlohmann/json.hpp>
 
 namespace hex::plugin::builtin {
+    namespace folders {
```
WerWolv, line 17: Personally, I'd just make all these functions local lambdas instead

WerWolv (changes requested):

plugins/builtin/source/content/settings_entries.cpp
PredatorCZ: thx, forgot about windows

plugins/builtin/source/content/settings_entries.cpp
PredatorCZ: returned saver as lambda, but loader is needed in global scope

PR #579, opened 2022-07-11
fix: invariables not working, add recusive pattern scan

PredatorCZ:
A few things.

1. Added recursive pattern scan for mime lookup
2. In variable weren't working/crashing because of cleaning before eval, also in-vars disappeared after eval from settings.
3. Removed some dead code.

### DIFF
--- a/plugins/builtin/source/content/views/view_pattern_editor.cpp
+++ b/plugins/builtin/source/content/views/view_pattern_editor.cpp
@@ -93,9 +93,7 @@ namespace hex::plugin::builtin {
         });
 
         EventManager::subscribe<RequestSetPatternLanguageCode>(this, [this](const std::string &code) {
-            this->m_textEditor.SelectAll();
-            this->m_textEditor.Delete();
-            this->m_textEditor.InsertText(code);
+            this->m_textEditor.SetText(code);
         });
 
         EventManager::subscribe<EventFileLoaded>(this, [this](const std::fs::path &path) {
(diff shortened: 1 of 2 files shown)


plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -716,9 +714,6 @@ namespace hex::plugin::builtin {
         std::thread([this, code] {
             auto ast = this->m_parserRuntime->parseString(code);
 
-            this->m_patternVariables.clear();
```
WerWolv, line 719: This will cause old and removed In/Out values to not be removed anymore from the list and new ones just being added on top of it.

How are In/Out values not working for you? They seem to work fine here

WerWolv (changes requested):

plugins/builtin/source/content/views/view_pattern_editor.cpp
WerWolv: Okay nevermind, something is broken. But this still feels like it makes it worse

WerWolv: Okay, the issue is in fact not the clearing there, that is necessary for the above mentioned reasons, the issue is that we call `parsePatterns` with an empty string in the `clearPatterns` function which is called at the start of `evaluatePatterns`. Removing this call fixes the issue.

PredatorCZ: I'll see what can I do about that

WerWolv (approved):

WerWolv:
Thank you!

PR #581, opened 2022-07-15
Add ability to paste hex bytes with 0x

Diadlo:
Sometimes you may have string in the following format: `0xaa 0x1 0xbb 0x2`. It's useful to have an ability to paste them as is, without editing

### DIFF
--- a/plugins/builtin/source/content/views/view_hex_editor.cpp
+++ b/plugins/builtin/source/content/views/view_hex_editor.cpp
@@ -1158,26 +1158,38 @@ namespace hex::plugin::builtin {
         ImGui::SetClipboardText(str.c_str());
     }
 
+    static std::optional<std::vector<u8>> parseWholeHexBytesWithoutPrefix(const std::string &input) {
+        std::string str = input;
+        // Check for non-hex characters
+        bool isValidHexString = std::find_if(str.begin(), str.end(), [](char c) {
+            return !std::isxdigit(c) && !std::isspace(c);
+        }) == str.end();
+
+        if (!isValidHexString)
+            return std::nullopt;
+
+        // Remove all whitespace
+        str.erase(std::remove_if(str.begin(), str.end(), [](char c) { return std::isspace(c); }), str.end());
+
+        // Only paste whole bytes
+        if (str.length() % 2 != 0)
+            return std::nullopt;
+
+        // Convert hex string to bytes
+        std::vector<u8> buffer = crypt::decode16(str);
+        return std::optional<std::vector<u8> > {buffer};
+    }
+
     static void pasteBytes(const Region &selection) {
         auto provider = ImHexApi::Provider::get();
 
         std::string clipboard = ImGui::GetClipboardText();
 
-        // Check for non-hex characters
-        bool isValidHexString = std::find_if(clipboard.begin(), clipboard.end(), [](char c) {
-            return !std::isxdigit(c) && !std::isspace(c);
-        }) == clipboard.end();
+        auto parseRes = parseWholeHexBytesWithoutPrefix(clipboard);
+        if (!parseRes.has_value())
+            return;
 
-        if (!isValidHexString) return;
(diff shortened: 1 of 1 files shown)


plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -5,13 +5,15 @@
...
 #include <hex/helpers/crypto.hpp>
+#include <hex/helpers/logger.hpp>
 
 #include "math_evaluator.hpp"
 
 #include <imgui_internal.h>
 #include <nlohmann/json.hpp>
 
 #include <thread>
+#include <sstream>
```
WerWolv, line 16: There's not really a reason to use stringstream, right?

Please avoid it if you can as it's absolutely gigantic and slows down compile times even more. As far as I can tell, you can get away by using std::string and hex::format

WerWolv (changes requested):
Looks like there's an issue with the submodules that got downgraded for some reason

Diadlo:
Fixed
But to be honest, I didn't test it after that. That's because after dropping clang support it ImHex uncompilable on my Mac.

1. It doesn't check if compiler is GCC and silently stop working with clang
2. After installing and specifying GCC12 as a compiler I have a lot of issues in some macOS internals

FYI [error.log](https://github.com/WerWolv/ImHex/files/9386273/error.log)

WerWolv:
@Diadlo Thanks for the changes!

Your compile issues seem to mostly stem from objc files being compiled with gcc.
We only compile C and C++ code with gcc, everything else that's apple specific still is using clang. 

This here is the build command the CI uses which works fine there:
```
CC=$(brew --prefix gcc@12)/bin/gcc-12      \
CXX=$(brew --prefix gcc@12)/bin/g++-12  \
OBJC=$(brew --prefix llvm)/bin/clang           \
OBJCXX=$(brew --prefix llvm)/bin/clang++  \
cmake .. 
```
Basically we tell cmake to use gcc as C and C++ compiler and llvm clang as ObjC and ObjC++ compiler. It should work fine after that. If you need any more help with getting it to work, feel free to ask!

Diadlo:
@WerWolv It works. Thanks!

Now PR is tested and one issue is fixed

Diadlo:
@WerWolv Is there anything else that should be done?

plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -4,13 +4,15 @@
...
```
WerWolv, line 15: Remove unused sstream and logger include

plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -1158,26 +1160,99 @@ namespace hex::plugin::builtin {
...
+        const char *ptr = hexByte.c_str() + 2; // Skip 0x
+        const size_t size = hexByte.size() - 2;
+        return std::string_view(ptr, size);
+    }
+
+    static std::optional<std::vector<u8>> parseHexBytesWithPrefix(const std::string &input) {
+        std::vector<std::string> hexBytes;
+        std::string currentHexByte;
+
+        for (const char &c: input) {
```
WerWolv, line 1198: What exactly does this code do? To me it looks like it's splitting the input at spaces so it could easily be replaced with the `hex::splitString` function. However, is this really a good idea? What about byte strings that look like this: `AABBCCDD`. These should be pastable as well as `AA BB CC DD` or `0xAA, 0xBB, 0xCC, 0xDD` etc. 

I think it would be better to first remove all occurrences of `0x`, `0X`, commas and whitespace characters (space, tab, newline etc) then check if there's only valid hex characters in there and at the end parse that data which should always be in the form of `AABBCCDD` then

WerWolv (changes requested):

aboooody16s (approved):

PR #586, opened 2022-07-20
Fix "Content Store" persist in flatpak.

catsout:
fix not saving to "XDG_DATA_HOME", when "XDG_DATA_DIRS" is available.   
In flatpak, downloaded `libraries/patterns` will be saved to `/run/host/user-share/imhex`, which will be cleared when app exit.  

> XDG specification specifies how to find config and data directories on linux systems. Specifically, it says this:
> - Data should be written to $XDG_DATA_HOME
> - Config should be written to $XDG_CONFIG_HOME
> - Data should be read from $XDG_DATA_HOME:$XDG_DATA_DIRS
> - Config should be read from $XDG_CONFIG_HOME:$XDG_CONFIG_DIRS


WerWolv:
Thank you!

PR #593, opened 2022-07-27
add zh-tw translation

SiderealArt:


plugins/builtin/source/lang/zh_TW.cpp
```diff
@@ -0,0 +1,761 @@
+#include <hex/api/content_registry.hpp>
+#include <hex/api/localization.hpp>
+
+namespace hex::plugin::builtin {
+
+    void registerLanguageZhTW() {
+        ContentRegistry::Language::registerLanguage("Chinese (Traditional)", "zh-TW");
+        LangEntry::setFallbackLanguage("en-US");
```
WerWolv, line 8: This can be removed, it only needs to be set once

plugins/builtin/source/lang/zh_TW.cpp
```diff
@@ -0,0 +1,761 @@
...
+                    { "hex.builtin.view.hashes.function", "雜湊函式" },
+                    { "hex.builtin.view.hashes.name", "名稱" },
+                    { "hex.builtin.view.hashes.type", "類型" },
+                    { "hex.builtin.view.hashes.result", "結果" },
+                    { "hex.builtin.view.hashes.remove", "移除雜湊" },
+                    { "hex.builtin.view.hashes.hover_info", "懸停在十六進位編輯器的選取範圍上，並按住 Shift 以查看該區域的雜湊。" },
+
+                { "hex.builtin.view.help.name", "幫助" },
+                    { "hex.builtin.view.help.about.name", "關於" },
+                        { "hex.builtin.view.help.about.translator", "由 WerWolv 翻譯" },
```
WerWolv, line 232: I don't think I was the one translating ImHex to chinese :P

WerWolv (changes requested):

plugins/builtin/source/lang/zh_TW.cpp
SiderealArt: Fixed! 🤣

WerWolv (approved):

PR #602, opened 2022-07-30
Updated get_deps_debian.sh for proper permissions

Megachar0x01:


jonathanspw:
This is not necessary for the CI runs as they run as root already.

WerWolv:
Thanks for the PR!
I'm usually not a big fan of using sudo in scripts though. People using the scripts should be aware that they will be running things as root (and the implications of that) by using sudo themselves.

PR #608, opened 2022-07-30
Fedora packaging/CI

jonathanspw:
Please for the love of all things holy squash the commits here when merging, it's a mess.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -390,17 +390,19 @@ jobs:
 
       - name: ⬇️ Install setup dependencies
         run: |
-          dnf install -y \
-          ccache         \
-          fmt-devel      \
-          git            \
-          json-devel     \
-          libcurl-devel  \
-          llvm-devel     \
-          mbedtls-devel  \
-          mock           \
-          python3-devel  \
-          rpm-build      \
+          dnf install -y     \
+          capstone-devel     \
+          ccache             \
+          desktop-file-utils \
+          fmt-devel          \
+          git                \
+          json-devel         \
+          libcurl-devel      \
+          llvm-devel         \
+          mbedtls-devel      \
+          mock               \
+          python3-devel      \
+          rpm-build          \
           yara-devel
           
       - name: 🧰 Checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -487,32 +487,18 @@ jobs:
           key: fedora-stable-${{ secrets.CACHE_VERSION }}-build-${{ github.run_id }}
           restore-keys: fedora-stable-${{ secrets.CACHE_VERSION }}-build
         
-      - name: 📜 Restore CMakeCache
-        uses: actions/cache@v3
-        with:
-          path: |
-            build/CMakeCache.txt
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -376,3 +376,129 @@ jobs:
...
+      image: fedora:rawhide
+
+    steps:
+      - name: ⬇️ Update all packages
+        run: |
+          dnf upgrade -y
+
+      - name: ⬇️ Install setup dependencies
+        run: |
+          dnf install -y     \
```
WerWolv, line 393: These things should probably be part of `get_deps_fedora.sh`. Since they're needed for building anyways

.github/workflows/build.yml
```diff
@@ -376,3 +376,129 @@ jobs:
...
+      - name: 📦 Build RPM
+        run: |
+          rpmbuild -ba --define "_version ${{env.IMHEX_VERSION}}" --define "_src_path $GITHUB_WORKSPACE" $GITHUB_WORKSPACE/dist/rpm/imhex.spec
+          
+      - name: ⬆️ Upload RPM
+        uses: actions/upload-artifact@v3
+        with:
+          name: Fedora Rawhide RPM
+          path: |
+            ~/rpmbuild/RPMS/*/*.rpm
```
WerWolv, line 445: Please update the release.yml CI script as well so it can add the .rpms to the latest release as well

dist/rpm/imhex.spec
```diff
@@ -0,0 +1,84 @@
+# ftbfs without this
+%global _lto_cflags %{nil}
+
+Name:           ImHex
+Version:        %{_version}
+Release:        1%{?dist}
+Summary:        A hex editor for reverse engineers and programmers
+
+License:        GPL-2.0-or-later
```
WerWolv, line 9: License is GPL-2.0, later versions are not included

WerWolv (changes requested):

.github/workflows/build.yml
jonathanspw: If you build on a normal Fedora install these things are not needed or are already present.  Some of them are only required because the Fedora docker image is minimal and is missing some of the packages, and some others are only needed because the packages here are built as close to how Fedora official packages will be built using system libs where possible.

People who just download and follow your compilation instructions won't need any of this except git which is actually already in the `get_deps_fedora.sh` file so I'll remove it from here.

dist/rpm/imhex.spec
jonathanspw: Sorry about that.  We just moved to SPDX standardized license naming in packages at Fedora so I grabbed the proper format from there for GPL2 not realizing there were two versions of it in SPDX.

.github/workflows/build.yml
jonathanspw: I think https://github.com/WerWolv/ImHex/pull/608/commits/a76d74a9ccffd6abac5258b972dd2983ddbc6b0a will do what's needed?

I'm not terribly familiar with GH CI.

WerWolv (approved):

PR #613, opened 2022-07-31
Fix CI for Fedora + some other things

iTrooz:


.github/workflows/build.yml
```diff
@@ -376,12 +376,21 @@ jobs:
...
-  fedora-build-rawhide:
-    name: 🐧 Fedora Rawhide
+  fedora-build:
+    strategy:
+      matrix:
+        include:
+          - docker_image: fedora:latest
+            release: Latest
+          - docker_image: fedora:rawhide
+            release: RawHide
```
jonathanspw, line 386: "H" in "Rawhide" should be lowercase.  Seems silly but branding is important :)

iTrooz:
Note : I didn't make too much effort with the commit names because they usually get squashed

iTrooz:
Fixed the submodule problem

.github/workflows/build.yml
```diff
@@ -376,12 +376,21 @@ jobs:
...
-  fedora-build-rawhide:
-    name: 🐧 Fedora Rawhide
+  fedora-build:
+    strategy:
+      matrix:
+        include:
+          - docker_image: fedora:latest
+            release: Latest
+          - docker_image: fedora:rawhide
+            release: rawHide
```
jonathanspw, line 386: You made the R lowercase.  It should be "Rawhide".

jonathanspw (changes requested):

iTrooz:
I think this is ready to be merged !

This PR fix the Fedora CI build in various ways discussed on discord, and strips the binaries for all plateforms (I don't know how to test if this works for CPack tho)

iTrooz:
I didn't know if MacOS had soversions so I only added them to Linux

WerWolv:
Both Arch and Fedora don't seem to find libimhex anymore now. You probably forget to change the path somewhere

iTrooz:
Okay, this time it should be good !

Issue #605, opened 2022-07-30
[Feature] Display LEB128 info in data inspector


gudzpoz:
### What feature would you like to see?

Display values for LEB128-encoded integers in the data inspector.

A PoC is available [here](https://github.com/gudzpoz/ImHex/commit/9a5916151edb99f335e103a2eb764ce7541fde96).

### How will this feature be useful to you and others?

LEB128 is used [in DRAWF files](https://en.wikipedia.org/wiki/DWARF), [in Dalvik Executable format](https://source.android.com/devices/tech/dalvik/dex-format#leb128) and [in Protobuf for unsigned integers](https://developers.google.com/protocol-buffers/docs/encoding#varints). It would be really convenient to have the editor translate the encoded bytes to the actual values for us.

### Request Type

- [X] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

WerWolv:
Hi
That would definitely be a cool thing to have. Please submit your PoC as a PR if you can and I'll review and merge it :)

PR #615, opened 2022-07-31
feat: Added LEB128 in data inspector

gudzpoz:
This pull request adds in the ability to inspect and edit LEB128 data in the data inspector: #605


lib/libimhex/source/helpers/utils.cpp
```diff
@@ -34,6 +34,68 @@ namespace hex {
         return vector * ImHexApi::System::getGlobalScale();
     }
 
+    template<typename T>
```
WerWolv, line 37: I put these kind of things into crypto.cpp/crypto.hpp usually

plugins/builtin/source/lang/en_US.cpp
```diff
@@ -431,6 +431,8 @@ namespace hex::plugin::builtin {
                 { "hex.builtin.inspector.float", "float (32 bit)" },
                 { "hex.builtin.inspector.double", "double (64 bit)" },
                 { "hex.builtin.inspector.long_double", "long double (128 bit)" },
+                { "hex.builtin.inspector.sleb128", "Signed LEB128" },
```
WerWolv, line 434: Please add commented out placeholders for these two translations to all the other language files as well so it won't be forgotted when translations get updated

lib/libimhex/include/hex/api/content_registry.hpp
```diff
@@ -208,11 +208,13 @@ namespace hex {
                     size_t requiredSize;
                     impl::GeneratorFunction generatorFunction;
                     std::optional<impl::EditingFunction> editingFunction;
+                    size_t maxSize = 0;
```
WerWolv, line 211: This should probably be set explicitly to 0 in the `add` function instead of being done in here

WerWolv (changes requested):

WerWolv:
Thanks a lot for the PR!

plugins/builtin/source/lang/en_US.cpp
gudzpoz: Added. Thanks!

gudzpoz:
Thanks for the review! I have made modifications accordingly. Would you please review it again?

WerWolv (approved):

WerWolv:
Thanks a lot!

PR #644, opened 2022-08-03
do not iterate over XDG home dir

iTrooz:


lib/libimhex/source/helpers/fs.cpp
```diff
@@ -131,10 +131,10 @@ namespace hex::fs {
         #elif defined(OS_LINUX)
 
             auto dataDirs = xdg::DataDirs();
-            auto dataDirsHome = xdg::DataHomeDir();
 
             std::copy(dataDirs.begin(), dataDirs.end(), std::back_inserter(paths));
-            std::copy(dataDirsHome.begin(), dataDirsHome.end(), std::back_inserter(paths));
+            
+            paths.insert(paths.begin(), xdg::DataHomeDir());
```
WerWolv, line 137: You can simply move this call over the std::copy and use push_back. That should be easier

iTrooz: I'll do, I was simply trying to do something similar to before

iTrooz:
tested and ready to be merged !

PR #650, opened 2022-08-04
Some CI changes

iTrooz:
Hey ! I made some CI changes

You may not like some of them, so I invite you to take a look at the changes (the commit names are explicit)

Changes resumed :
- removed useless cmake flags (rust_path, objc compilers...)
- use as much system libraries as possible on Arch

I'd like this pull request to be merged **after** the one that changes the directory structure #642


iTrooz:
To fix this error you need to reset the cache (change CACHE_VERSION)

this is because the old build environment was generated without the dependencies in ArchLinux, and since the build env is cached..

I think cleanup up the cache from time to time can only be good anyway

iTrooz:
Oh, merging with master already  invalidated the cache

PR #654, opened 2022-08-05
Update ja_JP.cpp

Berylskid:
- Corrected some mistranslations
- Updated some outdated class names
- Added some translations
- Replaced all exclamation marks (！) with Japanese reading marks (。)
- Replaced all 3-dot readers (...) with 2-byte character versions (…)


WerWolv:
Thanks a lot!

PR #661, opened 2022-08-07
build: Update nfd-extended to get correct feature test detection

btzy:
This update to NFD Extended fixes the deployment target detection so we no longer need to pass the `NFD_USE_ALLOWEDCONTENTTYPES` flag manually.  NFD Extended now detects the correct deployment target that was set by the `MACOSX_DEPLOYMENT_TARGET` export.


WerWolv:
Thanks a lot!

PR #669, opened 2022-08-12
Explicitly connect to desired signals, rather than every one.

eoineoineoin:
This avoids a crash on POSIX after running a command from the command
pallete; on completion, a SIGCHILD will be raised, which was incorrectly
calling the error handler and terminating ImHex.


WerWolv:
Thank you!

PR #672, opened 2022-08-13
Small fixes for Fedora

iTrooz:
This PR adds very small fixes directed at Fedora :
- use 'imhex' as package name in imhex.spec, so it's the same name as in the official Fedora repository (it's weird to have "imhex" in the official repository and "ImHex" with the github rpm, plus I think this allowed both packages to be installed at the same time, which I don't think is a good thing)
- Remove CMakeCache cache in the CI : it doesn't work as of now, and after thinking about it, CMake is handled by rpmbuild and not us, which I think makes it risky to try to manage it (by caching it)


WerWolv:
Thank you!

PR #677, opened 2022-08-16
Update CMakeLists.txt minimum required cmake version

vtronko:
As per cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html,
`set(CMAKE_CXX_STANDARD 23)` requires minimum cmake version bump to 3.20


WerWolv:
Thank you!

PR #689, opened 2022-08-18
Fix "build twice" bug in the CI

iTrooz:
The pull request to change the install directory structure/add support for Fedora introduced a bug in the CI, essentially caused by the fact that the deb and AppImage version shared the same build directory (I was updating the build directory of the deb with a flag for the appimage).
This caused the cache to not be taken into account for the appimage build (I don't know why), thus rebuilding everything


WerWolv:
Thank you!

PR #696, opened 2022-08-19
fix: crash when evaluating patterns

PredatorCZ:
App crashed when pattern was evaluating caused by race condition resulting in SEGV.
Fix might be a bit crude.


WerWolv:
Thank you!

PR #716, opened 2022-08-31
add NoGPU (Software Rendering) version for Windows

iTrooz:


iTrooz:
Should be ready to be merged (tested on a Win11 VM)

.github/workflows/build.yml
```diff
@@ -101,19 +101,35 @@ jobs:
...
+          build/*.msi
+
     - name: ⬆️ Upload Portable ZIP
       uses: actions/upload-artifact@v3
       with:
         name: Windows Portable
         path: |
           build/install/*
+      
+    - name: Download Mesa3D for NoGPU version
```
WerWolv, line 118: Missing emoji.
Maybe something like `⬇️ Download NoGPU dependencies` so we can call it the same on all OSes

.github/workflows/release.yml
```diff
@@ -59,7 +59,9 @@ jobs:
           done
 
       - name: 🟩 Rename Windows Portable Zip
-        run: mv "Windows Portable.zip" imhex-${{env.IMHEX_VERSION}}-Windows-Portable.zip
+        run: |
+          mv "Windows Portable.zip" imhex-${{env.IMHEX_VERSION}}-Windows-Portable.zip
+          mv "Windows Portable NoGPU.zip" imhex-${{env.IMHEX_VERSION}}-Windows-Portable.zip
```
WerWolv, line 64: NoGPU is missing in the output name here

WerWolv (changes requested):

.github/workflows/release.yml
iTrooz: Oops my mistake

.github/workflows/build.yml
iTrooz: If you don't mind, I'd like to keep it like that, so we can know explicitly what this step does just by glancing
+ it would be difficult to use that step name for MacOS since there are two steps ("Checkout glfw" and "Patch and install custom glfw")
(And Linux doesn't need such a step because the normal version can already run in software rendering, at least I was able to run it on VMs)

It's your call tho

WerWolv (approved):

PR #733, opened 2022-09-07
updates to the AppImage package

iTrooz:
This adds the following :
- add update information to the AppImage/add zsync file to the release to allow AppImage users to perform upgrades with differential download
- modify the metainfo file according to what was said on discord
- include .metainfo.xml (new standard) and .appdata.xml (old standard) in all Linux packages.
The old standard is included as a symlink to the new one, because of https://github.com/AppImage/appimage.github.io/issues/2978, but also because in general some applications may want the old standard, so better include both of them since it doesn't cost us much
- Fix the release CI to not do some steps without a release token


iTrooz:
I just realized I didn't rename the metainfo file correctly (should be net.werwolv.imhex.metainfo.xml), I'm going to fix this ASAP

WerWolv:
Ohh really cool
I read about zsync but never figured out how to properly do it. Thank you :)

iTrooz:
Seems good :D

WerWolv:
Thank you!

PR #736, opened 2022-09-08
Update README.md

cameron-keene:
Added table of contents to readme and information about the LICENSE

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,31 @@
   <a title="Total Downloads" href="https://github.com/WerWolv/ImHex/releases/latest"><img alt="Total Downloads" src="https://img.shields.io/github/downloads/WerWolv/ImHex/total?longCache=true&style=for-the-badge&label=Downloads&logoColor=fff&logo=GitHub"></a>
 </p>
 
+Table of Contents
+=================
+
+  *[Supporting](#supporting)
+  *[Screenshots](#screenshots)
+  *[Features](#features)
+  *[Pattern Language](#pattern-language)
+  *[Database](#database)
+  *[Requirements](#requirements)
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

I don't mind adding a section about the license, however the table of contents looks kinda weird in my opinion. Removing the `*` in front of every link and centering the list might help? Not sure

README.md
```diff
@@ -196,3 +221,9 @@ All dependencies that aren't bundled, can be installed using the dependency inst
 - Thanks to btzy for [nativefiledialog-extended](https://github.com/btzy/nativefiledialog-extended)
 - Thanks to danyspin97 for [xdgpp](https://sr.ht/~danyspin97/xdgpp)
 - Thanks to all other groups and organizations whose libraries are used in ImHex
+
+
+### License
+
+- ImHex is available under the GNU General Public License. Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed.
```
iTrooz, line 228: I propose to explicitly say the license is the GPL **2**

KillyMXI:
Seems like TOC formatting was messed up and it was supposed to be just another unordered list but can't be parsed as such by GitHub.
In that case I would use dashes for consistency (and fix spacing of course).

Compact inline TOC might be not a bad choice for a sigle-level list (indented entries would have to go if it were up to me) but that would require more deliberate formatting compared to what is provided in the PR.

WerWolv:
Sorry for ignoring this for so long. Closing it though since I've since refactored the readme and moved this information to the INSTALL.md file (including some of the proposed changes from you). Thank you though!

PR #737, opened 2022-09-09
Check ImHex-Patterns directory when IMHEX_OFFLINE_BUILD is set

ColinKinloch:
This allows the patterns to be bundled with flatpak version https://github.com/flathub/net.werwolv.ImHex/pull/24

### DIFF
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -392,6 +392,12 @@ function(downloadImHexPatternsFiles dest)
 
         FetchContent_Populate(imhex_patterns)
 
+    else ()
+        # Maybe patterns are cloned to a subdirectory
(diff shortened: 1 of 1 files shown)


cmake/build_helpers.cmake
```diff
@@ -392,6 +392,12 @@ function(downloadImHexPatternsFiles dest)
 
         FetchContent_Populate(imhex_patterns)
 
+    else ()
+        # Maybe patterns are cloned to a subdirectory
+        set(imhex_patterns_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/imhex-patterns")
```
WerWolv, line 397: This should probably be `/ImHex-Patterns` as that's the name of the repo (and therefor the default folder name when cloning)

WerWolv:
Thank you!

cmake/build_helpers.cmake
ColinKinloch: Good point.

PR #739, opened 2022-09-11
feat(i18n): update Chinese(Simplified) translation

xtexChooser:
1. Add new translations
2. Use full-width brackets, because the width of menus can be increased automatically
3. Add spaces between characters and numbers
4. Revert translation of `half double`, because it is a kind of primitive type and should be synchronized with other types

Thanks,
-- xtex


WerWolv:
Thank you!

Issue #760, opened 2022-10-02
[Bug] Bad std::optional access when inspectors.hexpat is not found


markhc:
### Operating System

Linux

### What's the issue you encountered?

When opening a file through ImHex, if inspectors.hexpat is not found on the config folder, `ViewDataInspector::drawContent` will throw an exception on `auto error = runtime.getError().value();` and crash the application.

This is because ` runtime.getError()` returns an std::optional, but that optional for some reason does not have a value on this case, so `.value()` throws a `std::bad_optional_access` exception.

This was introduced on commit [3367237da3805220eeea44893c09f67f495b577c](https://github.com/WerWolv/ImHex/commit/3367237da3805220eeea44893c09f67f495b577c)

### How can the issue be reproduced?

1. Build ImHex from source (commit 3367237da3805220eeea44893c09f67f495b577c)
2. Make sure "inspectors.hexpat" is not on the config folder
3. Try to open a file
4. Watch it crash :(

### ImHex Version

1.23.2-3367237da3805220eeea44893c09f67f495b577c

### ImHex Build Type

- [X] Nightly or built from sources

### Additional context?

This was tested on Arch Linux with a clean build from source with latest commit being 3367237da3805220eeea44893c09f67f495b577c.

markhc:
A simple fix would be to check if the error has a value (i.e an exception) before trying to retrieve it.

e.g 

```
const auto& error = runtime.getError();

if (error.has_value()) {
    log::error("Failed to execute inspectors.hexpat:\n {}", error.value().what());
} else {
    log::error("Failed to execute inspectors.hexpat");
}```

Though since I am unfamiliar with what "inspectors.hexpat" is supposed to do, i don't know if that's the best solution.

PR #761, opened 2022-10-02
Fix: bad optional access when inspectors.hexpat is missing

markhc:
Fixes the issue discussed on #760


WerWolv:
Nice thanks for the quick fix!

PR #767, opened 2022-10-02
Fix deprecated MAINTAINER flag.

Mou-Ikkai:
According to DockerFile reference: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

The MAINTAINER flag is now deprecated, as such I turned it into a label flag as shown in https://docs.docker.com/engine/reference/builder/#label


WerWolv:
Thank you!

PR #775, opened 2022-10-10
CI: Use action to manage ccache + minor fixes

iTrooz:
This PR uses a cache action to manage ccache. This allows to easily set a cache size limit (I think this is why builds are taking so long right now), and print statistics after the run to show if the cache worked

Other minor changes :
- Remove Rust and libpython from the CI, as discussed on discord
- Fix the cache key name for MacOS, this caused the normal/NoGPU build caches to mix up, and slow down the build
- Do not pin appimage-builder version anymore as the issue we faced with it was resolved in the latest release

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -25,14 +25,13 @@ jobs:
       with:
         languages: 'cpp'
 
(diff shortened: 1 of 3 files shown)


WerWolv:
Thank you!

PR #787, opened 2022-10-17
add Ctrl+Shift+Z shortcut to redo

iTrooz:
On some editors, the redo shortcut is mapped to Ctrl+Shift+Z

I think it would be great to support both shortcuts (the other one being Ctrl+Y)


WerWolv:
Thanks!

PR #789, opened 2022-10-19
Minor additions

iTrooz:
The commit names are self-explanatory

I really would like a code review ^^

### DIFF
--- a/lib/libimhex/source/helpers/net.cpp
+++ b/lib/libimhex/source/helpers/net.cpp
@@ -120,8 +120,11 @@ namespace hex {
 
     std::optional<i32> Net::execute() {
         CURLcode result = curl_easy_perform(this->m_ctx);
-        if (result != CURLE_OK)
-            log::error("Net request failed with error {0}: '{1}'", u32(result), curl_easy_strerror(result));
+        if (result != CURLE_OK){
+            char *url = NULL;
+            curl_easy_getinfo(this->m_ctx, CURLINFO_EFFECTIVE_URL, &url);
+            log::error("Net request '{0}' failed with error {1}: '{2}'", url, u32(result), curl_easy_strerror(result));
+        }
 
         long responseCode = 0;
         curl_easy_getinfo(this->m_ctx, CURLINFO_RESPONSE_CODE, &responseCode);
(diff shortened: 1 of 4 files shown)


lib/libimhex/source/helpers/net.cpp
```diff
@@ -120,8 +120,11 @@ namespace hex {
 
     std::optional<i32> Net::execute() {
         CURLcode result = curl_easy_perform(this->m_ctx);
-        if (result != CURLE_OK)
-            log::error("Net request failed with error {0}: '{1}'", u32(result), curl_easy_strerror(result));
+        if (result != CURLE_OK){
+            char *url = NULL;
```
WerWolv, line 124: C++ has a nice keyword called `nullptr` that should be used instead of `NULL`

main/source/init/splash_window.cpp
```diff
@@ -51,8 +52,14 @@ namespace hex::init {
                         this->m_currTaskName = name;
                     }
 
+                    using namespace std::chrono;
```
WerWolv, line 55: Using namespace is very frouned upon since it dumps the entire content of the namespace into your current namespace. I don't think it's needed here really.

I use it above for chrono_literals because you can't use user defined literals (Such as writing `100ms` to refer to `std::chrono::milliseconds(100)`) without a using namespace. But that only dumps like 5 functions into your namespace instead of the entirety of the chrono library

WerWolv (changes requested):

main/source/init/splash_window.cpp
```diff
@@ -51,8 +52,14 @@ namespace hex::init {
...
+
+                    auto startTime = high_resolution_clock::now();
                     if (!task())
                         status = false;
+                    auto endTime = high_resolution_clock::now();
+
+                    log::info("Task {} done in {}ms", name, duration_cast<milliseconds>(endTime-startTime).count());
```
WerWolv, line 62: `"Task {} finished in {} ms"` sounds better in my opinion

WerWolv (changes requested):

WerWolv:
Thank you!

main/source/init/splash_window.cpp
iTrooz: > I use it above for chrono_literals

Oops, how did I not see that

In this case, should I use `std::chrono::` everywhere ?

WerWolv: Yeah that's usually better. There's really only very very few good usecases where `using namespace` makes your code better. For `std` types, you should generally avoid it.

WerWolv (approved):

WerWolv:
Thank you!

Issue #794, opened 2022-10-23
Remove/Limit net rquests on startup


iTrooz:
ImHex currently makes (AFAIK) 3 requests on startup :

- https://api.werwolv.net/imhex/tip : Could we embed a file with the tips as a resource ? 
- https://api.github.com/repos/WerWolv/ImHex/releases/latest : Could we make a flag to disable update checks (for versions used by package managers), and else ask users if they wish to check for updates on first startup ?
- https://api.werwolv.net/imhex/store : We could make the request when the user clicks on the "content store" button

I think these requests should we removed if possible because :
- we usually do not have the consent on the user for them
- they slow down the startup

NOTE : This issue is mostly a way for me to reference all issues/PRs related to this

iTrooz:
I can confirm that with these PRs, we can configure ImHex to not send any request on startup !

PR #797, opened 2022-10-23
Store tips locally

iTrooz:
This PR aims to store the tips locally in ImHex (using libromfs) instead of making a net request at launch

I've noticed that even if my code is in the builtin plugin, I need to put the tips file in `resources/romfs` and not `plugins/builtin/romfs`. Is this normal ?


iTrooz:
I didn't know if you wanted to keep the same behavior as the website API (a tip per day) so I did it, tell me if you want to remove it (I just have to remove the last commit)

iTrooz:
https://github.com/WerWolv/ImHex/issues/794

WerWolv:
> I've noticed that even if my code is in the builtin plugin, I need to put the tips file in resources/romfs and not plugins/builtin/romfs. Is this normal ?

This is definitely not right and works fine when I build it here.
There was a report of another person though (also building locally on Arch) that didn't get logos and icons displayed that I put in the built-in plugin romfs. I think some compiler flags on Arch make the linker think that the two romfs libraries are the same and it just yeets one of them

iTrooz:
Ubuntu also wants the file to be in resources/romfs so I think this is a more a Linux vs Windows issue

Testing on Windows, it indeed wants the file to be in plugins/builtin/romfs

I will investigate more

iTrooz:
This commit puts the tips in the builtin romfs, and check if the romfs data is valid before trying to load the tips

I think everything is working now

Issue #796, opened 2022-10-23
[Feature] Load unicode glyphs asynchronously (or load them only as needed ?)


iTrooz:
### What feature would you like to see?

I would like ImHex to load unneeded unicode glyphs asynchronously after startup, or load them only when they need to be displayed (for UI locales/custom encodings/language pattern output/etc...)

### How will this feature be useful to you and others?

This would greatly decrease the ImHex startup time

My timings indicate
```
[12:59:34] [INFO]  [main] Task 'Loading plugins' finished in 97 ms
[12:59:34] [INFO]  [main] Task 'Checking for updates' finished in 381 ms
[12:59:34] [INFO]  [main] Task 'Downloading information' finished in 708 ms
[12:59:39] [INFO]  [main] Task 'Loading fonts' finished in 5447 ms
```
on a computer with relatively good specs. I think we can expect the time to be even greater on a computer with low specs

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

### Other details

This issue is essentially a summary of what was decided on discord :
https://discord.com/channels/789833418631675954/789833418631675957/1032220826247888917
https://discord.com/channels/789833418631675954/789833418631675957/1033430308231577642

Implementation idea by WerWolv :
> I think the better way would be to load the minimum required glyphs at the start (maybe add the required ranges to the localization definition?) and then after that load everything else asynchronously and switch out the font atlas once loading is completed
That way we don't have a blank UI at the start and the glyphs will probably still be available before the user gets the chance to load something that uses the rest of the glyphs

This feature request is currently **not planned**. Leave a reaction if you would like us to implement it !

PR #798, opened 2022-10-24
Add a setting to disable loading of unicode characters

iTrooz:
Helps #796

### DIFF
--- a/plugins/builtin/source/lang/en_US.cpp
+++ b/plugins/builtin/source/lang/en_US.cpp
@@ -775,7 +775,7 @@ namespace hex::plugin::builtin {
                     { "hex.builtin.setting.general.show_tips", "Show tips on startup" },
                     { "hex.builtin.setting.general.auto_load_patterns", "Auto-load supported pattern" },
                     { "hex.builtin.setting.general.sync_pattern_source", "Sync pattern source code between providers" },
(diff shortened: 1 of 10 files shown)


plugins/builtin/source/lang/en_US.cpp
```diff
@@ -775,6 +775,7 @@ namespace hex::plugin::builtin {
                     { "hex.builtin.setting.general.show_tips", "Show tips on startup" },
                     { "hex.builtin.setting.general.auto_load_patterns", "Auto-load supported pattern" },
                     { "hex.builtin.setting.general.sync_pattern_source", "Sync pattern source code between providers" },
+                    { "hex.builtin.setting.general.enable_unicode", "Load unicode characters (will speed up startup but some characters may not appear)" },
```
WerWolv, line 778: This text is a bit misleading I think. What it says to me is, If I turn this setting on and have it load unicode characters, it will speed up the loading process.

I'd just call it `Load unicode characters`. There isn't much space for a comment like this anyways.

Also please add the same string to all the other language files but comment it out there.

WerWolv (changes requested):

plugins/builtin/source/lang/en_US.cpp
iTrooz: Oops yeah, I didn't think it through

iTrooz: Done !

WerWolv (approved):

PR #803, opened 2022-10-27
Ask the users if they want to check for updates

iTrooz:
The principle is that the default value is 2 ("not set") and the popup makes it switch to 1 ("check for updates") or 0 ("do not check for updates") (0 being the default is the popup is closed)

I wasn't sure how to do something cleaner

### DIFF
--- a/main/source/init/tasks.cpp
+++ b/main/source/init/tasks.cpp
@@ -25,6 +25,7 @@ namespace hex::init {
     using namespace std::literals::string_literals;
 
     static bool checkForUpdates() {
+        // documentation of the value above the setting definition
         int showCheckForUpdates = ContentRegistry::Settings::read("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 2);
         if (showCheckForUpdates == 1){
             hex::Net net;
--- a/main/source/init/tasks.cpp
+++ b/main/source/init/tasks.cpp
@@ -25,23 +25,26 @@ namespace hex::init {
     using namespace std::literals::string_literals;
 
     static bool checkForUpdates() {
-        hex::Net net;
+        int showCheckForUpdates = ContentRegistry::Settings::read("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 2);
+        if(showCheckForUpdates==1){
+            hex::Net net;
 
-        auto releases = net.getJson(GitHubApiURL + "/releases/latest"s, 2000).get();
-        if (releases.code != 200)
-            return false;
+            auto releases = net.getJson(GitHubApiURL + "/releases/latest"s, 2000).get();
+            if (releases.code != 200)
+                return false;
 
-        if (!releases.body.contains("tag_name") || !releases.body["tag_name"].is_string())
-            return false;
+            if (!releases.body.contains("tag_name") || !releases.body["tag_name"].is_string())
+                return false;
 
-        auto versionString = std::string(IMHEX_VERSION);
-        size_t versionLength = std::min(versionString.find_first_of('-'), versionString.length());
-        auto currVersion   = "v" + versionString.substr(0, versionLength);
-        auto latestVersion = releases.body["tag_name"].get<std::string_view>();
+            auto versionString = std::string(IMHEX_VERSION);
+            size_t versionLength = std::min(versionString.find_first_of('-'), versionString.length());
+            auto currVersion   = "v" + versionString.substr(0, versionLength);
+            auto latestVersion = releases.body["tag_name"].get<std::string_view>();
 
-        if (latestVersion != currVersion)
-            ImHexApi::System::impl::addInitArgument("update-available", latestVersion.data());
+            if (latestVersion != currVersion)
+                ImHexApi::System::impl::addInitArgument("update-available", latestVersion.data());
 
+        }
         return true;
     }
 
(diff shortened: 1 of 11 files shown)


iTrooz:
https://github.com/WerWolv/ImHex/issues/794

main/source/init/tasks.cpp
```diff
@@ -25,23 +25,26 @@ namespace hex::init {
     using namespace std::literals::string_literals;
 
     static bool checkForUpdates() {
-        hex::Net net;
+        int showCheckForUpdates = ContentRegistry::Settings::read("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 2);
```
WerWolv, line 28: Could you document in a comment what the different values for this setting mean? 0, 1 and 2 alone aren't really self explanatory

main/source/init/tasks.cpp
```diff
@@ -25,23 +25,26 @@ namespace hex::init {
...
+        if(showCheckForUpdates==1){
```
WerWolv, line 29: Spaces before and after operator are missing

plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -35,6 +35,17 @@ namespace hex::plugin::builtin {
 
         /* General */
 
+        ContentRegistry::Settings::add("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 2, [](auto name, nlohmann::json &setting) {
```
WerWolv, line 38: 1 and 2 will both be treated here as "Enabled". Is that the desired effect?

plugins/builtin/source/content/welcome_screen.cpp
```diff
@@ -517,6 +517,17 @@ namespace hex::plugin::builtin {
                 loadDefaultLayout();
         });
 
+        EventManager::subscribe<EventWindowInitialized>([]{
+            int showCheckForUpdates = ContentRegistry::Settings::read("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 2);
+            if(showCheckForUpdates==2){
```
WerWolv, line 522: Spaces before and after operator are missing

plugins/builtin/source/content/welcome_screen.cpp
```diff
@@ -517,6 +517,17 @@ namespace hex::plugin::builtin {
...
+                ContentRegistry::Settings::write("hex.builtin.setting.general", "hex.builtin.setting.general.check_for_updates", 0); 
+                View::showYesNoQuestionPopup("hex.builtin.welcome.check_for_updates_text"_lang, [] {
```
WerWolv, line 524: Formatting of the lambdas here is kinda weird. I'd put the head of the lambda and its content on two different lines and also move the head of the first lambda on a new line

plugins/builtin/source/lang/en_US.cpp
```diff
@@ -45,6 +45,9 @@ namespace hex::plugin::builtin {
                     { "hex.builtin.welcome.learn.plugins.link", "https://github.com/WerWolv/ImHex/wiki/Plugins-Development-Guide" },
                 { "hex.builtin.welcome.header.various", "Various" },
                 { "hex.builtin.welcome.tip_of_the_day", "Tip of the Day" },
+                { "hex.builtin.welcome.check_for_updates", "Check for updates" },
```
WerWolv, line 48: These should also be added to the other language files

WerWolv (changes requested):

WerWolv:
Thank you!

main/source/init/tasks.cpp
iTrooz: Sure ! Actually, aren't enums just named integers in C++ ? In this case, may I create an enum somewhere ? (I'm not sure where that would fit)

main/source/init/tasks.cpp
iTrooz: Oops, sorry for making you repeat yourself in this PR, I should have checked

plugins/builtin/source/content/welcome_screen.cpp
iTrooz: Like that ? (see my last commit)

main/source/init/tasks.cpp
WerWolv: Yeah an enum would be great for this too. Just requires some casting when writing it to the json object probably

WerWolv (approved):

plugins/builtin/source/content/welcome_screen.cpp
WerWolv: Yeah an enum would be great for this too. Just requires some casting when writing it to the json object probably

iTrooz: Where should I create the enum ? We would need it in `main` and `builtin`, so how about in `tasks.hpp` ?

WerWolv: Right it's split up between multiple files...
Yeah I honestly also don't know where to put it best.
For now I'd probably just put a comment above the definition of the setting.

iTrooz: Done !

WerWolv (approved):

WerWolv:
Thank you!

PR #808, opened 2022-10-30
fix: Pattern auto-loading issue on Windows

Asvel:
Both Windows API and `std::wstring` constructor use wide-char count as size parameter, so the ` * sizeof(TCHAR)` is not needed.

And the size and output content of Windows API include the terminating null character, but C++ string should not. If includes, [magic.cpp#L30](https://github.com/WerWolv/ImHex/blob/bececff9e5a9d1e5f42b3227ea9c2bb1f46c1e83/lib/libimhex/source/helpers/magic.cpp#L30) will concatenate a string like `"path1\0:path2\0:"`, and only the part before `\0` is passed to libmagic when use `.c_str()`.


WerWolv:
Thank you!

PR #810, opened 2022-10-31
Change "Esc" key behavior on hex editor popups

### LINKED ISSUE #809: [Bug] Various issues from u/egudu
### Operating System

Not known

Various issues were described by this user on https://www.reddit.com/r/netsec/comments/wbbyfn/comment/ii8g5o3, and some of them are still present (and I'm willing to fix at least some of them)

This issue will be used to reference all PRs made regarding this

Not applicable

### ImHex Version

Not known

### ImHex Build Type

### Installation type

Not known

### Additional context?

Original message :

> 
> I'm using HxD so I will give you my feedback comparing to it. Some of this might come off as mean, it's not intended to.
> 
>     Stop connecting to the internet without my permission.
> 
>     I don't want the fullscreen mode, I want to see my taskbar. Make an option for that.
> 
>     I cannot drag & drop multiple files on it.
> 
>     The dropdown list for files is annoying, I want a tab bar to quickly switch. The tab bar needs a [x] to close a tab, and a tab needs to be closed with middle-mouse click.
> 
>     File -> Close need to be keybound to Ctrl+W.
> 
> I use HxD to compare files and get their hashes.
> 
>     "Diffing" should have a button to quickly add the last two files to it (and/or do that automatically upon selecting the tab). Messing with two dropdown boxes is too much work.
> 
>     The diffing tool is completely useless. Why is there no button/keybind to go to the next difference? Why does it not go to the first difference found automatically? Why does it not tell me that two files are identical?
> 
>     I'll hash a small 2 GB file and the program freezes - heck I cannot even Alt+Tab out because it's in fullscreen mode. Why not dedicated tool to hash that also has the option to compare with a given hash and shows a progress bar like HxD? Though it's much faster than HxD, good.
> 
>     Why does F3 and Shift+F3 not continue/reverse the search results? Do you expect me to use mouse-clicks on those buttons?
> 
>     Ctrl+G (goto) cannot be closed with ESC, but Ctrl+F (search) can.
> 
> Ok that was all for the beginning.
> 
> Have not looked in other tools

iTrooz:
This PR fix the "'Esc" key on the Ctrl+G (goto) ( helps #809 )

It also closes popups directly instead of exiting the input **and then** exiting the popup (it doesn't seem to be possible to do globally per an option in imgui, see https://github.com/ocornut/imgui/issues/2620#issuecomment-501136289)

### DIFF
--- a/plugins/builtin/source/content/views/view_hex_editor.cpp
+++ b/plugins/builtin/source/content/views/view_hex_editor.cpp
@@ -44,8 +44,10 @@ namespace hex::plugin::builtin {
                     ImGui::EndTabItem();
                 }
 
-                ImGui::SetKeyboardFocusHere();
-                ImGui::SetNextFrameWantCaptureKeyboard(true);
+                if(this->m_requestFocus){
+                    ImGui::SetKeyboardFocusHere();
+                    this->m_requestFocus = false;
+                }
                 if (ImGui::InputTextIcon("##input", ICON_VS_SYMBOL_OPERATOR, this->m_input, ImGuiInputTextFlags_EnterReturnsTrue | ImGuiInputTextFlags_AutoSelectAll)) {
                     if (auto result = this->m_evaluator.evaluate(this->m_input); result.has_value()) {
                         const auto inputResult = result.value();
@@ -92,6 +94,7 @@ namespace hex::plugin::builtin {
 
         Mode m_mode = Mode::Absolute;
 
+        bool m_requestFocus = true;
         std::string m_input;
         MathEvaluator<i128> m_evaluator;
     };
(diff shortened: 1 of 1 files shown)


plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -44,8 +44,10 @@ namespace hex::plugin::builtin {
                     ImGui::EndTabItem();
                 }
 
-                ImGui::SetKeyboardFocusHere();
-                ImGui::SetNextFrameWantCaptureKeyboard(true);
```
iTrooz, line 48: I removed `ImGui::SetNextFrameWantCaptureKeyboard(true);` because I wanted to follow the way PopupFind does it

iTrooz:
I opened a discussion to explain some change, but this is ready to be merged

WerWolv:
Thank you!

PR #820, opened 2022-11-15
Tell users patterns are being downloaded

iTrooz:
I've been wondering for months why configuring CMake took so long (sometimes 30secs ?), I thought users would like to know what's happening


WerWolv:
Thank you!

PR #821, opened 2022-11-16
Tell the user if he have a proxy set

iTrooz:
Not sure if it's worth merging, but this might help users in some cases

It's your call !


lib/libimhex/source/helpers/net.cpp
```diff
@@ -124,6 +124,9 @@ namespace hex {
             char *url = nullptr;
             curl_easy_getinfo(this->m_ctx, CURLINFO_EFFECTIVE_URL, &url);
             log::error("Net request '{0}' failed with error {1}: '{2}'", url, u32(result), curl_easy_strerror(result));
+            if(!Net::s_proxyUrl.empty()){
+                log::error("Note: You appear to have a proxy set. Is it working ?");
```
WerWolv, line 128: This should probably be a Info log instead. A error log isn't really a "Note" so I'd remove that. Also, extra space before the question mark again :P

WerWolv:
Thank you!

Issue #825, opened 2022-11-21
[Feature] Quick "save project"


ssmid:
### What feature would you like to see?

Currently:
File -> Save project... -> File Dialog starting in _home directory_ not _current directory_ -> Select project file -> Save

Better:
- rename "Save project..." action to "Save project as..."
- create new "Save project" action, which remembers the current project file

### How will this feature be useful to you and others?
ImHex crashes from time to time. When this happens, it's really great to not lose hours of work in the pattern editor. A quick "Save project" action would help to prevent data loss.

### Request Type

- [x] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

ssmid:
I implemented the feature in above pull request. Would you may have a look at it? Thanks for all your work!

PR #826, opened 2022-11-21
Quick "save project" action

ssmid:
Issue: #825 

Save projects with just one hotkey or two clicks.

- now there are two actions "Save Project" and "Save Project As..." instead of one "Save Project..."

### Why:
ImHex crashes from time to time. Having a quick way to save the project (which includes the pattern editor) helps to minimize data loss.

### TODO

- [ ] Add "recent projects" on welcome page


ssmid:
Glad you like it. I really enjoy using ImHex, it makes reverse engineering a lot easier, so well done to you!
Both issues should be fixed now, however there are two concerns remaining, one major and one minor:
- When closing a file and then saving with a shortcut, everything inside the pattern editor is lost if I understand correctly
- The Alt + S shortcuts do work, but they trigger the tab bar on the top of the current widget because of Alt

ssmid:
There should probably also be "recent projects" listed in the front page.

ssmid:
Could we maybe add a warning on project save, if not all previous files are part of the new project save?

WerWolv:
I see your point. 
To properly do that though, we'd need something like a project view where all the providers that belong to that project are displayed. Closing the provider with the tabs or CTRL + W then would only close the view of the file but not the file itself, it would still be in the project view and only gets closed when explicitly removing it from the project itself.

The way it is now, there's only a popup when trying to close a file after anything in its "virtual" project file has changed. So after editing pattern code or modifying a byte

WerWolv:
> I think it does not make sense to force the providers concept onto projects

All projects need to have a provider. That's kind of the basis of everything that happens in ImHex. Even if you click on the little X at the top right of the welcome screen to close it, you actually open a "NullProvider" that just reports its size as 0 and discards all read and write requests.

Issue #832, opened 2022-11-23
[Feature] Ignore "Found more than one built-in plugin" if it's a debug build


ssmid:
### What feature would you like to see?

ImHex refuses to work if it finds multiple builtin plugins. While this is fine for the release version, this is not great when developing on it. It prevents having multiple versions of ImHex at the same time (release and debug).

### How will this feature be useful to you and others?

It would make it easier to develop ImHex.

### Request Type

- [X] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

### Additional context?

I suggest two possible solutions to this problem:
- Ignore this error if compiled in debug mode, prefer the local plugin version
- Don't count local versions of the builtin plugin at all

PR #833, opened 2022-11-23
[Feature] Ignore "Found more than one built-in plugin" if it's a debug build

ssmid:
Issue: #832

Built-in plugins coming from non-local sources are ignored now in debug builds. This allows developers to run a debug build of ImHex, while still having their normal version installed.


WerWolv:
Thank you! 
I added some more checks to your proposed changes and implemented it in the latest commit

PR #838, opened 2022-12-02
Add tip about unicode symbols

iTrooz:
This will help users to know about this "feature". It's not really documented anywhere else that disabling unicode symbols make ImHex load faster


WerWolv:
Thanks!

PR #839, opened 2022-12-03
fix: 3b94a42783911b59d821f4f4ff4d433b68eb1488 missing zh_TW.json

5idereal:


WerWolv:
Thank you! I was testing if unicode works correctly with that file and apparently deleted it by accident

PR #851, opened 2022-12-29
lang: Update german translation

123456abcdef:
Added missing translation, added missing/new entries, fixed typos and improved translations


WerWolv:
Thanks a lot!

PR #861, opened 2023-01-08
build: Improve Fedora builds and add EL9 build.

jonathanspw:
This converts the Fedora builds over to using the `fedpkg` utility for generation which uses mock instead of pure rpmbuild.  This will result in builds that are identical to the official Fedora/EPEL builds.

ccache is fully working with these and they are the faster than any of the other builds :)

The only maintenance for this now will be the spec file which is near-identical to the official Fedora/EPEL spec file, and every 6 months updating the numerical versions of Fedora that are being built as new versions of Fedora are released.

Rawhide is disabled in this as the builds are still failing citing the `FORTIFY_SOURCES` error and I'm not sure why.  I can only figure it must be kernel-related in some way.  My initial tests to set it back to level 2 instead of 3 weren't successful so I'll send another PR once I have time to look more into that so the rawhide builds will succeed again.

This took a lot of trial and error so of course squash all the commits when merging.


iTrooz:
Is there a reason you removed a lot of blank lines across build.yml ?
it seems a mistake as some are removed and some are not

Plus, I kinda liked them, felt it was easier to read

.github/workflows/build.yml
```diff
@@ -446,71 +420,111 @@ jobs:
...
+            release_num: 37
+            mock_config: fedora-37
+          - name: Fedora
+            mock_release: f36
+            release_num: 36
+            mock_config: fedora-36
+          - name: RHEL, AlmaLinux
+            mock_release: epel9
+            release_num: 9
+            mock_config: "alma+epel-9"
```
iTrooz, line 469: Not sure if adding even more release artifacts for Fedora if a good idea, given that the current ones aren't downloaded really much (17 download for the rawhide artifact for https://github.com/WerWolv/ImHex/releases/tag/v1.25.0 , which was the latest for almost 2 months)

The CI itself could stay if there are significant changes between f36/f37/rawhide/RHEL. Is it the case ?

.github/workflows/build.yml
```diff
@@ -446,71 +420,111 @@ jobs:
...
-        uses: hendrikmuhs/ccache-action@v1.2
+        uses: hendrikmuhs/ccache-action@v1.2.5
         with:
-          key: fedora-${{ matrix.release }}-${{ secrets.CACHE_VERSION }}-build-${{ github.run_id }}
-          restore-keys: fedora-${{ matrix.release }}-${{ secrets.CACHE_VERSION }}-build
-          max-size: 50M
-   
+          key: rpm-${{ matrix.mock_release }}-${{ secrets.CACHE_VERSION }}-build-${{ github.run_id }}
+          restore-keys: rpm-${{ matrix.mock_release }}-${{ secrets.CACHE_VERSION }}-build
+          max-size: 1G
```
iTrooz, line 508: Any reason for such a huge cache size ? I feel like useless cache would just accumulate with time

jonathanspw:
Hmm that wasn't intentional...copy/paste issue I think as I had trimmed it down to just Fedora builds while I was working on this.  I'll fix it.

dist/rpm/imhex.spec
```diff
@@ -13,16 +17,37 @@ BuildRequires:  file-devel
...
-BuildRequires:  mesa-libGL-devel
+BuildRequires:  libappstream-glib
+BuildRequires:  libglvnd-devel
 BuildRequires:  glfw-devel
 BuildRequires:  json-devel
 BuildRequires:  libcurl-devel
 BuildRequires:  llvm-devel
 BuildRequires:  mbedtls-devel
-%if 0%{?fedora} >= 37
+BuildRequires:  python3-devel
```
iTrooz, line 27: I don't think ImHex requires python anymore. @WerWolv am I right ?

dist/rpm/imhex.spec
```diff
@@ -37,50 +62,73 @@ same time ImHex is completely free and open source under the GPLv2 language.
...
 %build
-%cmake \
- -DCMAKE_BUILD_TYPE=%{_build_type} \
- -D USE_SYSTEM_NLOHMANN_JSON=ON \
- -D USE_SYSTEM_FMT=ON \
- -D USE_SYSTEM_CURL=ON \
- -D USE_SYSTEM_LLVM=ON \
- -DCMAKE_C_COMPILER_LAUNCHER=ccache        \
- -DCMAKE_CXX_COMPILER_LAUNCHER=ccache      \
- -D IMHEX_PATTERNS_PULL_MASTER=ON          \
```
iTrooz, line 53: Usually we want the nightly packages to have the latest stuff hence `IMHEX_PATTERNS_PULL_MASTER`

.github/workflows/build.yml
```diff
@@ -446,71 +420,111 @@ jobs:
...
+
       - name: ⬆️ Upload RPM
         uses: actions/upload-artifact@v3
         with:
-          name: Fedora ${{ matrix.release }} RPM
+          name: ${{ matrix.name }} ${{ matrix.release_num }} RPM
           path: |
-            imhex-${{env.IMHEX_VERSION}}-Fedora-${{matrix.release}}.rpm
\ No newline at end of file
+            results_imhex/${{env.IMHEX_VERSION}}/*/imhex-${{env.IMHEX_VERSION}}-0.*.x86_64.rpm
```
iTrooz, line 556: We try to keep the same convention for every package name, which would be imhex-1.26.2-Fedora36.rpm as of now (we don't distribute other architectures right now)

iTrooz (changes requested):

dist/rpm/imhex.spec
iTrooz: We add `-${{secrets.CACHE_VERSION }}` to the cache key to be able o invalidate the cache when needed. Could you add it ?

iTrooz (changes requested):

.github/workflows/build.yml
jonathanspw: How would you suggest doing it for EL builds then?  That covers RHEL, AlmaLinux, Oracle Linux, etc.  Generally it's denominated by "el" and the version so "el9".  "fc37" for example is just the fedora version of that same designation and what people in the ecosystem expect.

Since GH is (stupidly) zipping artifacts anyway does it really matter since they all have "pretty" names in the artifact list, ie https://github.com/WerWolv/ImHex/actions/runs/3868575604#artifacts ?

iTrooz:
We add -${{secrets.CACHE_VERSION }} to the cache key to be able o invalidate the cache when needed. Could you add it ?

https://github.com/WerWolv/ImHex/pull/861/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R492
and https://github.com/WerWolv/ImHex/pull/861/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R541 (for some reason I wasn't able to link the lines from github's UI)

.github/workflows/build.yml
jonathanspw: The actual ccache size per run right now is about 325M, so 50M was too small regardless.  Setting it to 1G does not harm, allows for future growth, and GH will auto purge the older stuff anyway before adding more.

Basically it does no harm.

.github/workflows/build.yml
jonathanspw: @WerWolv said the more the merrier basically...maybe he just meant CI.  Regardless, this is the currently supported official Fedora/EPEL releases so I think doing at least CI for all the officially supported releases makes sense.

dist/rpm/imhex.spec
jonathanspw: Is that the intent of this, to be nightly builds?  My goal was to test the builds as if they were official Fedora builds so that issues can be caught and figured out before pushing a release only to find it won't build on Fedora/EPEL for whatever reason.

It's still a nightly in terms of having the latest ImHex code.  I think using systems libs as the official RPMs will makes the most sense.  Not updating deps should have 0 effect on the resulting package unless it prevents ImHex from building to start with.

jonathanspw: Ohhhh patterns.  I see.  Maybe we should adjust that one.

.github/workflows/build.yml
iTrooz: > The actual ccache size per run right now is about 325M

https://github.com/jonathanspw/ImHex/actions/caches
Woah, indeed. Weird this only happens with the rpm-based jobs, but if it works, I'm not going to complain

.github/workflows/build.yml
jonathanspw: > We add -${{secrets.CACHE_VERSION }} to the cache key to be able o invalidate the cache when needed. Could you add it ?
> 
> https://github.com/WerWolv/ImHex/pull/861/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R492 and https://github.com/WerWolv/ImHex/pull/861/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R541 (for some reason I wasn't able to link the lines from github's UI)

Fixed this one.

dist/rpm/imhex.spec
jonathanspw: Is flipping `IMHEX_PATTERNS_PULL_MASTER=ON` compatible with `IMHEX_OFFLINE_BUILD=ON` ?

.github/workflows/build.yml
iTrooz: > How would you suggest doing it for EL builds then? That covers RHEL, AlmaLinux, Oracle Linux, etc. Generally it's denominated by "el" and the version so "el9". "fc37" for example is just the fedora version of that same designation and what people in the ecosystem expect.

Well.. on one hand, the goal is to allow people to quickly see what file they need, so I would go with the designation in the ecosystem.
That said, I also use Fedora and have never heard of the "fc37" acronym. I understand it's my fault and I should know better, but.. I don't. And some other people probably don't either. And as I said I would like everyone to be able to download it easily
If EL covers RHEL, AlmaLinux, Oracle Linux, etc.., I think "EL" would be a good name in that case
So my proposition would be `imhex-1.26.2-Fedora37.rpm` and `imhex-1.26.2-EL.rpm`

What do you think ? I'd especially like to hear your thought about `fc37`

> Since GH is (stupidly) zipping artifacts anyway does it really matter since they all have "pretty" names in the artifact list, ie [WerWolv/ImHex/actions/runs/3868575604#artifacts](https://github.com/WerWolv/ImHex/actions/runs/3868575604#artifacts) ?

Yes, because the release workflow unzips the (stupidly I agree) zipped files, and upload them automatically. So these are the name that actually show on the release page

.github/workflows/build.yml
iTrooz: > so I think doing at least CI for all the officially supported releases makes sense.

You're right
Let's wait for WerWolv's call on adding the artifacts to the release then

dist/rpm/imhex.spec
iTrooz: > Is that the intent of this, to be nightly builds? My goal was to test the builds as if they were official Fedora builds so that issues can be caught and figured out before pushing a release only to find it won't build on Fedora/EPEL for whatever reason.

At least, that's the intent I've seen for them. Testing the nightly version for a bugfix before the actual release to the Fedora repositories

> Is flipping IMHEX_PATTERNS_PULL_MASTER=ON compatible with IMHEX_OFFLINE_BUILD=ON ?

I don't think so. But the github CI do not have the same constraint as the Fedora build system, so I'd say it's not a problem to turn IMHEX_OFFLINE_BUILD back off (like it is now on master)

.github/workflows/build.yml
WerWolv: I'd probably drop Fedora 36 sometime in the future since Fedora 37 has been out for a few months already. RedHat is no problem.

With the more the merrier I mostly meant the more distros we support the better. We don't _need_ to support older versions of the same distro though in my opinion

dist/rpm/imhex.spec
WerWolv: Yep, the python dependency has been removed entirely from the build process with 1.26.0

.github/workflows/build.yml
WerWolv: The nightly artifact names do matter partially because I need their name to allow downloading them from my website. There you download the zipped versions directly

dist/rpm/imhex.spec
WerWolv: Yeah I also think we should enable non-offline build there and pull from latest master. Otherwise release builds will also not include the patterns (if offline build is used) or the patterns from the previous release (if we don't pull the latest master) as the artifacts get built before a new release happens on the pattern repo

.github/workflows/build.yml
jonathanspw: "fc37" for example is the designation used by all official RPMs.  Generally speaking if you're manually downloading packages for the OS from anywhere they'll follow this designation.

Also for EL distros, it needs to be "elX" with X being the major version (as I have it outputting now).  el9 package for example aren't compatible with el8, el7, etc.

I'd really really like to stick with filenames that match ecosystem expectations and are more technically accurate.

dist/rpm/imhex.spec
jonathanspw: Fixed, thanks for the info.

.github/workflows/build.yml
jonathanspw: Fedora 36 will go EOL in roughly 6-8 months.  I think it should be kept until then.

Many people, myself included, intentionally stay a version behind latest stable in order to get the most stability possible.  Fedora 36 is still extremely widely used today.  I haven't parsed the data lately but I'd be willing to be that 36 is currently still in wider use than 37.

dist/rpm/imhex.spec
jonathanspw: Fixed

.github/workflows/build.yml
iTrooz: In this case let's try with the ecosystem names. I will keep an eye to see if some people are confused

And I think there is going to be a problem with `-0` because IIRC it's the number of commits after the release, which won't always be 0 (the CI is run for every commit)

As for the architecture (x86-64).. I don't know. It's true that we only distribute x86-64 builds, but it could be good to clarify it
@WerWolv would you mind changing all artifact names to add the architecture ?

in any case, I'd like the artifact names to have a convention, so either they all have it or they all don't

WerWolv: Do we gain anything from adding the architecture though? Like, I doubt I'll ever add real support for other architectures since GitHub only supports x86_64 too. Having too many different tags in the release will probably just make things confusing for users

jonathanspw: It's standard practice and keeps things consistent.  In the case of Fedora there are also aarch64 and ppc64le images as well so at least for the Fedora ecosystem where there ARE other "official" builds it keeps things straight.

It's hard to imagine people that use ImHex lack a basic understanding of architectures, or at least the realization that they're on x86_64.  Since there are no others being produced by CI here I cannot see it as a point of confusion.

My whole goal here is to have the CI as easy to maintain as possible, and in the current state it's basically a matter taking the specs that I'll maintain for the official Fedora/EPEL builds and paste them here to get the updates/improvements without having a whole slew of customizations to deal with that make it harder for me to keep up with.

I mean this isn't a hill I'm going to die on by any means but I hope you'll see where I'm coming from on this naming :)

iTrooz: Up on the `-0` thing, I think I didn't get a response

WerWolv: I think it's going to be very difficult to properly update that with every single build we make. And just having `-0` there just to have it doesn't feel very useful either

jonathanspw: The `-0` is not "just to have it".  It's the standard release designation on rpm packages used by every single RPM distro out there.

jonathanspw: http://ftp.rpm.org/max-rpm/s1-rpm-inside-tags.html see "Package Naming Tags"

WerWolv: It will kinda always be hardcoded to `-0` then. If it makes things easier, feel free to add it to the Fedora builds but I think it's pretty useless. The other builds definitely don't need it

iTrooz: Well sorry, but I don't think it's really useful to put it, and we already rename all other artifacts, there would be no reason to keep the Fedora filename but not the deb, the AppImage, etc..

I understand you want to keep it the official filename, but personally I care more about have a standard among all artifacts

jonathanspw: > It will kinda always be hardcoded to `-0` then. If it makes things easier, feel free to add it to the Fedora builds but I think it's pretty useless. The other builds definitely don't need it

It is hard-coded to 0 because so nightlies are always the 0 release of the version.  Fedora and EL builds always start at -1 so if someone manually installs a nightly and subsequently updates through Fedora/EL they can get the official builds from repos instead of ending up with an unmaintained version.

jonathanspw: Adjusted per Discord convo.

iTrooz (changes requested):
Seems good to me !
i don't think this last change is something we would disagree on, and I don't want to bother you more, so I'm going to commit it myself and merge

Thanks !

iTrooz (approved):

PR #873, opened 2023-01-12
cleanup build CI + separate Ubuntu and AppImage builds

iTrooz:
Two things in one !

First, this PR cleans up the CI file. I mostly remove comments with the rationale that we can also get them back from the older commits

The PR also moves the AppImage build to a different runner than Ubuntu. I made a mistake thinking that AppImage was "just a repackaging of the deb file". Some flags are changing, and this is causing the build to occur twice in the runner.

Ready for merge


iTrooz:
Drafted because I just noticed a small problem, I'm double checking just to be sure

iTrooz:
Ready now !

WerWolv:
Thanks a lot!

PR #889, opened 2023-01-16
Support macOS keyboard shortcuts

Kuruyia:
This PR adds a new modifier definition, `CTRLCMD`, which is defined as `SUPER` (command) on macOS and `CTRL` on other platforms, for a more streamlined experience on Mac.

Keyboard shortcuts in menus are also changed depending on the platform (`CTRL` becomes `CMD` and `ALT` becomes `OPT` on macOS), although I'm waiting for your input on whether this can be better implemented.


WerWolv:
Thanks a lot!

PR #890, opened 2023-01-17
fix: Frame rate not unlocking correctly

birdstakes:
Repeatedly mashing a key (**not** holding it down) eventually starts to lag.

https://user-images.githubusercontent.com/25270458/212798881-120a792d-c481-4f1d-9138-12c6b6fe689e.mp4

I believe this is because m_hadEvent could be set during glfwWaitEventsTimeout and then immediately cleared before ever being checked. This patch changes the order so that m_hadEvent is preserved until the next check so that the frame rate can be unlocked. This also no longer clears m_hadEvent when the window is invisible or iconified but I'm not sure if that matters.


WerWolv:
Thank you!

Issue #910, opened 2023-01-30
Build fails in macOS Ventura


ghmanoj:
### Operating System

MacOS

### What's the issue you encountered?

CMake is identifying the compiler as ``AppleClang`` causing the build to fail.

``
CMake Error at cmake/build_helpers.cmake:339 (message):
  ImHex can only be compiled with GCC or Clang.  AppleClang is not supported.
Call Stack (most recent call first):
  CMakeLists.txt:28 (verifyCompiler)
``

### How can the issue be reproduced?

Build fails when compiling in macOS Ventura.
$ mkdir _build && cd _build
$ cmake ..

### ImHex Version

master 

### ImHex Build Type

- [X] Nightly or built from sources

### Installation type

From Git Clone

### Additional context?

Last commit hash 434ced44f068008116647f15df561457b9de3219

WerWolv:
As mentioned in #911, Apple's clang is multiple versions outdated and doesn't support many of the latest C++ features ImHex is using. GCC can however be installed on macOS through homebrew which is able to compile ImHex successfully.

gtackett:
> As mentioned in #911, Apple's clang is multiple versions outdated and doesn't support many of the latest C++ features ImHex is using. GCC can however be installed on macOS through homebrew which is able to compile ImHex successfully.

Hmm. My problem seems to be similar to that of @ghmanoj, although I'm trying to build on Big Sur, and I do have `gcc` et al. installed through **homebrew**. I followed the MacOS build instructions at (https://github.com/WerWolv/ImHex/tree/master/dist/compiling/macos.md). CMake is nevertheless identifying the compiler as AppleClang causing the build to fail.
1. Cloned the repo using git clone https://github.com/WerWolv/ImHex --recurse-submodules
2. Installed all the dependencies using brew bundle --no-lock --file dist/Brewfile
3. Tried to build ImHex itself using the following commands:
cd ImHex
mkdir -p build
cd build
CC=$(brew --prefix llvm)/bin/clang        \
CXX=$(brew --prefix llvm)/bin/clang++     \
OBJC=$(brew --prefix llvm)/bin/clang      \
OBJCXX=$(brew --prefix llvm)/bin/clang++  \
cmake -G "Ninja"                          \
  -DCMAKE_BUILD_TYPE=Release              \
  -DCMAKE_INSTALL_PREFIX="./install"      \
  -DIMHEX_GENERATE_PACKAGE=ON             \
  ..
ninja install
[...]

PR #911, opened 2023-01-30
 fix: macOS cmake build error

ghmanoj:
Fixes macOS cmake build error. #910


WerWolv:
Hey! Thank you. Did you check if building actually works with AppleClang now? For a very long time it was massively outdated and didn't support half of the features I've been using on other platforms which is why you need to compile everything using gcc

ghmanoj:
Hi, I tried to compile. make is giving many errors like compiler flag errors and C++20 language feature errors..

WerWolv:
Yeah that's the issues I was talking about. Apple Clang is currently based on LLVM 13 afaik and some of the features I'm using in ImHex require either GCC 12 or later or Clang 15 or later.

I might add support for AppleClang in the future when it caught up a little bit but until then, please use GCC on macOS (you can use the homebrew install script from the /dist folder) to build ImHex

PR #967, opened 2023-03-10
Allow building with `AppleClang`

fwcd:
The version of Clang shipped with Xcode identifies itself as `AppleClang` and should be largely equivalent to the open-source Clang distribution.

Note that Clang 14 (the version that currently ships with Xcode) cannot build the project, however, e.g. due to default implementations of `==` not being `constexpr` ([see here](https://github.com/WerWolv/PatternLanguage/blob/572a481803aa19c5bd0c3ccce08d63666d01499c/lib/include/pl/core/token.hpp#L160)).


WerWolv:
Thanks for the PR.
The reason AppleClang is not included is because it's always multiple versions behind regular Clang. Since ImHex uses various C++20/23 features, I decided to simply not support it and use gcc for building on all platforms. That way I don't need to hold back on actually using modern features because some platform doesn't update their compilers.

fwcd:
Hm, but in that case, shouldn't we just check for <= Clang 15?

WerWolv:
AppleClang's version sadly has absolutely nothing to do with regular Clang's version as you can see here:
https://en.wikipedia.org/wiki/Xcode#Xcode_11.0_-_14.x_(since_SwiftUI_framework)_2

Clang 15.0.0 _should_ support everything but that's only included in XCode 14.3 Beta 2 according to that table which reports as 14.0.3.

fwcd:
The versioning still seems reasonably close to upstream, I wouldn't call it multiple versions behind. Checking for AppleClang >= 15 should still catch this, right, even if it means waiting for the next major Xcode release?

WerWolv:
I didn't want to include version checks because I need to keep them up-to-date then constantly. Which becomes difficult since I have no mac myself. I had big trouble properly supporting Apple in the past because their toolchain was different from the ones on Windows and Linux so I made the decision to just use GCC everywhere. Regular clang is also supported because I can test it on other OSes but for AppleClang that's just not possible for me.

Also the AppleClang version will diverge from the Clang version again in the future. It's really just a coincidence that it's somewhat close right now. If you go a bit further up in the table you can see AppleClang 11.0.0 uses Clang 8.0.0, 11.0.3 uses 9.0.0 and so on. So Apple might stay on version 14.X.X of their toolchain for a while now while Clang gets updated to 15.0.0 and maybe 16.0.0 in the future.

fwcd:
Hm, that's understandable, but thanks for clarifying. Perhaps if Apple manages to catch up enough, we could add GitHub Actions CI for AppleClang to take care of the testability issue?

WerWolv:
The Apple CI previously was using AppleClang and it just ended up breaking constantly. And since I don't have a mac, the only way to figure out why it was breaking and to fix it was to push commits to the repo, wait 10 minutes until it failed and then try again. That was just extremely frustrating.
There's not really any benefit to the end user to support a bunch of different compilers and it makes my life so much easier to just support a single compiler so I'd really rather just stick to GCC on all platforms.

fwcd:
What do you think about building the macOS version using open-source Clang in CI? That would probably also make it easier to cross-compile to native arm64 or Universal binaries (since we wouldn't need a separate compiler for every architecture as with GCC)?

WerWolv:
You'd still need multiple toolchains to cross-compile for ARM with clang. The only thing you can reuse is the compiler. So if possible at all, I'd still prefer using GCC there.

I'd be up for trying though if you can get the CI to build ARM or universal binaries. Having native M1 builds would be neat, I just have no way to test it myself

WerWolv:
We're not using Xcode literally at all though right now.
If that natively works with regular clang too, feel free to try. I'm happy to merge a PR that adds ARM binaries

fwcd:
Hm, this seems to be a larger endeavour, aside from some other `constexpr`-related things that Clang 15 lacks, `codecvt_utf8_utf16` seems to be deprecated in libc++...

Building with GCC works, but segfaults at launch. I'll have to look at this when I have some more time.

PR #993, opened 2023-03-23
Move config file to XDG_CONFIG_HOME

iTrooz:
This pull request changes Config Directories on Linux to only include the XDG_CONFIG_HOME directory, as opposed to all directories in XDG_DATA_DIRS before (introduced in https://github.com/WerWolv/ImHex/pull/644/files#diff-c1a4d2b63fed168a9a3568944e9cadeae096f2ddcec3649e4a9b2d29fd104be0L162-L166).

Reasons:
- This changes the location of the config file to the standard directory meant for configurations
- This prevents the config file from being read/written in system locations, like /usr/share

This PR also includes a migration task that will run on Linux and move config/GUI dimensions to the new directory

as a bonus, as discussed on discord, it writes the logs to a Data directory instead of a Config directory

### DIFF
--- a/lib/libimhex/include/hex/helpers/fs.hpp
+++ b/lib/libimhex/include/hex/helpers/fs.hpp
@@ -48,4 +48,7 @@ namespace hex::fs {
 
     std::vector<std::fs::path> getDefaultPaths(ImHexPath path, bool listNonExisting = false);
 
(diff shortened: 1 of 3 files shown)


WerWolv:
Thank you!

PR #995, opened 2023-03-24
Handle and show NFD errors

iTrooz:
This PR handles errors that NFD might encounter (both in Init() and the other method to open the dialog), and log them in the logs and in the GUI

This (among other) fix the crash I had running ImHex as root and opening a file


iTrooz:
MacOS Build is failing due to a (presumed) issue with nfd-extended, issue is linked

PR #1004, opened 2023-03-29
feat: allow loading and saving pattern code via events

H1X4Dev:
Currently there is no way to save the pattern code progamically from a plugin unless the builtin plugin is modified to add those events. This pull request will be adding ability to load and save pattern code from specified file.


WerWolv:
Thank you!

PR #1036, opened 2023-04-18
build: Add option to disable update checking

classabbyamp:
This is aimed at use by linux distros, where package updates come from a central location, and users shouldn't need to worry about updating ImHex on their own. This disables parts of the ImHex UI that would not be useful in that case.

Tested and confirmed that this works in both states of the of the `-DIMHEX_DISABLE_UPDATE_CHECK` switch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ option(IMHEX_PATTERNS_PULL_MASTER "Download latest files from master branch of t
 option(IMHEX_IGNORE_BAD_COMPILER "Allow compiling with an unsupported compiler" OFF)
 option(IMHEX_USE_GTK_FILE_PICKER "Use GTK file picker instead of xdg-desktop-portals" OFF)
 option(IMHEX_DISABLE_STACKTRACE "Disables support for printing stack traces" OFF)
+option(IMHEX_DISABLE_UPDATE_CHECK "Disables built-in update check" OFF)
 
 # Basic compiler and cmake configurations
 set(CMAKE_CXX_STANDARD 23)
@@ -41,7 +42,7 @@ set(PLUGINS
 # Add various defines
 detectOS()
 detectArch()
-addVersionDefines()
+addDefines()
 configurePackingResources()
 setUninstallTarget()
 addBundledLibraries()
--- a/main/source/init/tasks.cpp
+++ b/main/source/init/tasks.cpp
@@ -30,6 +30,7 @@ namespace hex::init {
 
     using namespace std::literals::string_literals;
 
+#if defined(HEX_UPDATE_CHECK)
     static bool checkForUpdates() {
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -46,6 +47,10 @@ configurePackingResources()
 setUninstallTarget()
 addBundledLibraries()
 
+if (NOT IMHEX_DISABLE_UPDATE_CHECK)
+    add_compile_definitions(HEX_UPDATE_CHECK)
+endif()
```
classabbyamp, line 52: wasn't quite sure where to put this. I can move it if you have a better spot

iTrooz:
fine by me ! What do you think @WerWolv  ?

CMakeLists.txt
WerWolv: This here should definitely go into the `cmake/build_helpers.cmake` file instead.
Might need a bit of refactoring at some point but you can probably put it in `addVersionDefines` (or rename that function to `addDefines`)

main/source/init/tasks.cpp
```diff
@@ -29,6 +29,7 @@ namespace hex::init {
 
     using namespace std::literals::string_literals;
 
+#ifdef HEX_UPDATE_CHECK
```
WerWolv, line 32: Use `#if defined(HEX_UPDATE_CHECK)` instead of ifdef

main/source/init/tasks.cpp
```diff
@@ -474,7 +476,9 @@ namespace hex::init {
             #endif
             { "Loading settings",        loadSettings,        false },
             { "Loading plugins",         loadPlugins,         false },
+#if HEX_UPDATE_CHECK
```
WerWolv, line 479: This should probably also use `#if defined`?

plugins/builtin/source/content/settings_entries.cpp
```diff
@@ -45,6 +45,7 @@ namespace hex::plugin::builtin {
         1 - check for updates on startup
         2 - default value - ask the user if he wants to check for updates. This value should only be encountered on the first startup.
         */
+#ifdef HEX_UPDATE_CHECK
```
WerWolv, line 48: Ditto

plugins/builtin/source/content/welcome_screen.cpp
```diff
@@ -511,6 +511,7 @@ namespace hex::plugin::builtin {
                 loadDefaultLayout();
         });
 
+#ifdef HEX_UPDATE_CHECK
```
WerWolv, line 514: Ditto

WerWolv (changes requested):

WerWolv:
Thanks a lot for the PR! I'm sorry it took so long for me to look at it

WerWolv (approved):

PR #1041, opened 2023-04-20
Add Spanish translations

XorTroll:
This translates ImHex for the Spanish language


WerWolv:
Thank you so much!

Could you make sure that all submodules are synced with the version the master branch uses?

PR #1047, opened 2023-04-27
Fixed a number of errors and added some features to the IEEE745 tool. There are no issues reportimg these errors or feature request gor the added features.

paxcut:
I just realized one feature request existed about this tool and have added a comment to it referring this pr.  Errors and additions are described in the fork commit already. I'm not sure if I should repeat them here again. I have tested the changes thoroughly, but it is always possible  some fringe case was not tested and is incorrect. The tests were done using the many similar online calculators for IEEE 754 floating point formats.
IEEE 745 floating point tool redesign modeled after 'float toy' web app (http://evanw.github.io/float-toy/)

Streamlined output using colors and compact layout which can be further simplified.
Chosen display mode (detailed or simplified) is automatically saved and set on new sessions.
Edit the binary bits, the integer hexadecimal or the floating point decimal values and the entire app will update with the change.
Supports the main IEEE745 standard formats (half, single and double precision) together with custom formats of size <= 64 bits.
Each format choice uses and displays the number of significant decimal digits defined by the mantissa size.
Added labels to identify the location of each bit box inside the binary representation.
Satisfies round trip / idempotent (reproducing) conversion property
Added theme colors, radio buttons for display mode and a clear button that resets the tool.
Removed previously and incorrectly added locale translation to various labels and languages
Attempted to adhere to code style formatting using existing code as example. 
An effort was made to use preferred variable types and functions from std namespace when appropriate.
Attempted to document code using comments. 

Not implemented / left to complete at an later time

Arbitrary width and precision formats. 
Extended precision formats.
Shortest string property.
hexadecimal floating point display and conversions.


plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1151,112 +1154,140 @@ namespace hex::plugin::builtin {
...
             const static auto ExtractBits = [](u32 startBit, u32 count) {
                 return hex::extract(startBit, startBit - (count - 1), value);
             };
-
-            const auto signBits     = ExtractBits(signBitPosition, 1);
-            const auto exponentBits = ExtractBits(exponentBitPosition, exponentBitCount);
-            const auto mantissaBits = ExtractBits(mantissaBitPosition, mantissaBitCount);
-
+            const auto signBits = ExtractBits(signBitPosition, 1);
+            long long  exponentBits = ExtractBits(exponentBitPosition, exponentBitCount);
```
WerWolv, line 1165: Any reason that you changed these values from `const auto` to `long long`? If you need a signed type, prefer `i64`.

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1151,112 +1154,140 @@ namespace hex::plugin::builtin {
...
+
+            enum class NumberType {
+                Regular,
+                SignalingNaN,
+                QuietNaN,
+                NegativeInfinity,
+                PositiveInfinity,
+            } numberType;
+
+            if (std::popcount(static_cast<u64>(exponentBits)) == static_cast<long long>(exponentBitCount)) {
```
WerWolv, line 1234: Should be `i64` again probably

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1266,73 +1297,188 @@ namespace hex::plugin::builtin {
...
+                        exponentString = "NaN";
+                    else
+                        exponentString = std::to_string(exponentBits);
+                    ImGui::TextFormatted("2^({0} - {1})", exponentString, exponentBias);
                     ImGui::TableNextColumn();
                     ImGui::TextUnformatted("=");
                     ImGui::TableNextColumn();
-                    ImGui::TextFormatted("{0:.8G}", exponentValue);
-
+                    u32 precision =  ceil(1+mantissaBitCount*0.30102999566398);//    (exponentBitCount + mantissaBitCount + 1) / 4;
```
WerWolv, line 1315: - `std::ceil` instead of `ceil`
- What does this magic number here mean? Is this really necessary? If yes, please define it as a `constexpr static auto` constant somewhere and give it a proper name

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1266,73 +1297,188 @@ namespace hex::plugin::builtin {
...
             }
+            // For a number written in decimal floating point notation with p exponent bits and q mantissa bits,
+            // find the closest IEEE 754 floating point number.
+            // The exponent is stored as an integer in excess notation, with a bias of 2^(p-1)-1.
+            // The mantissa is stored as a fraction with an implicit leading 1 bit only for normal values
+            // (i.e. for values with an exponent of 2^(p-1)-1 or greater).
+            // For subnormal values, the leading 1 bit is omitted.
+            // The sign bit is stored as the most significant bit.
+            static std::string decimalFloatingPointNumberString;
+            u32 precision =  ceil(1+mantissaBitCount*0.30102999566398);
```
WerWolv, line 1394: Same thing as above. This can probably be cleaned up a bit

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1266,73 +1297,188 @@ namespace hex::plugin::builtin {
...
+            decimalFloatingPointNumberString = fmt::format("{0:.{1}G}", resultFloat, precision);
+            static const  ImGuiInputTextFlags flags = ImGuiInputTextFlags_EnterReturnsTrue;
+            char *endp;
+            if (ImGui::InputText("hex.builtin.tools.ieee754.result.float"_lang, decimalFloatingPointNumberString, flags)) {
```
WerWolv, line 1398: Try to stick to the same formatting style as the rest of ImHex uses.
For example, insert new lines, use spaces between `if` and the `(` and so on.
Also generatlly don't use the raw C functions such as `strtod` or `isnan` but use the ones in the `std::` namespace instead

WerWolv (changes requested):

plugins/builtin/source/content/tools_entries.cpp
paxcut: There was no reason for the choice. I'll switch all variables that are I created as long long to i64.

plugins/builtin/source/content/tools_entries.cpp
paxcut: yes, though this code may not be relevant but I will incorporate all the suggestions to the new code before committing it.

plugins/builtin/source/content/tools_entries.cpp
paxcut: the 0.301... is the value of log10(2). The quantity ceil(1+"mantissa bit count" * log10(2)) is a good approximation to the number of significant digits present in the decimal representation.  I can call the constant logBaseTenOfTwo. I am making the changes to my local clone where the new code that hasn't been committed to the fork lives, hence all the comments without code changes.

plugins/builtin/source/content/tools_entries.cpp
paxcut: will do, sorry about working on changes in the middle of the review. I'll try to complete this soon but I am having some trouble with the addition of the display mode to the project file so it may take a little bit more time. I will also undo all changes to docs for languages other than those I am fluent and I may even skip those because I am not that confident in my ability as a translator. Thank you for your patience and continued support.

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1154,328 +1182,577 @@ namespace hex::plugin::builtin {
...
             const auto mantissaBitPosition = totalBitCount - 1 - exponentBitCount;
+
             const static auto ExtractBits = [](u32 startBit, u32 count) {
                 return hex::extract(startBit, startBit - (count - 1), value);
             };
-            const auto signBits = ExtractBits(signBitPosition, 1);
-            long long  exponentBits = ExtractBits(exponentBitPosition, exponentBitCount);
-            long long mantissaBits = ExtractBits(mantissaBitPosition, mantissaBitCount);
+
+            i64 signBits = ExtractBits(signBitPosition, 1);
```
paxcut, line 1197: I changed the types of all 8 byte signed integers to i64 as requested

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1154,328 +1182,577 @@ namespace hex::plugin::builtin {
...
 
-            enum class NumberType {
-                Regular,
-                SignalingNaN,
-                QuietNaN,
-                NegativeInfinity,
-                PositiveInfinity,
-            } numberType;
+                // Check if all exponent bits are set.
+                if (std::popcount(static_cast<u64>(exponentBits)) == static_cast<i64>(exponentBitCount)) {
```
paxcut, line 1463: Also made the requested change here.

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1130,16 +1127,51 @@ namespace hex::plugin::builtin {
...
-            static u128 value = 0x00;
+
+            static u128 value = 0;
             static int exponentBitCount = 8, mantissaBitCount = 23;
+            long double exponentValue;
+            long double mantissaValue;
+            static long double resultFloat;
+            i64 exponentBias;
+            constexpr static auto flags = ImGuiInputTextFlags_EnterReturnsTrue;
+            constexpr static auto logBaseTenOfTwo = 0.30102999566398119521373889472449L;
```
WerWolv, line 1150: Constexpr variables in general should be PascalCase.

Also I think `logBaseTenOfTwo` deserves a name that better describes its functionality. 
Instead of hardcoding the value, you should also be able to just use `std::log10(2)` here. Most compilers have constexpr implementations for this function

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1160,179 +1194,610 @@ namespace hex::plugin::builtin {
...
+
+                if (displayMode == 0) {
+
+                    unsigned signColorU32 = ImGui::GetCustomColorU32(ImGuiCustomCol_IEEEToolSign);
+                    unsigned expColorU32 = ImGui::GetCustomColorU32(ImGuiCustomCol_IEEEToolExp);
+                    unsigned mantColorU32 = ImGui::GetCustomColorU32(ImGuiCustomCol_IEEEToolMantissa);
+
+                    // this has the effect of dimming the color of the numbers so user doesn't try
+                    // to interact with them.
+                    if (isThemeDark)
```
WerWolv, line 1668: Instead of doing this manually, consider using `ImGui::BeginDisabled` and `ImGui::EndDisabled`. That automatically dimms the colors of all widgets in that block

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1160,179 +1194,610 @@ namespace hex::plugin::builtin {
...
+
+                // For C++ from_chars is better than strtold.
+                // the main problem is that from_chars will not process special numbers
+                // like inf and nan, so we handle them manually
+                static std::string decimalFloatingPointNumberString;
+                static std::string_view decimalStrView;
+                // use qnan for quiet NaN and snan for signaling NaN
+                if (numberKind == NumberKind::NaN) {
+                    if (numberType == NumberType::QuietNaN)
+                        decimalFloatingPointNumberString = fmt::format("qnan");
```
WerWolv, line 1522: Don't need to pass this through fmt::format

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1160,179 +1194,610 @@ namespace hex::plugin::builtin {
...
+                std::from_chars_result res;
+                if (ImGui::InputText("##resultFloat", decimalFloatingPointNumberString, flags)) {
+                    // Always obtain sign first.
+                    if (decimalFloatingPointNumberString[0] == '-') {
+                        signBits = 1; // and remove it from the string.
+                        decimalFloatingPointNumberString.erase(0, 1);
+                    } else //important to switch from - to +.
+                        signBits = 0;
+
+                    std::string inputType;
```
WerWolv, line 1544: This should probably be an enum and not a string

WerWolv (changes requested):

WerWolv:
Looks good for the most part, thank you!

plugins/builtin/source/content/tools_entries.cpp
paxcut: I removed the constant and called the function instead as suggested.

plugins/builtin/source/content/tools_entries.cpp
paxcut: Yes, using BeginDisabled does exactly what I wanted and it looks better this way. Changes as requested.

plugins/builtin/source/content/tools_entries.cpp
```diff
@@ -1160,179 +1194,610 @@ namespace hex::plugin::builtin {
...
+                        ImGui::TextFormatted("{}", labelIndex);
+                        ImGui::Unindent(indentSize + checkBoxWidth / 2 - labelWidth / 2);
+                    }
+                }
+                ImGui::TableNextRow();
+                // Row for bit checkboxes
+                // Result
+                ImGui::TableNextColumn();
+
+                u64 mask = ((u128(1) << (totalBitCount + 1)) - 1);
```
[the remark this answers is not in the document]
paxcut: I switched to the suggested function and tested to make sure it was done correctly.

plugins/builtin/source/content/tools_entries.cpp
paxcut: Yes, sorry about that. Copilot syndrome...

plugins/builtin/source/content/tools_entries.cpp
paxcut: I turned the string to an enum which should be faster and more readable.

WerWolv (approved):

WerWolv:
Thanks a lot!

PR #1050, opened 2023-04-28
Show an error message when we can't open a file

iTrooz:
This PR does two correlated things:

- Show a generic error message (Failed to open provider) rather than a file-specific message (Failed to open file) when a provider fails to open
- Set the error to something more specific when opening a file fails


WerWolv (approved):

WerWolv:
Thanks!

PR #1053, opened 2023-05-01
Add a way to restore default layout when there are no views

iTrooz:


WerWolv:
ImGui::DimmedButton can't be found right now so the PR doesn't build

iTrooz:
Weird, I usually make sure my code compile before opening a PR.  probably my cache

Issue #1057, opened 2023-05-04
[Feature] Automatically open the project when opening a .hexproj file


iTrooz:
In most cases users didn't intend to open the .hexproj directly, but the project

PR #1061, opened 2023-05-06
open .hexproj files as a Project

iTrooz:


iTrooz:
Fixes https://github.com/WerWolv/ImHex/issues/1057

iTrooz:
By the way, I am really unsure about my implementation, especially `reinterpret_cast`. This PR as it is is more for me to get feedback, because I doubt the implementation is good

WerWolv:
A regular `static_cast` should be enough here because we're just down-casting through a inheritance hierarchy. If you want to be extra careful you can use a `dynamic_cast` and then check if the result is not `nullptr` but that's probably not necessary

iTrooz:
Done ! I went with `static_cast`

PR #1065, opened 2023-05-10
Explicitely convert data processor project data to json

iTrooz:
This will correct the errors `cannot use operator[] with a string argument with string[..]` when loading a project


WerWolv:
Thank you!

PR #1089, opened 2023-05-19
build: add support of compilers/stdlibs with broken `std::expected`, like the latest Clang.

KOLANICH:


WerWolv:
Thanks a lot for the effort but I'm not sure I'm comfortable with having to fallback to a bunch of random libraries to work around standard library bugs. In my opinion, the better option would be to just implement a version of std::expected ourselves since it's pretty easy to do.

What exactly is the problem with std::expected there? Does it just not exist or is its implementation actually completely broken?

KOLANICH:
The problem is https://github.com/llvm/llvm-project/issues/62801 . `libstdc++` requires C++ "concepts" of version no less than 202002L. The very latest clang doesn't support concepts at all. I guess it is surely a bug in `libstdc++` that for this feature concepts are required for such a feature, but for now we have to work around it. It can be worked around by using another C++ standard library (by adding `-stdlib=libc++` flag to `clang`, and its runtime is ~200 kiB, so there is no problem to just use it from the memory PoV), but distros usually require packages to use a certain libc, and usually it is the libc used by `gcc`. Also I'm not sure that there exists `libc++` for Windows (while MinGW exists and I usually use Clang+MinGW (and don't use Visual Studio) when build software targetting Windows).

>In my opinion, the better option would be to just implement a version of std::expected ourselves since it's pretty easy to do.

As you wish :) In my code I just use fallback impls and it used to work pretty well.

WerWolv:
After some internal discussions we decided to close this PR.
Basically, there's not really any good reason to use libc++ over libstdc++ other than personal preference. All our builds are compiled using GCC with libstdc++ on all platforms which makes it much easier to support all major OSes without needing a bunch of custom implementations for each of them. 
libc++ isn't truly supported therefor anyways and I honestly don't want to add a bunch of extra work for myself to support it either because again, there's little we gain from it.  

Thanks a lot for the effort anyways though

KOLANICH:
>Basically, there's not really any good reason to use `libc++` over `libstdc++`

It is `libstdc++` that is broken. When using `clang` with `libc++`, it has `std::expected`. But when using `clang` with `libstdc++`, then there is no `std::expected` because `libstdc++` requires concepts.

raquentin:
To anyone coming across this, I was able to fix the annoying lsp issues using a `.clangd` with:
```yaml
CompileFlags:
  Add:
    - "-std=c++23"
    - "-D__cpp_concepts=202002L"
```

PR #1090, opened 2023-05-19
build: comment out `newProvider->open();` in `plugins/builtin/source/content/events.cpp`. `newProvider::open` requires return value to be used, but it is not used. 

KOLANICH:
Given that there is the same invocation nearby, I guess it is just forgotten to be removed.

### DIFF
--- a/plugins/builtin/source/content/events.cpp
+++ b/plugins/builtin/source/content/events.cpp
@@ -94,7 +94,7 @@ namespace hex::plugin::builtin {
                             ImHexApi::Provider::createProvider("hex.builtin.provider.file", true)
                         );
                         newProvider->setPath(path);
-                        newProvider->open();
+                        //newProvider->open();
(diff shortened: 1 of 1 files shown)


WerWolv:
First of all, thanks a lot for all the PRs!

Have you tested this change? I'm pretty sure that just removing that line will cause the provider to just not be opened at all so no data can be read from it.

KOLANICH:
No, I haven't. TBH I'm for now (in future I'm thinking about making a plugin) unfamiliar to the code base and its internal working and currently just fix the issues preventing me to properly build this awesome tool using just common sense. But I guess you know how to deal with that line properly.

WerWolv:
Turns out you were right anyways. I corrected the whole logic there a bit now. Thank you!

KOLANICH:
Thank you for fixing this :)

PR #1091, opened 2023-05-19
build: Move the flags causing warnings in Clang into an if block adding them only for GCC.

KOLANICH:
Move all the flags added into an if block, since the flags are for GCC/Clang and are not suitable for Visual Studio.


iTrooz:
Hey, you might want to rebase/merge your branch on current master

iTrooz:
I am not sure what the issue is. Would you add more context to the PR ?

ImHex only supports gcc, not clang/MSVC

KOLANICH:
>I am not sure what the issue is. Would you add more context to the PR ?

Certain flags are causing warnings by Clang because it doesn't understand those flags and reports about it. But the flag making warnings errors is used in this project. This makes the warnings about the unrecognized flags errors, this way breaking compilation. This PR makes some flags being supplied only if GCC is used.

>ImHex only supports gcc, not clang/MSVC

I'm not the person who is in position to take political decisions about supporting/non-supporting compilers and/or platforms in this project. I can only say, that my previous PRs fixing some minor code smell issues in code were because of the issues in code detected by clang and missed by the used version of gcc. Given that warnings are considered errors in this project, I guess this project is strongly interested in not having code smell issues. It can be enough rationale to start supporting latest versions of clang (including CI, they can be installed from LLVM Debian repo). Another rationale is that clang is is a convenient compiler for cross-building.

WerWolv:
Thank you!

PR #1092, opened 2023-05-19
Implement automatically using of the deps installed in system.

KOLANICH:


WerWolv:
I'm unsure if this is a good choice. Setting e.g `USE_SYSTEM_FMT` should probably be an error if libfmt is not installed instead of just silently falling back to the bundled vesion

KOLANICH:
I don't understand what exactly you meant in your last sentence, so I'd elaborate my position about the issue.


It is written

```cmake
function(initDependencySourceChoice varName name)
    if(NOT DEFINED CACHE{${USE_SYSTEM_${varName}}})
        cmake_parse_arguments("" "PKGCONFIG" "VERSION" "" ${ARGN})
        message(STATUS "varName ${varName} name ${name} _PKGCONFIG ${_PKGCONFIG} _VERSION ${_VERSION}")
        if(_PKGCONFIG)
            if(_VERSION)
                pkg_check_modules("${varName}" "${name}>=${_VERSION}")
            else()
                pkg_check_modules("${varName}" "${name}")
            endif()
            set(found "${${varName}_FOUND}")
        else()
            if(_VERSION)
                find_package("${name}" "${_VERSION}")
            else()
                find_package("${name}")
            endif()
            set(found "${${name}_FOUND}")
        endif()
        set("USE_SYSTEM_${varName}" "${found}")
        set("USE_SYSTEM_${varName}" "${USE_SYSTEM_${varName}}" CACHE BOOL "Use ${varName} from system-installed packages")
        set("USE_SYSTEM_${varName}" "${USE_SYSTEM_${varName}}" PARENT_SCOPE)
        set("${varName}_DESIRED_VERSION" "${_VERSION}" PARENT_SCOPE)
        set("${varName}_PACKAGE_NAME" "${name}" PARENT_SCOPE)
    endif()
endfunction()

    initDependencySourceChoice(FMT fmt VERSION "8.0.0")
    if(NOT USE_SYSTEM_FMT)
        add_subdirectory(${EXTERN_LIBS_FOLDER}/fmt EXCLUDE_FROM_ALL)
        set_target_properties(fmt PROPERTIES POSITION_INDEPENDENT_CODE ON)
        set(FMT_LIBRARIES fmt::fmt-header-only)
    else()
        find_package(fmt "${FMT_DESIRED_VERSION}" REQUIRED)
        set(FMT_LIBRARIES fmt::fmt)
        list(APPEND badCloneExclusions "fmt")
    endif()
```

`initDependencySourceChoice` searches for compatible versions of a module installed in the system. `varName` is the prefix for populated variables. As the result of the call `USE_SYSTEM_${varName}` is set to TRUE if the compatible version of the dependency is found in system. It is a `CACHE` variable, it means the one in `CMakeCache.txt` (the file edited when we set the variables via command line when initially invoking `cmake` or via `cmake-gui`) takes precedence.  After `initDependencySourceChoice(FMT fmt VERSION "8.0.0")` `USE_SYSTEM_FMT` should be
1. if there is a value in the cache - the one from cache.
2. if there is a compatible version found - it sets the cache var into `ON` by default
3. else - it sets the cache var into `OFF` by default
Then the cache var is used according to either defaults, or to user's overrides.

I think that using the deps from distros is beneficial and distros' policy usually require software use deps from the packages in the distros, and I consider it a right thing, because there are certain benefits:
* It spares ones building software from downloading and recompiling the dependencies.
* it brings low memory footprints (in the case of dynamic linkage)
* it allows to deliver updates to deps without involving the devs of software using them.
* the deps can have distro-specific patches that modify their behaviour to integrate into the distro better
* the dev cannot patch the deps and has to upstream improvements. If the improvements are not really improvements but workarounds of the dev bugs, the dev has to fix the bugs in own code.

There are drawbacks:
* Using dependencies from system results in not using the exact version of the dependency used within a submodule, but a version called "compatible". It may be not the versions of deps the dev himself tests his software with.
* A builder has to use distro package manager to install the deps dev packages (a user usually can rely on the package metadata).
* the dev cannot patch the deps and has to upstream improvements. If the improvements are not really improvements but workarounds of the dev bugs, the dev has to fix the bugs in own code.

This way I think the benefits of using system-installed packages outweight drawbacks.

iTrooz:
Hey, I'm going to take a look at this when I get time, feel free to ping me if I didn't look at it in 7 days

Also, could you take a look at the CI ? I think your PR breaks it

KOLANICH:
For some of the deps (nfd, LLVM) using system-installed packages is not enabled because of the issues in the CI scripts and some of the deps themselves. I feel like it is easier to deal with those issues after this PR is landed.

iTrooz:
Hey, in the end I don't think this option is going to be useful, so it doesn't justify adding so much code in CMake.
I am closing this.

Thanks anyways !

KOLANICH:
If deps from submodules had been fully dropped and it would have been much less code. When one has to support multiple ways of doing the same thing, some overhead is inevitable.

iTrooz:
> When one has to support multiple ways of doing the same thing, some overhead is inevitable.

I totally agree with this

But in this case, I don't think it is worth it to support this. When we/packagers will build ImHex, we want maximum reproducibility, so we will continue to manually select system-provided dependencies

PR #1097, opened 2023-05-22
remove fedora 36 builds

jonathanspw:
Fedora 36 is EOL.

Build roots for it were taken offline this morning so builds against it will fail now.


WerWolv:
Thanks!

PR #1101, opened 2023-05-24
Also display titlebar buttons when not running borderless

iTrooz:
Draft for now because I am waiting for feedback from Jay

### DIFF
--- a/main/source/window/window.cpp
+++ b/main/source/window/window.cpp
@@ -152,6 +152,8 @@ namespace hex {
         this->exitGLFW();
     }
 
+    static std::terminate_handler originalHandler;
+
     void Window::registerEventHandlers() {
         // Initialize default theme
         EventManager::post<RequestChangeTheme>("Dark");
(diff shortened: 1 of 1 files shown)


iTrooz:
Uh, not sure what happened with the title of the PR

iTrooz:
Anyway, should be ready for merge !

main/source/window/window.cpp
```diff
@@ -243,12 +245,11 @@ namespace hex {
...
             }
 
-            // save crash.json file
-            saveCrashFile();
+            // the handler should eventually release a signal, which will be caught and used to handle the crash
+            originalHandler();
 
-            // send the event. It may affect things (like the project path),
-            // so we do this after saving the crash file
-            EventManager::post<EventAbnormalTermination>(0);
```
WerWolv, line 251: And this?

iTrooz:
Yep ! The idea is that the original handler should emit an ABORT signal
https://en.cppreference.com/w/cpp/error/terminate_handler says
> The C++ implementation provides a default std::terminate_handler function, which calls [std::abort](https://en.cppreference.com/w/cpp/utility/program/abort)()

I agree that this seems a little dangerous, but.. well, the standard says so

PR #1104, opened 2023-05-24
Add a native error message when glfw window creation fails

iTrooz:
Draft because I want to test it again tomorrow with all OSes, first

### DIFF
--- a/main/include/window.hpp
+++ b/main/include/window.hpp
@@ -14,6 +14,8 @@ struct ImGuiSettingsHandler;
 
 namespace hex {
 
+    void nativeErrorMessage(std::string&);
+
     class Window {
     public:
         Window();
--- a/main/include/window.hpp
+++ b/main/include/window.hpp
@@ -14,7 +14,7 @@ struct ImGuiSettingsHandler;
 
 namespace hex {
 
-    void nativeErrorMessage(const std::string&);
+    void nativeErrorMessage(const std::string &message);
 
     class Window {
     public:
--- a/main/include/window.hpp
+++ b/main/include/window.hpp
@@ -14,7 +14,7 @@ struct ImGuiSettingsHandler;
 
 namespace hex {
 
-    void nativeErrorMessage(std::string&);
+    void nativeErrorMessage(const std::string&);
 
     class Window {
     public:
--- a/main/source/init/splash_window.cpp
+++ b/main/source/init/splash_window.cpp
@@ -1,3 +1,4 @@
+#include "window.hpp"
 #include "init/splash_window.hpp"
 
 #include <hex/api/imhex_api.hpp>
@@ -216,6 +217,7 @@ namespace hex::init {
         // Create the splash screen window
         this->m_window = glfwCreateWindow(1, 400, "Starting ImHex...", nullptr, nullptr);
         if (this->m_window == nullptr) {
+            hex::nativeErrorMessage("Failed to create GLFW window. <TODO explain what happened and instructions on what to do>");
             log::fatal("Failed to create GLFW window!");
             exit(EXIT_FAILURE);
         }
--- a/main/source/init/splash_window.cpp
+++ b/main/source/init/splash_window.cpp
@@ -217,7 +217,7 @@ namespace hex::init {
         // Create the splash screen window
         this->m_window = glfwCreateWindow(1, 400, "Starting ImHex...", nullptr, nullptr);
         if (this->m_window == nullptr) {
-            hex::nativeErrorMessage("Failed to create GLFW window. <TODO explain what happened and instructions on what to do>");
+            hex::nativeErrorMessage("Failed to create GLFW window.");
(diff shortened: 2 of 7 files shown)


main/include/window.hpp
```diff
@@ -14,6 +14,8 @@ struct ImGuiSettingsHandler;
 
 namespace hex {
 
+    void nativeErrorMessage(const std::string&);
```
WerWolv, line 17: Give the parameter a name here.

main/source/init/splash_window.cpp
```diff
@@ -216,6 +220,12 @@ namespace hex::init {
         // Create the splash screen window
         this->m_window = glfwCreateWindow(1, 400, "Starting ImHex...", nullptr, nullptr);
         if (this->m_window == nullptr) {
+            hex::nativeErrorMessage(hex::format(
+                "Failed to create GLFW window: [{}] {}.\n"
+                "You may not have a renderer available.\n"
+                "The most common cause of this is using a virtual machine\n"
+                "You may want to try an artifact ending with 'NoGPU'\n"
+                , lastGlfwError.first, lastGlfwError.second));
```
WerWolv, line 228: I think instead of "Artifact" I'd use "Release" or "Build" here. Might be easier to understand

Also instead of using .first and .second, consider either defining a new struct that you use instead of the std::pair or use structured bindings to decompose the pair and give the two variables better names

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
     #include <unistd.h>
 
     #include <imgui_impl_glfw.h>
+    #include <string.h>
+    #include <ranges>
 
 namespace hex {
 
+    bool isFileInPath(std::fs::path filename){
```
WerWolv, line 25: Should be a const reference

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
...
+        auto path_var = std::string(getenv("PATH"));
```
WerWolv, line 26: We have `hex::getEnvironmentVariable` which is a bit safer than this here

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
...
+        for(auto dir : std::views::split(path_var, ':')){
```
WerWolv, line 27: Missing spaces after for and if and before {

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
...
+            if(std::fs::exists(std::fs::path(std::string_view(dir)) / filename)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    void executeCmd(std::vector<std::string> strVector) {
```
WerWolv, line 35: Pass in as const reference and maybe give the parameter a name that describes what it does better

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
...
+        std::vector<char*> cVector;
+        for (std::string& str : strVector) {
```
WerWolv, line 37: const auto&

main/source/window/linux_window.cpp
```diff
@@ -17,9 +17,43 @@
...
+            exit(EXIT_FAILURE);
+        }
+    }
+
+    void nativeErrorMessage(const std::string &message) {
+        if (isFileInPath("zenity")) {
+            executeCmd({"zenity", "--error", "--text", message});
+        } else if(isFileInPath("notify-send")) {
+            executeCmd({"notify-send", "-i", "script-error", "Error", message});
+        } // hopefully one of these commands is installed
```
WerWolv, line 54: Maybe as a fallback, just print the error message to the console using log::fatal?

WerWolv (changes requested):

main/source/window/linux_window.cpp
iTrooz: Humm, now that you talk about it, I'd like the message to *always* appear in the console, independently from the GUI.

So I propose to simply log the message a second time after calling nativeErrorMessage(), and keep the current behaviour of "do nothing if no command was found" in the function. What do you think ?

main/source/init/splash_window.cpp
iTrooz: Honestly I think all three of them are difficult to understand. How about "Release artifact" ? (Yup, it's literally a combination, but it feels easier to understand to me)

WerWolv: Sure yeah

main/source/window/linux_window.cpp
WerWolv: Doesn't really make sense to separate it if you want the message to be logged to the console whenever that error popup is shown

iTrooz: you're right. Done !

WerWolv (approved):

PR #1105, opened 2023-05-24
save opened projects as recent entries

iTrooz:
This PR does two things :
- save opened projects as recent entries
- refactor stuff about recent entries in a separate file. The reason is that I felt like welcome_screen.cpp was really big ( 685 lines before this, 500 now). What do you think ?

### DIFF
--- a/plugins/builtin/include/content/recent.hpp
+++ b/plugins/builtin/include/content/recent.hpp
@@ -9,11 +9,32 @@
 #include <wolv/io/fs.hpp>
 
 namespace hex::plugin::builtin::recent {
+    
+    /**
+     * @brief Structure used to represent a recent other
+     */
     struct RecentEntry {
+        
+        /**
+         * @brief Name that should be used to display the entry to the user
+         */
         std::string displayName;
+
+        /**
+         * @brief type of this entry. Might be a provider id (e.g. hex.builtin.provider.file)
+         * or "project" in case of a project
+         */
         std::string type;
+
+        /**
+         * @brief path of this entry file
+         */
         std::fs::path filePath;
 
+        /**
+         * @brief Entire json data of the recent entry (include the fields above)
+         * Used for custom settings set by the providers
+         */
         nlohmann::json data;
 
         bool operator==(const RecentEntry &other) const {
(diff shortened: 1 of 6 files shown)


plugins/builtin/include/content/recent.hpp
```diff
@@ -0,0 +1,81 @@
+#pragma once
+
+#include<string>
```
WerWolv, line 3: Missing space after include

plugins/builtin/source/content/recent.cpp
```diff
@@ -0,0 +1,230 @@
...
+    static std::list<RecentEntry> s_recentEntries;
+
+    void registerEventHandlers() {
+        // Save every opened provider as a "recent" shortcut
+        (void)EventManager::subscribe<EventProviderOpened>([](prv::Provider *provider) {
+            if (ContentRegistry::Settings::read("hex.builtin.setting.general", "hex.builtin.setting.general.save_recent_providers", 1) == 1) {
+                auto fileName = hex::format("{:%y%m%d_%H%M%S}.json", fmt::gmtime(std::chrono::system_clock::now()));
+
+                // do not save to recents if the provider is part of a project
+                if(ProjectFile::hasPath())return;
```
WerWolv, line 31: Missing space

plugins/builtin/source/content/recent.cpp
```diff
@@ -0,0 +1,230 @@
...
+            if (!s_recentEntriesUpdating) {
+                auto it = s_recentEntries.begin();
+                while (it != s_recentEntries.end()) {
+                    const auto &recentEntry = *it;
+                    bool shouldRemove = false;
+
+                    ImGui::PushID(&recentEntry);
+                    ON_SCOPE_EXIT { ImGui::PopID(); };
+
+                    if (ImGui::BulletHyperlink(recentEntry.displayName.c_str())) {
```
WerWolv, line 175: Would it maybe make sense to use different icons depending on if the entry is a file or a full project?

plugins/builtin/source/content/recent.cpp
```diff
@@ -0,0 +1,230 @@
...
+                if (!found)
+                    wolv::io::fs::remove(path);
+            }
+
+            std::copy(uniqueProviders.begin(), uniqueProviders.end(), std::front_inserter(s_recentEntries));
+        });
+    }
+
+    void loadRecentEntry(const RecentEntry &recentEntry) {
+        if(recentEntry.type == "project"){
```
WerWolv, line 138: Missing space

WerWolv (changes requested):

plugins/builtin/source/content/recent.cpp
iTrooz: Yeah ! I came up with this, but 
image

I'd like to make something more complicated (like a separate color) but I don't think the ImGui extensions would allow me to do that easily. I think I would like to keep it that way for now, and submit another PR later to make the extensions more modular

WerWolv (approved):

PR #1108, opened 2023-05-26
Starting loop at 9 and ending it at normals.size() prevents the norma…

paxcut:
…ls from the last triangle from ever being computed resulting in 0 normals that created visual artifacts in some models. Re-indexing the loop to start at zero solves the problem.


WerWolv:
Thanks!

PR #1115, opened 2023-05-27
Handle crashes that may happen before the main loop

iTrooz:
Draft because I absolutely do not trust myself writing good code at 2AM. I will review it tomorrow


iTrooz:
Should be good to merge !

PR #1131, opened 2023-06-08
do not create file if it doesn't exist when trying to open file provider

iTrooz:
Draft until I also do it for project loading, just in case I change my mind about the strategy


iTrooz:
Integrated in https://github.com/WerWolv/ImHex/pull/1135

Issue #1106, opened 2023-05-24
[Bug] Remove file content will result in leftover duplicated content and empty bytes.


qux-bbb:
### Operating System

Windows

### What's the issue you encountered?

When I remove a file header content, duplicate content and "\x00" are generated at end of file.  

### How can the issue be reproduced?

prepare a file test.txt, content: "hello"  
If you remove first byte, it will show:  
![remove_first_byte](https://github.com/WerWolv/ImHex/assets/18598419/b9cc32c6-6d8b-4cd3-b43f-bf06637e36dc)  

If you remove 2 bytes, it will show:  
![remove_2_bytes](https://github.com/WerWolv/ImHex/assets/18598419/5776375f-51a0-4f97-87dc-c29c541b455d)  

### ImHex Version

1.29.0

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

### Additional context?

My OS is Windows 11.

qux-bbb:
Sorry for that I am not familiar ImHex source code, so I can not find a fix method.

qux-bbb:
Yeah, @reserge fixed it. Thank you!

PR #1132, opened 2023-06-09
Fix bug #1106

reserge:
The bug appears on Windows, but this change can also be applied to other platforms.


reserge:
already fixed in https://github.com/WerWolv/ImHex/commit/de76c37ffb737752557e0c26d3eda04cf29d3a3f

PR #1135, opened 2023-06-11
move project loading/storing logic to builtin plugin + show error messages to the user when loading a project

iTrooz:
In order to do this I add to make some other additions :
- Add a warning popup (TODO, maybe add some icons to differentiate error/warning popups in a future PR ?)
- create showError() and showWarning() functions, as helpers to show a message both to the logs and as a popup


iTrooz:
Also, to test this properly I add to include https://github.com/WerWolv/ImHex/pull/1131 in this PR. I think it's fair to keep the commit here, and close the other PR.

iTrooz:
Should be ready to merge !

lib/libimhex/include/hex/helpers/tar.hpp
```diff
@@ -45,6 +51,10 @@ namespace hex {
         std::fs::path m_path;
 
         bool m_valid = false;
+        
+        // these will be updated when the constructor is called
+        int m_tarOpenErrno = MTAR_ESUCCESS;
```
WerWolv, line 56: Do these need to be `int`s or is there a better type available for this?

lib/libimhex/source/helpers/tar.cpp
```diff
@@ -20,16 +21,23 @@ namespace hex {
...
 
         this->m_path = path;
-        this->m_valid = (error == MTAR_ESUCCESS);
+        this->m_valid = (tar_error == MTAR_ESUCCESS);
+
+        if (!this->m_valid) {
+            this->m_tarOpenErrno = tar_error;
+            
+            // hopefully this errno corresponds to the file open call in mtar_open
+            this->m_fileOpenErrno = errno;
```
WerWolv, line 39: Does mtar actually use errno? Most things moved away from it a long time ago

plugins/builtin/source/content/project.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+            if (!result && handler.required) {
+                return false;
+            }
+        }
+
+        for (const auto &provider : ImHexApi::Provider::getProviders()) {
+            const auto basePath = std::fs::path(std::to_string(provider->getID()));
+            for (const auto &handler: ProjectFile::getProviderHandlers()) {
+                // handlers are supposed to show the error/warning popup to the user themselves, so we don't show one here
```
WerWolv, line 89: Upper case first letter in comments

plugins/builtin/source/content/providers.cpp
```diff
@@ -55,19 +59,47 @@ namespace hex::plugin::builtin {
...
-                         success = false;
-                     else
+                     if (!provider->open() || !provider->isAvailable() || !provider->isReadable()) {
+                        providerWarnings[provider] = provider->getErrorMessage();
+                     } else
                          EventManager::post<EventProviderOpened>(provider);
                  }
 
-                 return success;
+                 std::string warningMsg;
```
WerWolv, line 79: Weird indentation here?

plugins/builtin/source/content/providers.cpp
```diff
@@ -55,19 +59,47 @@ namespace hex::plugin::builtin {
...
+                for(const auto &warning : providerWarnings){
+                    ImHexApi::Provider::remove(warning.first);
+                    warningMsg.append(
```
WerWolv, line 82: This will cause a new line at the start of the warning list. Is that intended?

WerWolv (changes requested):

lib/libimhex/include/hex/helpers/tar.hpp
iTrooz: Nope, `errno` is a int, and the mtar errors are an anonymous enum https://github.com/WerWolv/ImHex/blob/a20d6aa2b2b6885c9cbc1dadadfdbeb98f962274/lib/external/microtar/include/microtar.h#L20

lib/libimhex/source/helpers/tar.cpp
iTrooz: not directly, but `mtar_open()` uses `fopen()`, which sets errno. And the mtar error itself is not useful by itself ("Failed to open") so we need errno to know exactly what happens ("Permission defined")

plugins/builtin/source/content/project.cpp
```diff
@@ -0,0 +1,167 @@
...
+            showError(hex::format("hex.builtin.popup.error.project.load"_lang,
+                hex::format("hex.builtin.popup.error.project.load.file_not_found"_lang,
+                    wolv::util::toUTF8String(filePath)
+            )));
+            return false;
+        }
+
+        Tar tar(filePath, Tar::Mode::Read);
+        if (!tar.isValid()) {
+            // TODO: make tar return an actual error string
```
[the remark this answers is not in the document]
iTrooz: oops nope, sorry

plugins/builtin/source/content/project.cpp
```diff
@@ -0,0 +1,167 @@
...
+                return false;
+            }
+        }
+
+        auto providers = auto(ImHexApi::Provider::getProviders());
+        for (const auto &provider : providers) {
+            ImHexApi::Provider::remove(provider);
+        }
+
+        // TODO there is a bug here with the var result ?
```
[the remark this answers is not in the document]
iTrooz: I swear I should make a pre-push hook to check for TODOs

The problem that I suspected is that if a non-required provider fails, `result` gets set to false, and then a required provider that succeeds would trigger this branch https://github.com/WerWolv/ImHex/blob/d159e5fa345f4ac17e5f61cda95892c01b7ad44b/plugins/builtin/source/content/project.cpp#L80 , and make the loading fail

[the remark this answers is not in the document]
iTrooz: I removed the TODO, I will check/fix it in a later PR

plugins/builtin/source/content/providers.cpp
iTrooz: Oh, yeah. But the whole indentation is weird in this file, I'm going to fix it in (https://github.com/WerWolv/ImHex/pull/1138 (please merge this one first, it will be easier to re-do the other)

plugins/builtin/source/content/providers.cpp
iTrooz: Yep, that's what I intended. I think it's better so the user clearly sees the providers that fail

plugins/builtin/source/content/project.cpp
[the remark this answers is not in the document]
WerWolv: Oh yeah. We'd probably need two variables. One that states if a required provider has failed and one if any provider has failed. The "any provider failed" variable would then only trigger the error popup while the "required provider failed" variable would trigger that return false

WerWolv (approved):

WerWolv:
Thank you!

PR #1137, opened 2023-06-14
Fix full recompilation when commit hash has changed

KOLANICH:
### Problem description
The current approach adds the macrodefs to the most of compilation calls. This causes full recompilation on change of commit hash and/or branch.

This PR
1. fixes some mistakes in variable names discovered when implementing .2
2. adds the macrodefs only to the compilation calls of the files using them. Its allows `ninja` to bypass the recompilation when it is not really needed.

### Additional things
This can be optimized further by generating a `.c` file with the git-dependent constants only and having a header for them, and just linking the compiled object file. But it is a bigger refactoring, since it introduces a header file. I prefer not to do bigger refactorings myself, maintainers love to just reject them with the reasons like "I can live without that".

### DIFF
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -15,11 +15,6 @@ macro(addDefines)
     endif ()
 
     if (DEFINED GIT_COMMIT_HASH_LONG AND DEFINED GIT_COMMIT_HASH_SHORT AND DEFINED GIT_COMMIT_BRANCH)
-        add_compile_definitions(
-                GIT_COMMIT_HASH_LONG="${GIT_COMMIT_HASH_LONG}"
-                GIT_COMMIT_HASH_SHORT="${GIT_COMMIT_HASH_SHORT}"
-                GIT_BRANCH="${GIT_COMMIT_BRANCH}"
-        )
     else()
         # Get the current working branch
         execute_process(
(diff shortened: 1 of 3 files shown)


WerWolv:
Thank you! I didn't even think about it causing this.

Though, instead of having to individually include the files where we want to have access to that information, maybe it would be better to hardcode it to the `imhex_api.cpp` file and just provide functions there that return the value. That way only a single file needs to have the defines but all of ImHex can access it

iTrooz:
That said, why not directly put them in a dedicated file, like KOLANICH said ? It seems cleaner to me, because else imhex_api.cpp would gain a separate purpose, and I'd like to keep them separated.

@KOLANICH would you feel like doing such a refactor if I told you I loved that idea, and would want to merge it ? I can't promise to like the implementaion, but I really would like to have a separete file dedicated to that.

(All of this said, please wait for @WerWolv 's feedback on my comment before starting anything)

WerWolv:
Absolutely yeah. Having an entirely new cpp file for it makes sense. I'd probably still expose the values through the ImHexApi though. Any thoughts on that?

iTrooz:
Yeah I think it's a good idea

iTrooz:
[By the way @KOLANICH, completely off-topic comment, we often discuss on discord, so if you want to ask more questions, want to contribute often, etc... I think it would be a good idea for you to join the discord]

KOLANICH:
I'm sorry for the lack of response. I have implemented it the similar way, but with a bit of difference (1. the info is put into a dedicated cpp file 2. that cpp file is dedicated for metadata handling and the code within it is reusable (as well as the CMake function), can be used by all the main lib and plugins 3. the data are constants (within a metadata table), so can be looked up statically by their offsets without executing any code). Just I'm a bit busy to finish it right now.

Issue #1146, opened 2023-06-19
[Bug] First character of each line won't display when using  custom encoding.


shenlebantongying:
### Operating System

Linux

### What's the issue you encountered?

First char is missing in hex editor's decoding column

### How can the issue be reproduced?

Open a file with those content, then click File -> Import -> Custom Encoding -> utf8.tbl

```
你好世界！
```

The first char `你` is not displayed.

![image](https://github.com/WerWolv/ImHex/assets/20123683/b275c5d0-2645-4f50-a3e5-e8c4a46fd3fa)

"你" is in the table at: E4BDA0

https://github.com/WerWolv/ImHex-Patterns/blob/ea225edf12927c0dd7f2c3d5bcc2323ffba82caa/encodings/utf8.tbl#LL19061C1-L19061C9

### ImHex Version

v1.29.0

### ImHex Build Type

- [x] Nightly or built from sources

### Installation type

https://aur.archlinux.org/packages/imhex-bin

### Additional context?

Changing “Bytes per line” to certain values mythically fixes the problem.

![image](https://github.com/WerWolv/ImHex/assets/20123683/9a455550-c20d-4606-a78c-5a4609247bd7)

paxcut:
you need to turn on the display of advanced column it is a small icon shown in picture
![image](https://github.com/WerWolv/ImHex/assets/53811119/a219972c-bc49-490e-b431-63777e74eb62)

shenlebantongying:
Thanks for taking a look :blue_heart: 

But the the nightly release build, the problem still exists:

With the default 16 bytes per line, the first char is displayed as a red dot.

![image](https://github.com/WerWolv/ImHex/assets/20123683/a164936a-578f-4e95-9cee-ac0a5e760b3c)

This is what the text supposed to looks like:

![image](https://github.com/WerWolv/ImHex/assets/20123683/d601d309-2ef5-40f9-a29b-c23c620ec03b)


[cjk.txt](https://github.com/WerWolv/ImHex/files/11794009/cjk.txt)

paxcut:
it looks like it is selecting different ways to group the bytes depending on the number of bytes per column.  what encoding do you use to display the correct one? if i use utf-8 in the advanced decoding column it doesnt display either of the two results you see.

paxcut:
when column of bytes is a multiple of 3 it displays the correct value otherwise it is applying the encodings incorrectly because the result should not depend on such factors.

shenlebantongying:
[...]

PR #1150, opened 2023-06-20
fix: wrong lineStartAdresses for custom encoding characters

shenlebantongying:
fix https://github.com/WerWolv/ImHex/issues/1146
related https://github.com/WerWolv/ImHex/commit/3a840c4ceddd050371d744761b0194ffcbfe7773

### DIFF
--- a/plugins/builtin/source/ui/hex_editor.cpp
+++ b/plugins/builtin/source/ui/hex_editor.cpp
@@ -122,35 +122,13 @@ namespace hex::plugin::builtin::ui {
         ImColor color;
     };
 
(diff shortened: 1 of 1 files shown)


shenlebantongying:
I think the code is ready to be merged. @WerWolv @paxcut 

The idea is as follows:

* In every line, read from the current line's `m_encodingLineStartAddresses`
  * Increment `offset` by `advance`for every character recognized
  * Repeat until `offset` is larger than `bytesPerRow`
* Set the current line's surplus as the next line's `m_encodingLineStartAddresses`

Also, consider edge case situation:

If the current line's `m_encodingLineStartAddresses` is bigger than `bytesPerRow`, then there is nothing to read in the current line. Just output a red dot.


image

image

WerWolv:
Thanks a lot!

PR #1172, opened 2023-06-30
add subcommands system to CLI

### LINKED ISSUE #948: [Bug] Confusing behaviour when ImHex is ran from a terminal
### Operating System

Linux

If running from the terminal, attempting to open a file that does not exist will result in ImHex trying to open that file and failing to do so, resulting in a window that looks like this:

image

Reloading the file in ImHex just closes the file.

Additionally, if one was to try running ImHex with options passed, ImHex will treat those options as the name of the file it's supposed to be opening. Admittedly this part isn't necessarily a bug, but it is confusing behaviour none-the-less. If I run ImHex with the command `imhex --version` I'd expect ImHex to output it's version or at least tell me that the option was unrecognized, but instead ImHex opens a non-existent file named `--version`.

Assuming there is no file named `non-existent.file`, simply run the following in a terminal: `imhex non-existent.file`

### ImHex Version

1.27.1 f435191

### ImHex Build Type

### Installation type

AUR on both imhex-git and imhex-bin

### Additional context?

_No response_

iTrooz:
System design has been discussed on discord

Should fix #948


lib/libimhex/include/hex/plugin.hpp
```diff
@@ -21,3 +24,16 @@
         GImGui = ctx;                                                                                      \
     }                                                                                                      \
     extern "C" [[gnu::visibility("default")]] void initializePlugin()
+
+/**
+ * This macro is used to define subcommands defined by the plugin/
```
WerWolv, line 29: Stray `/`

lib/libimhex/include/hex/plugin.hpp
```diff
@@ -21,3 +24,16 @@
...
+ * A subcommand consists of a key, a description, and a callback
+ * The key is what the first argument to ImHex should be, prefixed by `--`
+ * For example, if the key if `help`, ImHex should be started with `--help` as its first argument to trigger the subcommand
+ * when the subcommand is trigerred, it's callback will be executed. The callback is executed BEFORE most of ImHex initialization
+ * so to do anything meaningful, you should subscribe to an event (like EventImHexStartupFinished) and run your code there.
+ */
+#define IMHEX_PLUGIN_SUBCOMMANDS() IMHEX_PLUGIN_SUBCOMMANDS_IMPL()
+
+#define IMHEX_PLUGIN_SUBCOMMANDS_IMPL()                                                                    \
+    [[gnu::visibility("default")]] std::vector<hex::SubCommand> subCommands
```
WerWolv, line 39: Is this truly portable? Won't name mangling apply to this in some cases? 
I think it would be better to define this variable as `static` and have a getter function similar to `getPluginName` which returns the commands.

I'm also not entirely sure how well this will work with plugins that are built out-of-tree since if the layout of std::vector is different in the plugin and the main application, this will segfault

lib/libimhex/source/subcommands/sub_commands.cpp
```diff
@@ -0,0 +1,40 @@
+#include<iostream>
+#include<stdlib.h>
+
+#include <hex/api/plugin_manager.hpp>
+#include <hex/helpers/logger.hpp>
+
+namespace hex::init {
+    void processArguments(std::vector<std::string> args){
```
WerWolv, line 8: Avoid copies here by passing as const&

lib/libimhex/source/subcommands/sub_commands.cpp
```diff
@@ -0,0 +1,40 @@
...
+        if (args.size() == 0) return;
```
WerWolv, line 9: `args.empty()`

lib/libimhex/source/subcommands/sub_commands.cpp
```diff
@@ -0,0 +1,40 @@
...
+        std::optional<SubCommand> selectedSubCommand;
+
+        // default subcommand if no subcommand has been specified
+        std::optional<SubCommand> openSubCommand;
+
+        for (auto &plugin : PluginManager::getPlugins()) {
+            for (auto &subCommand : plugin.getSubCommands()) {
+                if (hex::format("--{}", subCommand.commandKey) == args[0]) {
+                    if (selectedSubCommand) {
+                        log::error("Multiple subcommands entered: {} and {}", selectedSubCommand->commandKey, args[0]);
```
WerWolv, line 20: Might be nice to have the option to specify multiple commands that are being executed one after the other. Not sure if that makes sense though

main/source/main.cpp
```diff
@@ -10,15 +10,33 @@
...
+    }
+}
+
 int main(int argc, char **argv, char **envp) {
-    using namespace hex;
     hex::crash::setupCrashHandlers();
     ImHexApi::System::impl::setProgramArguments(argc, argv, envp);
 
+    std::vector<std::string> args;
+    for(int i=1;i<argc;i++){
```
WerWolv, line 33: A bit weird formatting here

Also it might make sense to use `std::string_view` for all the arguments.
argc will be valid for the entire duration of the program so it's not necessary to copy the content around constantly

plugins/builtin/source/plugin_builtin.cpp
```diff
@@ -40,8 +43,41 @@ namespace hex::plugin::builtin {
 
     void handleBorderlessWindowMode();
 
+    void handleSubCommands();
+
 }
 
+IMHEX_PLUGIN_SUBCOMMANDS() {
+    { "help", "Print help about this command", [](const std::vector<std::string>&) {
+        printf(
```
WerWolv, line 52: Use `hex::print` where needed

plugins/builtin/source/plugin_builtin.cpp
```diff
@@ -40,8 +43,41 @@ namespace hex::plugin::builtin {
...
+                printf("\t--%s\t\t%s\n", subCommand.commandKey.c_str(), subCommand.commandDesc.c_str());
+            }
+        }
+        exit(EXIT_SUCCESS);
+    }},
+    { "open", "Open files passed as argument. This is the default subcommand is none is entered",
+        [](const std::vector<std::string> &args) {
+        bool doubleDashFound = false;
+        for (auto &arg : args) {
+            if (arg == "--" && !doubleDashFound) {
```
WerWolv, line 70: Not sure if I understand this code entirely. Does this mean people need to type `imhex -- fileToOpen.bin` when opening a file?  
I think this will break `Open As...` in the file explorer and drag-n-dropping files onto the executable

WerWolv (changes requested):

lib/libimhex/include/hex/plugin.hpp
iTrooz: > Is this truly portable? Won't name mangling apply to this in some cases?

Honestly, I don't know C++ enough to tell

> I'm also not entirely sure how well this will work with plugins that are built out-of-tree since if the layout of std::vector is different in the plugin and the main application, this will segfault

Why would the layout of std::vector change, but not symbol names, or the size of other types ?

lib/libimhex/source/subcommands/sub_commands.cpp
iTrooz: The problem is that I modify the vector in that function, so I can't use `const`. What would be the cleanest thing to do her ?

lib/libimhex/source/subcommands/sub_commands.cpp
iTrooz: Yeah, that could make sense ! I'm going to do that

lib/libimhex/source/subcommands/sub_commands.cpp
iTrooz: Okay, https://github.com/WerWolv/ImHex/pull/1172#discussion_r1248768557 means I'm going to make multiple vectors anyway, so this problem isn't really relevant now

plugins/builtin/source/plugin_builtin.cpp
iTrooz: Essentially, this code just means we skip the first argument named `--`. It's kinda ugly, but I just really wanted to implement the fact that that `--` exists (because IIRC I've already seen it send to programs by the user environment)

So both `imhex -- fileToOpen.bin` and `imhex fileToOpen.bin` will work

lib/libimhex/source/subcommands/sub_commands.cpp
iTrooz: What do you think about the format `imhex --open file.txt /home/itrooz/file2.txt --othercmd --run-process ls al` ?

lib/libimhex/include/hex/plugin.hpp
WerWolv: All the other functions just return pointers to trivial types whose layout will never change.
The size and layout of std::vector is not fixed and can change between standard library implementations and versions. 

The portable way to do it would be to make a C API around it

lib/libimhex/source/subcommands/sub_commands.cpp
WerWolv: Ah if the data is being modified it's alright. I didn't see that, sorry

lib/libimhex/source/subcommands/sub_commands.cpp
WerWolv: Yeah I think that's okay. 
Also if you just run e.g `--version` multiple commands that don't require ImHex to start up fully, it would be nice if ImHex just exited after finishing the commands. Is that already how it works?

iTrooz: Right now, `--help` calls exit() itself. We could do the same with `--version` without problem

WerWolv: That would make `--version --help` only execute `--version` though

iTrooz: The user would need to do `imhex --version --othercmd --help` anyway to execute both, and I don't think I've ever seen programs that handle both options like this.

lib/libimhex/include/hex/plugin.hpp
iTrooz: Does this mean that the callback argumentscan't be a vector too ?

WerWolv: No that's okay. E.g returning a pointer to the struct would be okay

iTrooz:
Drafted because I realized there's going to be some problems with Windows support. You have some kind of "message sending" to existing window I haven't touched at all

WerWolv:
The message sending stuff is simply there that when you `Right click -> Open with -> ImHex` a file while there's already an ImHex window open, it will open that file in the open window instead of opening a new one. 
The receiving end is entirely implemented in the windowProc function and doesn't need to be touched at all I think. The sending end would need to be moved to the `--open` callback

iTrooz:
In that case we should probably expose an API so other plugins can do such a thing

WerWolv:
What features would be useful there? I can only really think of sending file paths over
Also this is highly OS dependent and is only implemented on Windows right now

iTrooz:
For example, What if Imhex is running and someone does `imhex --open a.txt --othercmd --run-process firefox` ? In that case, ImHex would open the file in the running instance, and open the process in a new one.

I think every callback should have the choice to forward data to a running instance

lib/libimhex/include/hex/api/event.hpp
```diff
@@ -232,4 +232,15 @@ namespace hex {
...
+
+    /**
+     * @brief Send an event to the main Imhex instance
+     */
+    EVENT_DEF(SendMessageToMainInstance, const std::string, const std::vector<u8>&);
+
+    /**
+     * @brief Checks if we are running in the main instance
+     */
+    EVENT_DEF(IsMainInstance, bool&);
```
WerWolv, line 245: This should probably rather be a getter-setter pair in `ImHexApi::System`. Similar to e.g the Portable version flag functions

iTrooz: The problem is that we need to communicate with main to get this information (since main is the one doing the work), so we still need an event subscriber in main

main/source/window/win_window.cpp
```diff
@@ -51,12 +53,28 @@ namespace hex {
                 auto message = reinterpret_cast<COPYDATASTRUCT *>(lParam);
                 if (message == nullptr) break;
 
-                auto data = reinterpret_cast<const char8_t *>(message->lpData);
-                if (data == nullptr) break;
+                int nullIndex = -1;
```
WerWolv, line 56: Should probably be a `ssize_t`

main/source/window/win_window.cpp
```diff
@@ -51,12 +53,28 @@ namespace hex {
...
+                    log::warn("Received invalid forwarded event");
+                    break;
+                }
+
+                std::string evtName(messageData, nullIndex);
 
-                std::fs::path path = data;
-                log::info("Opening file in existing instance: {}", wolv::util::toUTF8String(path));
-                EventManager::post<RequestOpenFile>(path);
+                std::vector<u8> evtData(messageData+nullIndex+1, messageData+messageSize);
```
WerWolv, line 75: Missing spaces between operands

main/source/main.cpp
```diff
@@ -4,20 +4,39 @@
...
+#include <hex/api/content_registry.hpp>
+#include <hex/helpers/fs.hpp>
+#include "hex/subcommands/subcommands.hpp"
 
 #include <wolv/io/fs.hpp>
 #include <wolv/utils/guards.hpp>
 
+using namespace hex;
+
+void initPlugins() {
```
WerWolv, line 24: This function doesn't actually initialize any plugins, it just loads them into our address space. The init function isn't being called until later when the Splash Screen is displayed

plugins/builtin/source/plugin_builtin.cpp
```diff
@@ -41,8 +45,58 @@ namespace hex::plugin::builtin {
 
     void handleBorderlessWindowMode();
 
+    void handleSubCommands();
+
 }
 
+IMHEX_PLUGIN_SUBCOMMANDS() {
+    { "version", "Print ImHex version and exit", [](const std::vector<std::string>&) {
+        std::string version_str = hex::format("{} (Commit {}@{})", ImHexApi::System::getImHexVersion(), ImHexApi::System::getCommitBranch(), ImHexApi::System::getCommitHash());
```
WerWolv, line 54: `versionString`

WerWolv (changes requested):

lib/libimhex/include/hex/api/event.hpp
WerWolv: Yeah but main can just set that flag at the start once in the ImHexApi. Just like the portable version flag.
And then it can be queried from there. That flag will never change afterwards

main/source/messaging/common.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <hex/api/imhex_api.hpp>
+#include <hex/api/event.hpp>
+#include <hex/helpers/logger.hpp>
+
+#include "messaging.hpp"
+
+namespace hex::messaging {
+
+    std::optional<bool> s_isMainInstance;
```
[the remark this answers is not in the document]
iTrooz: Could you explain me why ? I don't really understand what "internal linkage" (which is what `static` means apparently ?) does

lib/libimhex/source/api/imhex_api.cpp
```diff
@@ -621,4 +595,35 @@ namespace hex {
...
+
+                return handlers;
+            }
+
+            void runHandler(const std::string &evtName, const std::vector<u8> &args) {
+                const auto& handlers = impl::getHandlers();
+                auto matchHandler = handlers.find(evtName);
+                
+                if (matchHandler == handlers.end()) {
+                    throw std::runtime_error(hex::format("Forward event handler {} not found", evtName));
```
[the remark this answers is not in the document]
WerWolv: Fast-fail is okay but here people might lose a lot of progress if ImHex just crashes randomly when doing something

main/source/messaging/linux.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <hex/helpers/intrinsics.hpp>
+#include "messaging.hpp"
+
+namespace hex::messaging {
+
+    void sendToOtherInstance(const std::string &evtName, const std::vector<std::string> &args) {
+        hex::unused(evtName);
+        hex::unused(args);
+        std::logic_error("Not implemented");
```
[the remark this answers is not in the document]
iTrooz: This function should not even be called on Linux because https://github.com/WerWolv/ImHex/blob/e5473e9d8c3eb3a247eed760c4422eef6f75ee58/main/source/messaging/linux.cpp#L16 always returns true.

Calling this function on Linux/MacOS is a real bug from our part

I'll still replace it with a log::error() since we really don't want to crash ImHex, even if this would be a bug

main/source/messaging/common.cpp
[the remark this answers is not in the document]
WerWolv: Internal linkage means "This variable cannot be accessed from any other file. It's a private local variable for this file."
If it's not static, it means this is a global variable and you can use `extern std::optional<bool> s_isMainInstance;` from another file to access it. We want to avoid this to not pollute the global namespace basically

main/source/messaging/linux.cpp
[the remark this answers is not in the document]
iTrooz: *Plus, I realized that I'm really dumb, and didn't even throw the exception. No idea what I was even thinking when making this code*

lib/libimhex/include/hex/api/event.hpp
iTrooz: Done !

WerWolv (approved):

PR #1174, opened 2023-07-01
reset terminate handler directly when being called + some other crashes to crash handling

iTrooz:
This PR fixes some things about crash handling:
- when the terminate handler is called, immediately set it back to the original one, so can't make a recursion if the crash-handling code fails
- Only save projects if the crash occured after Imhex finished startup
- do not update the project location when saving the crash backup file: this will remove problems when `EventAbnormalTermination` is called before `crashCallback()`

I also added a bit more documentation


WerWolv:
Thank you!

PR #1183, opened 2023-07-03
separate the behaviour of being savable and being dumpable for provider

iTrooz:
### Problem description

Currently, the providers use the method `isSavable()` to determine both if they can use "Save" or "Save as".
This behaviour is problematic because some providers may need to be saveable but not saveable as: for example the view provider. The original provider may not allow to be saved.

I separate these two behaviour by creating another function: `isDumpable()`, that return true by default but can be overridden by the provider to return false, if the provider should not be dumped in any way.

### Additional things

While I was at it, I also marked "export" operations as needing the "dumpable" flag. That way, we can't accidentally export the whole address space of a process as base64.

I also added documentation for these some functions in Provider


WerWolv:
Thanks!

PR #1188, opened 2023-07-12
Add subcommands system

iTrooz:
As discussed on discord

Commits are kind of a mess, you should squash them


iTrooz:
Oops, I forgot I already opened a PR for that

Moving this to #1172 then I guess

PR #1240, opened 2023-08-09
Made result of demangler selectable

paxcut:
and can be highlighted to select portions. The string cannot be edited and the appearance is kept the same. Having to type those long names can be a pain so this is to avoid that.
The code uses Imgui multi line text input but taking care to remove all the unwanted effects that a naive application would cause.


paxcut:
This PR was steamrolled without me being aware so I feel it should be closed since there isn't any point to it anymore

PR #1247, opened 2023-08-11
Reverts the FPS limiter back to the one used in v1.30.1

LennardF1989:
As requested on Discord, a PR to revert the FPS limiter back to the one that was used in v1.30.1.

The new FPS limiter seems to be flawed in that it runs at about half the speed it is supposed to be.

See this illustration:
FPS

Left is v1.30.1, right is the new version (without this fix). See how long it takes to respectively reach 0xE90.

This is not a performance issue, because when you fully unlock the framerate on the right, it's just as fluent as on the left.


WerWolv:
Thanks!

Issue #1210, opened 2023-07-27
[Bug] Saving a file doesn't change the time/date stamp on Windows


gmestanley:
### Operating System

Windows

### What's the issue you encountered?

I was editing the executable of a game that I am translating, and decided to save it by clicking File -> Save (as key shortcuts hadn't been working well at that time). When I went to check the executable's timedate stamp, which is something I check everytime I save as I perform version control with the executable, I noticed that the timedate stamp wasn't altered and was still the same from the last time I had saved the file. This still happens in the current version, and goes back at most to 1.26.0.

### How can the issue be reproduced?

- Open a file with ImHex.
- Modify it.
- Save it with either File -> Save or CTRL+S.
- The file was saved, but the timedate stamp was not changed.

### ImHex Version

1.28.0

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

### Additional context?

Using 64-bit Windows 10.

iTrooz:
I can reproduce this on Windows, but not Linux. So I'm going to let another team member that works on Windows fix this issue

lorsanta:
I was looking into this issue, and I found the problem is caused by the [`MapViewOfFile`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile) function.

From what I've understood, the editor keeps a memory-mapped file that it updates when the file is saved.
https://github.com/WerWolv/ImHex/blob/ba9227c1e00d54516da8a4a3683685b795d6c33c/plugins/builtin/source/content/providers/file_provider.cpp#L70
Contrary to Linux and macOS, which use `mmap`, Windows uses `MapViewOfFile` that doesn't necessarily update the last modification timestamp automatically, as the documentation states.

If we really want to fix this bug, we could do as the documentation suggests and use [`SetFileTime`](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-setfiletime) to update the timestamp.

iTrooz:
It seems like this would fix it !
Would you make a PR for this ?

lorsanta:
Sure! I'll work on it.

PR #1248, opened 2023-08-11
fix: update timestamp when saving a file in windows

lorsanta:
### Problem description
Ref #1210 

Call [`SetFileTime()`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfiletime) everytime `FileProvider::save()` is called.

### Additional things
I moved the call to `File::close()` from `FileProvider::open()` to `FileProvider::close()` because `SetFileTime()` requires a file handler as input, so I need `File::m_file` to be valid.

### DIFF
--- a/plugins/builtin/source/content/providers/file_provider.cpp
+++ b/plugins/builtin/source/content/providers/file_provider.cpp
@@ -16,6 +16,10 @@
 
 #include <nlohmann/json.hpp>
 
+#if defined(OS_WINDOWS)
+    #include <windows.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


WerWolv:
Hey, thanks a lot for the PR!

Moving the call to File::close() to the close function is a bit problematic sadly. Basically, if you keep the file handle open until the provider is closed, you'll prevent many applications from opening the file while ImHex has it open. Instead, it's probably better to re-open the file, update the timestamp and then close it again

WerWolv:
Thank you!

PR #1253, opened 2023-08-18
Return early to remove nested code

aedrax:
### Problem description

This will simplify the codebase in the pattern drawer. It returns early on conditional statements and reduces the amount of nested code making it easier to read and track while developing.


iTrooz:
Assigning this to WerWolv because this PR is about code style

WerWolv:
Thank you!

PR #1268, opened 2023-08-28
feat: Adapt content store view for new API contents

iTrooz:
### Description
This PR modifies the content store view to accommodate for the new content the API will send, the pattern description and authors.
For that, I changed the "description" column to be wrapped (because the description might be so long it takes multiple lines) and added a "authors" column

### Screenshots
Before
image

After
image


iTrooz:
Needs https://github.com/WerWolv/WerWolvAPI/pull/7

Issue #1265, opened 2023-08-26
[Bug] Can't open any disk or partition


zorone:
### Operating System

Windows

### What's the issue you encountered?

I can't open any disk or partition at all. It shows weird dialog screen with unreadable text. And I can't click out from it at all. I have to close and open it again to interact with them.

![Screenshot (4696)](https://github.com/WerWolv/ImHex/assets/60736083/49e92b67-ac9f-4c3d-8ea5-f3729aa5592c)

### How can the issue be reproduced?

I don't remember it that clearly. I couldn't even remember if the program that I copied from was able to use this feature or not. But I remember a bit that I copied this portable program, that I already installed everything from content store.  (Which has been used from another computer before.) And I ran it. Everything seems to work as a first glance. Then I updated everything I could be able to update in content store. When I tried to use it, it happened. I tried to remove and reinstall but it still did not work.

### ImHex Version

1.30.1

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

Portable, then remove and use MSI.

### Additional context?

I can still open files from it, but not for disk and volume.
Also, can you improve searching speed? It is painfully slow to be able to bear.

WerWolv:
Hey
That's a very weird issue, it works fine on my machine. That popup is set to always auto resize so it should always display all the content in it. Could you try deleting `%localappdata%/imhex/config/interface.ini`?

About the search speed, could you give some more context? Searching is reasonably fast for me too. Even in a debug build, it searches through a 120MB file in about 10 seconds

zorone:
Wow, you were right. It just solved my issue completely. I am too amazed that it turns out like that.
![Screenshot (4698)](https://github.com/WerWolv/ImHex/assets/60736083/3c026de3-4912-4dcc-8d72-a48eefd95fda)

For your info, I also put interface.ini here too.
[https://pastebin.com/KGFysmyT](https://pastebin.com/KGFysmyT)

I assume it might be caused by wrongly set from different system. And it didn't handle it properly to adapt to new system, I guess?

About file searching speed, maybe I might be too biased in that manner. I just feel like HxD is faster. I don't really have any evidence though. Maybe I could try to set a timer and count for that later. And also open new issue, I supposed?

I don't know if I should close this issuse or not, because we didn't do anything and it just fixed itself before my eyes. It could still happen later too. You can close this issue if you want, though.

zorone:
Wait, it happened again...
[...]

PR #1272, opened 2023-08-29
Fixed error popup doesn't show immediately after click.

zorone:
From #1265, Looks like Error Popup doesn't handle properly in some circumstances.


WerWolv:
Thank you!

PR #1275, opened 2023-08-30
Refactoring of floating point tool to decrease code complexity.

paxcut:
I noticed the bad score on code factor so I reorganized it to make it more readable and maintainable. In order to break down the big function into it much smaller parts I encapsulated all the variables that the functions need to access in two classes, one for the imgui related statics and the other for non-static variables. 

When writing the smaller functions I was noticed that there was room to simplify the existing algorithms by writing functions that could be called by parts that previously shared no code. I tested the changes the same way I tested the original and it seems to work the same way but maybe a bit faster. Although it may be possible to further optimize the present code code factor no longer flags the function at all.


iTrooz:
Hey, if you rebase on master the CI should pass

paxcut:
Is something not passing? I haven't seen any messages about tests being done or succeeding/failing. Why do I need to rebase? Is this what you mean?

 git pull --rebase origin master

WerWolv:
Thank you!

PR #1276, opened 2023-08-30
feat: Remember bytes per row in Hex Editor

LennardF1989:
Before this setting was moved out of the global settings window, it would persist. This PR restores that, even though it's now part of the Hex Editor itself.


iTrooz:
This PR would introduce a setting that's not located in the settings window. I'm not against it, but I would prefer @WerWolv to confirm that this is okay before merging

WerWolv:
I'm generally fine with this. I used to do this before for other things

Issue #1281, opened 2023-09-02
[fix] Removal of e-Tugra root certificate


nagaripratap:
**Description**
Certifi 2023.07.22 removes root certificates from "e-Tugra" from the root store. These are in the process of being removed from Mozilla's trust store.

e-Tugra's root certificates are being removed pursuant to an investigation prompted by reporting of security issues in their systems. Conclusions of Mozilla's investigation can be found here.

**Information**

Manifest Path : [main/romfs/cacert.pem](https://github.com/WerWolv/ImHex/blob/f725d763d1443ec7f6d85e492e4c038d12a70ff8/main/romfs/cacert.pem#L1388)

iTrooz:
Thanks for the report !

So apparently, that file was part of an old bug fix (#680), that since got scrapped, that file is a remnant of that.

Whether or not that bugfix should have been scrapped is another question, but in any case we probably won't come back to bundling the cacert file inside this repository, so I'm removing it

Thanks !

nagaripratap:
Great, appreciate your swift response

Issue #680, opened 2022-08-16
[Bug] AppImage version fails to load remote data


joshas:
### Operating System

Linux

### What's the issue you encountered?

Starting program in terminal shows following error messages:
````
[ERROR] [libimhex] Net request failed with error 77: 'Problem with the SSL CA cert (path? access rights?)'
[ERROR] [main] Init task 'Checking for updates...' threw an exception: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal
[ERROR] [libimhex] Net request failed with error 77: 'Problem with the SSL CA cert (path? access rights?)'
[ERROR] [libimhex] Net request failed with error 77: 'Problem with the SSL CA cert (path? access rights?)'
````
Update checking doesn't work. Clicking links in "Learn" and "Help" sections does nothing, results only in error message in terminal: `*** stack smashing detected ***: terminated`

### How can the issue be reproduced?

Download latest AppImage version, set program as executable and run program.

### ImHex Version

1.21.1

### ImHex Build Type

- [ ] Nightly or built from sources

### Additional context?

Fedora Linux 35 (Workstation Edition) 64-bit

iTrooz:
Note : I could reproduce this on Fedora 35 but not Fedora 36
I will probably make more testing later

iTrooz:
Update : this is caused because the path to the CA bundle is not a standard location. *sigh*
https://gitlab.com/probono/platformissues#certificates
https://curl.se/mail/lib-2022-05/0039.html

I'm going to investigate more, but it seems we have two ways to fix it :
- Patch gnutls to check all the paths : https://github.com/darealshinji/vlc-AppImage/issues/1#issuecomment-321041496
- Check the paths ourselves in the application (ImHex) : https://curl.se/mail/lib-2022-05/0039.html

I'm not sure what direction to take, I've asked on the #AppImage IRC channel if there was a recommended way to handle this

iTrooz:
Well, we've gone another path
The AppImage now use a bundled version of the CA certificates

PR #1282, opened 2023-09-02
fix: Remove remannts of #680 bugfix

iTrooz:
https://github.com/WerWolv/ImHex/commit/0a6815da8f470e19e8dfa9205b7c8a51b531e6a5 removed the bugfix of #680 by accident (but it seems like this is okay because the situation seems to have evolved and now curl seems to find the system certificate bundle, in different distros, in different locations)

this PRs removes the remnant of this bugfix


iTrooz:
Fix #1281

PR #1302, opened 2023-09-08
Add support for the hex::group attribute and various fixes

LennardF1989:
As discussed (many times) on Discord, does the same as the new favorite tag, but instead allows you to add multiple groups.

Initially, this would cause some insane issues with draw/reset (apparantly) fighting eachother in the pattern drawer. After a lot of trial and error, I decided to rewrite the flow that is responsible for calling reset. Now evaluating patterns is the one to decide when the reset happens, not the core "game"-loop.

To make sure that draw and reset can never happen at the same time, the mutex originally used for the favorites has been repurposed. Due to the restructuring, the mutex in the favorite-task is no longer needed, as that will only ever kick-off after reset is called and if there are actually patterns, which can never line up to be accessed on different threads at the same time.

Last but not least, I noticed that hard crashes could result in your config file getting overridden. I added a check to prevent that.

Last I issue I can see is that if you use an excessive amount of favorites/groups, a crash can still happen, but it only happens when you close the program (occasionally, but unpredictable). Before, this would happen if you ran the evaluation a second time. I boiled the cause of the crash down to these lines of code in evaluator.cpp > patternDestroyed:

```cpp
if (pattern->isPatternLocal()) {
    if (auto it = this->m_patternLocalStorage.find(pattern->getHeapAddress()); it != this->m_patternLocalStorage.end()) {
        auto &[key, data] = *it;

        data.referenceCount--;
        if (data.referenceCount == 0)
            this->m_patternLocalStorage.erase(it);
    } else if (!this->m_evaluated) {
        err::E0001.throwError(fmt::format("Double free of variable named '{}'.", pattern->getVariableName()));
    }
}
```

Specifically, trying to access the `*it` is the reason for the crash (this was also the cause of the crashes before my fixes, but then during evaluation). 

I'm suspecting the root cause is somewhere in the `.clone` methods of the patterns. I'd say that for now a crash when closing the program is more acceptable than during evaluation (which can even happen if you use favorites).

### DIFF
--- a/lib/libimhex/include/hex/api/event.hpp
+++ b/lib/libimhex/include/hex/api/event.hpp
@@ -206,6 +206,7 @@ namespace hex {
     EVENT_DEF(EventWindowInitialized);
     EVENT_DEF(EventBookmarkCreated, ImHexApi::Bookmarks::Entry&);
     EVENT_DEF(EventPatchCreated, u64, u8, u8);
+    EVENT_DEF(EventPatternEvaluating);
     EVENT_DEF(EventPatternExecuted, const std::string&);
     EVENT_DEF(EventPatternEditorChanged, const std::string&);
     EVENT_DEF(EventStoreContentDownloaded, const std::fs::path&);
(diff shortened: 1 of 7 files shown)


WerWolv:
Thank you!

PR #1309, opened 2023-09-14
fix `unsubscribe` bug

Chaojimengnan:
### Problem description

Fixed possible bug of  `EventManager::unsubscribe`

`std::map` only allows unique key, but the same token can subscribe to multiple events.
https://github.com/WerWolv/ImHex/blob/1a2a926b772f32d4f5e4c723c48e0e6e65a16b0a/lib/libimhex/include/hex/api/event.hpp#L104-L107

If the previous token has already subscribed to an event, then when subscribing again, `getTokenStore().insert` will not do anything (Because its type is `std::map`)

https://github.com/WerWolv/ImHex/blob/1a2a926b772f32d4f5e4c723c48e0e6e65a16b0a/lib/libimhex/include/hex/api/event.hpp#L122-L134

At this point in `unsubscribe`, the `iter` may not be able to find the correct event and erase it

Change `tokenStore` to `std::multimap` instead of `std::map`,  which cannot unsubscribe multiple events correctly

### DIFF
--- a/lib/libimhex/include/hex/api/event.hpp
+++ b/lib/libimhex/include/hex/api/event.hpp
@@ -161,7 +161,7 @@ namespace hex {
         }
 
     private:
-        static std::map<void *, EventList::iterator>& getTokenStore();
+        static std::multimap<void *, EventList::iterator>& getTokenStore();
         static EventList& getEvents();
     };
 
--- a/lib/libimhex/include/hex/api/event.hpp
+++ b/lib/libimhex/include/hex/api/event.hpp
@@ -103,6 +103,17 @@ namespace hex {
          */
         template<impl::EventType E>
         static void subscribe(void *token, typename E::Callback function) {
+            if (getTokenStore().contains(token)) {
+                auto&& [begin, end] = getTokenStore().equal_range(token);
+                auto eventRegistered = std::any_of(begin, end, [&](auto &item) {
+                    return item.second->first == E::Id;
+                });
+                if (eventRegistered) {
+                    log::fatal("The token '{}' has already registered the same event ('{}')", token, wolv::type::getTypeName<E>());
+                    return;
+                }
+            }
+
             getTokenStore().insert(std::make_pair(token, subscribe<E>(function)));
         }
 
--- a/lib/libimhex/source/api/event.cpp
+++ b/lib/libimhex/source/api/event.cpp
@@ -2,8 +2,8 @@
 
 namespace hex {
 
-    std::map<void *, EventManager::EventList::iterator>& EventManager::getTokenStore() {
-        static std::map<void *, EventManager::EventList::iterator> tokenStore;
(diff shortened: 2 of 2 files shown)


Nowilltolife:
Are there any regressions after this fix? Or were there any occurrences of this before?

Chaojimengnan:
> Are there any regressions after this fix?

No.

> Or were there any occurrences of this before?

It may not have happened before, but as I mentioned above, this bug will cause your event handler to still be called after `unsubscribe`

WerWolv:
Thanks for the PR!
You're correct, this is an issue however I'd argue that the same event should never be registered twice using the same key. Doing so would be a programming error to begin with.

So instead of switching to a multimap, I suggest simply having a check that displays a fatal error message and returns if an event has already been registered with this key.

Chaojimengnan:
Hi, 
>  I'd argue that the same event should never be registered twice using the same key.

I agree.

> So instead of switching to a multimap, I suggest simply having a check that displays a fatal error message and returns if an event has already been registered with this key.

The key issue is that `map` can only store one event iterator for one key. If this key needs to be registered to multiple different events, there will be problems, such as the following code, which cannot properly unsubscribe:

https://github.com/WerWolv/ImHex/blob/1a2a926b772f32d4f5e4c723c48e0e6e65a16b0a/main/source/window/window.cpp#L96-L105

So we still need to switching `tokenStore` to `multimap`  to ensure that we can store iterators for multiple different events of a token

To avoid registering the same event multiple times with same key as you mentioned, it is also necessary to add corresponding checks.

paxcut:
There will never be any case when the same key is used to register different events. If such a situation arises for some unexpected reason then the problem must be dealt with it at that precise moment. When it is time to register a key,  the map is checked to see if the key is being used already. If it is not then the registration of the event proceeds as usual, but if it is then a fatal error message is printed and the function returns without registering any new events as was explained in the comment above. 
 
 Another way to handle the problem could be to create a completely different  new key when the one used as argument already exists and notify the caller about the key change. To switch `tokenStore` to use `multimap` means that there will be cases when the same key is used to register different events but the checks guarantee that this never actually happens.

paxcut:
I can confirm that subscribe is called using the same key and different values and after the first one any additional subscribe requests are simply ignored. I can't determine if this is intended behavior or a bug.  
On the one hand it seems that event subscription is an important component of the inner workings of ImHex, 
On the other hand nothing appears to be broken where the error could originate from incorrect or missing event subscriptions. Perhaps what is needed here is a rigorous suite of tests that can detect event subscription errors such as omissions, mismatches,...

Chaojimengnan:
Event subscription mainly relies on the non-token overload of `subscribe`, and `tokenStore` is mainly used on `unsubscribe`. So even if `tokenStore` is `map`, event subscription can still work correctly, but it cannot unsubscribe correctly.

Currently, it seems that `unsubscribe` is only called when the program ends and cleans up, so there are no issues at this time.

WerWolv:
Thanks a lot!

PR #1331, opened 2023-09-30
Linux memory provider

trumank:
### Problem description

Implement a Linux backend for the ProcessMemoryProvider plugin.

Most of the provider code is the same between Windows and Linux. The primary differences are:
- enumerate PIDs in `/proc/` to get the process list
- use `/proc/<PID>/cmdline` as the process name
- parse `/proc/<PID>/maps` to get the module list
- reading/writing from memory is done using `process_vm_readv`/`process_vm_writev`

NOTE: `sudo setcap CAP_SYS_PTRACE=+eip build/imhex` must be run to give the binary permission to read another process' memory. Running as root user should also work but I would not recommend it.

### Additional things
The existing translations keys no longer match since I moved the plugin from `windows` to `builtin`.

I'm not well versed in C++ so I attempted to keep my changes rather simple. Feedback is very welcome.


plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -13,28 +21,70 @@
...
+        };
+        const iovec remote = {
+            .iov_base = (void*) address,
+            .iov_len = size,
+        };
+
+        auto read = process_vm_readv(this->m_processId, &local, 1, &remote, 1, 0);
+
+        if (read == -1) {
+            // TODO error handling strerror(errno)
```
Nowilltolife, line 65: Can you expand this TODO before submiting the PR?

plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -125,18 +176,42 @@ namespace hex::plugin::windows {
...
+                try {
+                    processId = std::stoi(path.filename());
+                } catch (...) {
+                    continue; // not a PID
+                }
+
+                std::ifstream file(path / "cmdline");
+                if (!file) continue;
+                std::stringstream buffer;
+                buffer << file.rdbuf();
```
Nowilltolife, line 194: The use of a string stream for string conversion seems rather expensive, is there no other way?

plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -265,13 +349,51 @@ namespace hex::plugin::windows {
...
+
+        std::string line;
+        while (std::getline(file, line)) {
+            std::istringstream iss(line);
+
+            std::vector<std::string> tokens;
+
+            std::string token;
+            while (tokens.size() < 5 && iss >> token)
+                tokens.push_back(token);
```
Nowilltolife, line 374: prefer using a vector construction using iterators like this:
```c++
        std::vector<std::string> tokens{std::istream_iterator<std::string>{iss},
                                       std::istream_iterator<std::string>{}};
```

plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -265,13 +349,51 @@ namespace hex::plugin::windows {
...
+                continue;
+            name = wolv::util::trim(name);
+
+            std::istringstream rangeStream(tokens[0]);
+            std::string startStr, endStr;
+
+            std::getline(std::getline(rangeStream, startStr, '-'), endStr);
+
+            u64 start = std::stoull(startStr, nullptr, 16);
+            u64 end = std::stoull(endStr, nullptr, 16);
```
Nowilltolife, line 390: this can be compacted using stream operators:
```c++
u64 start, end;
char hyphen;

rangeStream >> std::hex >> start >> hyphen >> end;
```

plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -13,28 +21,70 @@
...
+        };
+        const iovec remote = {
+            .iov_base = (void*) address,
+            .iov_len = size,
+        };
+
+        auto read = process_vm_writev(this->m_processId, &local, 1, &remote, 1, 0);
+
+        if (read == -1) {
+            // TODO error handling strerror(errno)
```
Nowilltolife, line 85: "

Nowilltolife (changes requested):

plugins/builtin/source/content/providers/process_memory_provider.cpp
trumank: A cursory search didn't turn up anything very concise. Is there a better way to read a file to a string?

plugins/builtin/source/content/providers/process_memory_provider.cpp
trumank: Ah right, I neglected to note my concern in the PR description. `process_vm_readv` will fail without the `CAP_SYS_PTRACE` capability so would make sense to propagate this error to the user so permissions can be fixed. How are errors like these best handled?

plugins/builtin/include/content/providers/process_memory_provider.hpp
```diff
@@ -3,7 +3,11 @@
 #include <hex/providers/provider.hpp>
 #include <hex/api/localization.hpp>
 
-#include <windows.h>
+#ifdef _WIN32
```
WerWolv, line 6: We generally use `#if defined()` instead of `#ifdef` and we also have defines available so you don't need to rely on the system ones. `OS_WINDOWS` and `OS_LINUX` is what other code uses mainly

plugins/builtin/source/content/providers/process_memory_provider.cpp
WerWolv: It's difficult right now to show errors like that from the read/write functions. You can only properly return errors from the open function

plugins/builtin/source/content/providers/process_memory_provider.cpp
WerWolv: Definitely don't use `std::ifstream` but our own `wolv::io::File` class which is a wrapper around the C File API.
This also lets you just read the entire content of the file to a std::string without having to rely on any streams

plugins/builtin/include/content/providers/process_memory_provider.hpp
```diff
@@ -38,11 +48,11 @@ namespace hex::plugin::windows {
 
         void save() override {}
 
-        [[nodiscard]] std::string getName() const override { return hex::format("hex.windows.provider.process_memory.name"_lang, this->m_selectedProcess != nullptr ? this->m_selectedProcess->name : ""); }
+        [[nodiscard]] std::string getName() const override { return hex::format("hex.builtin.provider.process_memory.name"_lang, this->m_selectedProcess != nullptr ? this->m_selectedProcess->name : ""); }
         [[nodiscard]] std::vector<Description> getDataDescription() const override {
             return {
-                    { "hex.windows.provider.process_memory.process_name"_lang, this->m_selectedProcess->name },
-                    { "hex.windows.provider.process_memory.process_id"_lang, std::to_string(this->m_selectedProcess->id) }
+                    { "hex.builtin.provider.process_memory.process_name"_lang, this->m_selectedProcess->name },
```
WerWolv, line 54: Since you renamed those strings here, don't forget to also adjust them in all the localization files in the romfs

plugins/builtin/source/content/providers/process_memory_provider.cpp
```diff
@@ -265,13 +349,43 @@ namespace hex::plugin::windows {
...
+            if (memoryInfo.State & MEM_RESERVE) name += hex::format("{} ", "hex.builtin.provider.process_memory.region.reserve"_lang);
+            if (memoryInfo.State & MEM_PRIVATE) name += hex::format("{} ", "hex.builtin.provider.process_memory.region.private"_lang);
+            if (memoryInfo.State & MEM_MAPPED)  name += hex::format("{} ", "hex.builtin.provider.process_memory.region.mapped"_lang);
 
             this->m_memoryRegions.insert({ { (u64)memoryInfo.BaseAddress, (u64)memoryInfo.BaseAddress + memoryInfo.RegionSize }, name });
         }
+
+#elif __linux__
+
+        std::ifstream file(std::fs::path("/proc") / std::to_string(this->m_processId) / "maps");
```
WerWolv, line 362: This should also be replaced with the API mentioned above

WerWolv (changes requested):

WerWolv:
Amazing work, thanks a lot!

xtexChooser:
ptrace should also be enabled for flatpak, I think. [like this](https://github.com/xtexChooser/ImHex/commit/393f28af658c09ca4b39041dcd07129fbc707450)
```patch
From 393f28af658c09ca4b39041dcd07129fbc707450 Mon Sep 17 00:00:00 2001
From: xtex <xtexchooser@duck.com>
Date: Sun, 5 Nov 2023 08:51:58 +0800
Subject: [PATCH] Allow ptrace in Flatpak sandbox

Signed-off-by: xtex <xtexchooser@duck.com>
---
 dist/net.werwolv.ImHex.yaml | 1 +
 1 file changed, 1 insertion(+)

diff --git a/dist/net.werwolv.ImHex.yaml b/dist/net.werwolv.ImHex.yaml
index 28f0118..022289d 100644
--- a/dist/net.werwolv.ImHex.yaml
+++ b/dist/net.werwolv.ImHex.yaml
@@ -10,6 +10,7 @@ finish-args:
   - --socket=x11
   - --filesystem=host
   - --device=all
+  - --allow=devel
 
 modules:
   - name: libiconv
-- 
2.42.0

```

iTrooz:
@trumank Hey, do you plan to continue working on this ? If not, I'd like to pick it up at some point in the future

trumank:
> @trumank Hey, do you plan to continue working on this ? If not, I'd like to pick it up at some point in the future

Hey! I may at some point but if you want to pick it up before that happens, feel free.

One thing I should note is that `/proc` files report zero size (https://stackoverflow.com/a/58920635) so `wolv::io::File` won't work as-is as it attempts a stat to determine how large a buffer to read.

WerWolv:
Using `wolv::io::File` should work in general, you just need to tell it the size to read manually using `file.readVector(1000)`. That will try to read 1000 bytes from the file instead of trying to query the size from the file. In this case here, I'd just try to read maybe 1MiB from that file. If it's less than that, the function will just return a smaller vector containing the data.

WerWolv (approved):

PR #1341, opened 2023-10-04
MacOS: Don't force using discrete graphics card

anka-213:
### Problem description

When starting ImHex on a MacBook model with both integrated and discrete graphics, it will force the computer to use the discrete graphics card. This causes increased power usage, meaning the fans will spin up, the battery will drain faster, etc. This program is not very graphics intensive, so using the discrete graphics card shouldn't be needed.

I changed the [`GLFW_COCOA_GRAPHICS_SWITCHING`](https://www.glfw.org/docs/latest/window_guide.html#window_hints_osx) setting in GLFW to not enforce using the discrete graphics.

### Additional things

My editor is configured to automatically remove trailing whitespace, so I hope that those whitespace changes are ok

### DIFF
--- a/main/gui/source/init/splash_window.cpp
+++ b/main/gui/source/init/splash_window.cpp
@@ -320,6 +320,7 @@ namespace hex::init {
             glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
             glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
             glfwWindowHint(GLFW_COCOA_RETINA_FRAMEBUFFER, GLFW_FALSE);
+            glfwWindowHint(GLFW_COCOA_GRAPHICS_SWITCHING, GLFW_TRUE);
         #else
             glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
             glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 0);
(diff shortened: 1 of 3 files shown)


WerWolv:
Thanks a lot for the PR!
I don't have much experience with macs myself so I'm asking you. Does this change make things worse for "lower end" models? It automatically using the dedicated GPU is probably not great but if it makes the app unusable for people with low end macbooks, we might need to detect if you're running on a higher end model and just enable the setting then

anka-213:
@WerWolv As far as I can tell, this setting makes no difference whatsoever on the lower end models. All it does is allow the OS to select the integrated graphics as it wants to, instead of enforcing the use of the more powerful discrete graphics.

----

Here's the relevant snippet from the documentation I linked above. Emphasis mine:

> **GLFW_COCOA_GRAPHICS_SWITCHING** specifies whether to in Automatic Graphics Switching, i.e. to allow the system to choose the integrated GPU for the OpenGL context and move it between GPUs if necessary or whether to force it to always run on the discrete GPU. This *only affects systems with both integrated and discrete GPUs*. Possible values are GLFW_TRUE and GLFW_FALSE. *This is ignored on other platforms*.

So it would probably even be fine to enable it on non-macos systems, it would just be ignored.

> Simpler programs and tools may want to enable this to save power, while games and other applications performing advanced rendering will want to leave it disabled.

I believe that this would count as a simpler program without advanced rendering?

> A bundled application that wishes to participate in Automatic Graphics Switching should also declare this in its Info.plist by setting the NSSupportsAutomaticGraphicsSwitching key to true.

I only noticed this last paragraph now. I haven't looked up how to do that yet.

anka-213:
I've added the config setting mentioned in the last paragraph as well now. It seemed to work fine without it, but since the documentation said that it should be added, it's probably best to do so.

WerWolv:
ImHex can still get intensive for the GPU when there's a lot going on on the screen. For example lots of bytes being drawn and highlighted.
The change is reasonable though, especially if it always just fully spinns up the external GPU otherwise. I'll wait for the checks to finish and merge then, thanks a lot!

PR #1371, opened 2023-10-18
fix: Updated wrong link to pattern language docs

gzalo:
Just a very simple update to point to the right documentation URI


WerWolv:
Thank you!

PR #1387, opened 2023-10-23
Add delay for autoeval while editing

sisco0:

### DIFF
--- a/plugins/builtin/include/content/views/view_pattern_editor.hpp
+++ b/plugins/builtin/include/content/views/view_pattern_editor.hpp
@@ -179,6 +179,8 @@ namespace hex::plugin::builtin {
         std::atomic<bool> m_resetDebuggerVariables;
         int m_debuggerScopeIndex = 0;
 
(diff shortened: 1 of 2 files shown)


WerWolv:
Hmm I'm not sure adding a delay is the right way to go. This seems to be a race condition so a delay just makes the issue not appear as frequently anymore but doesn't actually fix the underlying bug.
Thanks a lot for the PR though, I'll look into it some more later

MattMills:
Would this make more sense as a config option that defaults to no delay? Personally I'd love to be able to have an execution delay after typing is finished just when editing patterns that take a 5-10 seconds to run, it can kind of annoying when it repeatedly executes with a large pause while still typing when I'm trying to edit the pattern and looking at the data the pattern is generating.

Tangentially, it'd be nice for the pattern to remain in place until the execution completes successfully, often I find I have to undo a bunch of changes mid-edit to get the pattern to run again to figure out what data I'm getting wrong within the data.

PR #1406, opened 2023-11-07
fix: Calculate default CRC-32

qux-bbb:
### Problem description

Before this PR, the CRC-32 value of ImHex is not like most of others.  

Just set m_reflectIn and m_reflectOut of CRC-32 true by default.


WerWolv:
Thank you!

PR #1407, opened 2023-11-08
Remove Anonfiles code

iTrooz:
@WerWolv Waiting for your approval since you commented it


WerWolv:
I commented it out because I wasn't sure if maybe anonfiles will come back in the future or if another similar service will pop up. I thought about maybe replacing it with https://catbox.moe/ or https://litterbox.catbox.moe/ but not sure yet

iTrooz:
Okay ! Closing, then

PR #1414, opened 2023-11-11
Make Imhex buildable for MacOS arm64

iTrooz:
This PR helps with #710 **but does not fix it**

This PR only fix minor things in ImHex, and adds build Dockerfiles to allow you to obtain a **.app folder with MacOS arm64 code** (currently packaged in an ISO file)

**What still needs to be done** (in another PR):
- **Package the .app in a dmg** (apparently the current packaging doesn't work on some setups (e.g. @StArBoY-Works)

- **Sign it**
	Context: MacOS wants all arm64 code to be signed (but it does not need to be a key associated with a Developer ID AFAIK). Right now the binaries seems to be signed, but the signature are invalidated by the `install_name_tool` command invoked after. This command seems to be invoked from https://github.com/WerWolv/ImHex/blob/6e6c5c4cb73676f948e5bc94c89fb45c2cbc8d68/cmake/build_helpers.cmake#L243 and https://github.com/WerWolv/ImHex/blob/5171bea0bf64befb000fc1a0a354e34a6004f891/cmake/modules/PostprocessBundle.cmake
	Example warning message: `/usr/bin/install_name_tool: warning: changes being made to the file will invalidate the code signature in: /mnt/ImHex/build/imhex.app/Contents/MacOS/plugins/script_loader.hexplug`


iTrooz:
Links that may be useful for that next PR:
https://github.com/mozilla/libdmg-hfsplus 
https://github.com/indygreg/apple-platform-rs
https://github.com/indygreg/apple-platform-rs/issues/2
https://github.com/multiarch/crossbuild/issues/51#issuecomment-1225058473
https://github.com/tpoechtrager/osxcross/issues/355

PR #1417, opened 2023-11-11
build: move third party libraries to lib/third_party

iTrooz:


WerWolv:
Should we maybe keep imgui in the external folder? Since there's a bunch of custom stuff in there

iTrooz:
We do, but isn't most of this stuff from different third party repositories ? (e.g. imgui, implot, imnodes, imstb..) Is there anything in there except CMakeLists.txt that we have created ?

WerWolv:
The fonts are custom, the glfw backend has a bunch of changes, the TextEditor has a ton of changes and the imgui config is in there

iTrooz:
To me it still seems like most of the code is third party, and we just made changes to it, but if you really want to move it to lib/external, I don't have anything against it

WerWolv:
Hm. No I think it's okay. Let's leave it the way you do it

Issue #1109, opened 2023-05-26
[Feature] macOS native arm64 builds


eirnym:
### What feature would you like to see?

Could you please add macOS M1/M2 native builds?

### How will this feature be useful to you and others?

No need to install Rosetta 2 or every time build manually

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

eirnym:
Xcode can build universal apps even on x86_64 and since Xcode 12 it's a default behaviour

WerWolv:
ImHex is built using GCC on macOS, we don't really use Apple's compiler or XCode at all because it's so outdated.

Of course we could still cross compile but figuring out how to do it without having access to any Mac is basically impossible. 
I'm very happy to merge pull requests adding support for this though. I just really don't want to do any more macOS debugging through GitHub Actions myself ever again lol

eirnym:
Xcode uses clang and it's quite fresh. Older Xcode versions use older clang, obviously. 

I personally prefer clang over gcc

eirnym:
BTW, cmake can generate Xcode project.

eirnym:
Nevertheless, could you please keep this feature request open. I guess somebody will find a solution faster if it'd be open.

WerWolv:
I created a new pinned issue now with this request amongst other things that I really need help with. #1111

eirnym:
About clang/debug issue, could you explain it further? In way it's described it's very unclear, what is the problem and how to overcome it if somebody would want to look into.

WerWolv:
I don't know how to better explain it to be honest.
We were using clang at the start which sometimes caused the build to break when I made changes. Mostly because either features weren't available yet, standard library features were implemented wrongly or differently than on other platforms, etc.
When these things happened, I always had to make a new branch, make small changes to my code, push them to github to build, wait like 20 minutes for it to fail and then continue until I found what was wrong. That was extremely frustrating. Using GCC on all platforms allows me to verify that things are correct on Windows or Linux and can simply expect them to work properly on macOS as well. 
Again, no issue if I had a mac but I don't so this is the best I can do realistically. ImHex is using very modern C++20 and C++23 features and I don't want to give up on that.

iTrooz:
Hey, we've decided to use this issue as the "official" one for discussion about this problem.
So essentially we're going to close any other issue and point at this one

Please comment here if you have ideas to share !
(Please just comment just to say "plz do it")

[...]

PR #1425, opened 2023-11-14
add MacOS arm64 job

iTrooz:
I said I'll do it in #1414

Warning: the output zip does not pass SIP (System Integrity Protection), and I have no plan to do it for now. So if someone here has a Mac, feel free to to try to fix the problem (which is probably our build system)

Link #1109


iTrooz:
Note: this PR adds a dependency on a binary artifact in my repository https://github.com/iTrooz/macos-crosscompile (the docker image package)

iTrooz:
Also, note that the job in this PR is dependent on a package in one of my repository: ghcr.io/itrooz/macos-crosscompile:clang17-nosdk

WerWolv:
Is ghcr.io/itrooz/macos-crosscompile:clang17-nosdk mainly meant for compiling ImHex now? I wouldn't mind having it on my / this repos docker repository if you don't want to have it depend on your repo. I'm fine with how it is now too though

iTrooz:
> Is ghcr.io/itrooz/macos-crosscompile:clang17-nosdk mainly meant for compiling ImHex now

Right now yeah, but I tried to make it generic, and added a README so that other people can use it. And I plan to *relatively* maintain it for general use

I think it's better if it stays in my repository, so we don't maintain two versions of the same thing

I still kept an (unused) copy of the Dockerfile is this repository (that I don't plan to update), just in case

Issue #1380, opened 2023-10-22
[Bug] File with chinese name can not open


qux-bbb:
### Operating System

Windows

### What's the issue you encountered?

I have a file named "测试.exe_". If I open it with Context "Open with ImHex", ImHex will show:  
```
Failed to open provider: Failed to open file
D:\recent\tmp\??.exe_: No such file or directory
```

### How can the issue be reproduced?

Rename any file to a chinse name, then open it with Context "Open with ImHex".

### ImHex Version

1.31.0

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

iTrooz:
After investigating, I can reproduce this on Windows in release (1.31.0) (but not 1.30.1), but not on Linux

So I'm going to unassign for this and let someone who use Windows more pick it up

PR #1435, opened 2023-11-19
fix: convert command line arguments to UTF-8 for windows builds

kyle-sylvestre:
### Problem description
command line characters outside of the system's ANSI codepage are converted to '?'. Issue is seen in [Issue 1380](https://github.com/WerWolv/ImHex/issues/1380). This is a regression from 1.30.1 in [this commit](https://github.com/WerWolv/ImHex/commit/1ed658bcdc33207fe1da2c09d60c7769f9bf5f28), where the following got removed:
```
        std::optional<std::u8string> getProgramArgument(int index) {
            if (index >= impl::s_programArguments.argc) {
                return std::nullopt;
            }

            #if defined(OS_WINDOWS)
                std::wstring wideArg = ::CommandLineToArgvW(::GetCommandLineW(), &impl::s_programArguments.argc)[index];
                std::string byteArg = std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>, wchar_t>().to_bytes(wideArg);

                return std::u8string(byteArg.begin(), byteArg.end());
            #else
                return std::u8string(reinterpret_cast<const char8_t *>(impl::s_programArguments.argv[index]));
            #endif
        }
```

replace main's argv with UTF-8 equivalent

### Screenshots
N/A

### Additional things
N/A

### DIFF
--- a/main/gui/source/main.cpp
+++ b/main/gui/source/main.cpp
@@ -227,13 +227,14 @@ int main(int argc, char **argv) {
 
 #if defined(OS_WINDOWS)
     // create UTF-8 argv
+    int wargc = 0;
     std::string argbuf;
     std::vector<char *> utf8_args;
-    WCHAR **wargv = ::CommandLineToArgvW(::GetCommandLineW(), &argc);
+    WCHAR **wargv = ::CommandLineToArgvW(::GetCommandLineW(), &wargc);
     if (wargv) {
         // convert WCHAR arguments to UTF-8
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot! I pulled my hair out over this issue before because for whatever reason it worked in debug mode.
I reworked and simplified your code a bit to better fit into the code where it's being used, thanks a lot for the PR though! All credit goes to you

PR #1436, opened 2023-11-19
fix: gl.h for MacOS

VeeBop:
### Problem description
Commit  1249eb326112a8caa518151f0d82fbb7619938c0 added `#include <GL/gl.h>` in a file, but MacOS uses `<OpenGL/gl.h>`.

If compiling on MacOS, use `<OpenGL/gl.h>`. Otherwise, use `<GL/gl.h>`.

### Additional things
There is a redefine, so I used the preprocessor to ignore it for MacOS. This feels a little awkward to me (but I don't know how to do it better).

I'm not particularly familiar with C++, so let me know if this is the right solution. This is also my first pull request, so feel free to roast me for anything else :)


WerWolv:
Thank you!

PR #1442, opened 2023-11-22
fix: bad entry in settings ends in .

paxcut:
The entry is "hex.builtin.setting.interface.window." but I think it should be "hex.builtin.setting.interface.window.x"


WerWolv:
Thanks!

PR #1465, opened 2023-12-16
Feat: added find-replace to pattern editor 

paxcut:
Modeled after VS code find-replace ui except some features which are hard to implement like multi-cursors. It includes selection and scrolling to next/previous match,  bidirectional search and replace, case, whole word and regular expression match and regular expression replace, position and number of matches display, replace all, menu and keyboard shortcuts, integration of replacement  with undo/redo system and full history recall of previous searches/replacements. I may have omitted some minor features and others will be implemented at a later time.

### Problem description
The pattern editor has needed a decent search-replace since it was created. Although the color text editor fork changes are not useful as a cut paste , it still is able to show me how to get started implementing a complete feature set.

The UI was coded based on the existing newly created context menu and on the text editor side, the search and replace code was based on the code written for SHADERed, an app that uses a fork of the same text editor ImHex uses.

### Additional things

https://github.com/WerWolv/ImHex/assets/53811119/28a6cb6a-b325-495b-8f60-81e820112e5f


WerWolv:
Thanks a lot! It looks really good

I just tried it out and there seem to be a few issues still.
- The popup appears at the current cursor position, I think it would be better if it was attached to the top of the view similar to the Find popup in the hex editor
- The `no results found` popup is larger than the `1 of 15` text so it resizes the popup while typing which feels a bit strange
- The selection gets only set to the correct location on the first element. On all following ones, the selection is a few characters further ahead (this also causes replace to replace the wrong characters)
- Somehow both the first and the second item that's found is displayed as `1 of 15`

image
image

paxcut:
> The popup appears at the current cursor position, I think it would be better if it was attached to the top  of the view similar to the Find popup in the hex editor

I dont agree with it being fixed. I have an issue with the Find popup of the hex editor that if the input signal has only 16 bytes or so then the find popup blocks the data completely and since it cannot be moved you need to close it to know if search succeeded. 
I have gotten used to it popping at the mouse cursor and i move the mouse to a convenient place before opening the search. As long as the pattern editor has focus ctrl-f should always open it. or you can right click away from the text to open it via the menu.

> The no results found popup is larger than the 1 of 15 text so it resizes the popup while typing which feels a bit strange
yes, 'no results found' should be just 'no results' as in vscode. i forgot about that. thanks.

>The selection gets only set to the correct location on the first element. On all following ones, the selection is a few characters further ahead (this also causes replace to replace the wrong characters)

hmm, that seems to be related to you using regular expression to search. It doesn't happen in string search. I'll look into it.

>Somehow both the first and the second item that's found is displayed as 1 of 15

When you first type before you press enter the indicator will tell you which is the next match but it wont select it until you press enter, hit f3 or the down arrow button.  the count of matches should be correct so find which is the first one by searching backwards until it wraps and the search forward once. That is the first match in the file, the next one is 2 and so on.

I found a bug where the item selected is not replaced first and i fixed it locally. I'll work on the issues you raised.

paxcut:
> Somehow both the first and the second item that's found is displayed as 1 of 15
Scratch my comment about that. I just found that if regexp is turned on the count is short by one and the first two matches are labeled as position 1. Working on fixing that now.

As hard as I try I am not able to reproduce the misaligned selection problem or any indication of how it could be caused. Doesn't the bug fix itself if you select a word before opening the popup? (placing the cursor on a word should have the same effect)
I have a feeling regular expressions are behind that bug too. To confirm or deny that claim turn off regular expressions once the bug clicks in and see if it changes the selection at all. (non reggexp use string find which allows you to use an offset to start searching but regexp use regexep::find so it searches using an iterator and i am not using that correctly it seems)

paxcut:
I fixed all the problems listed but I ran into a troubling issue. To see the problem import the biggest pattern file (I think it is 3ds.hexpat) and open find dialog and type 'a' (no quotes). Depending on the speed of the computer it may take from less than 1 to a few seconds to process the 1000+ entries. That really can't be helped, but selecting successive matches shouldn't take as long to update as the first one did. Te problem is that it recalculates the entire match set at every step.

I have created a version that caches all the entries so as to avoid the problem. It is working well but I had to restructure the entire find/replace workings which of course broke just about everything else. I have repaired most of it but I wanted to give you an update to explain why some simple fixes are taking this long. I estimate I should be done sometime today or tomorrow. Thanks.

paxcut:
I am done with all changes to this pr for now. Please review at your earliest convenience and post any issues you may encounter. thanks.

WerWolv:
Fucked up and had to merge it manually into master now, sorry!

PR #1466, opened 2023-12-18
build: Add support for Ubuntu 23.10 (Mantic)

rasa:
### Problem description

No Ubuntu 23.10 (Mantic) build, which was released on 12-Oct-23. Build succeeded [here](https://github.com/rasa/ImHex/actions/runs/7251941095/job/19755373662).


farmerbb:
I noticed that both the Ubuntu 22.04 and 23.04 packages for v1.32.0 refuse to install on Ubuntu 23.10 due to the `libarchive` package being missing from the 23.10 repos.  Would be nice to get this PR merged if this resolves that issue.

WerWolv:
Thanks for the PR!
We're currently internally discussing whether it's worth it to have so many releases for every different distro (and especially multiple ones for different versions of distros). 
The consensus right now is that we already an AppImage and a Flatpak release that will work on every distro and the rest should probably be handled by packagers to bring ImHex to the package repository of that distro.

We're analyzing the downloads of these distro-specific releases right now to see what we'll do.

applecuckoo:
@WerWolv Ubuntu 23.10 hasn't been supported for a while, so this PR should probably be closed as stale.

PR #1486, opened 2024-01-06
Update plugin-bundle.cpp.in

pg83:
fix possible ODR - multiple classes with same name in different .cpp files


WerWolv:
Thanks a lot!

PR #1489, opened 2024-01-12
Update plugin_manager.cpp

pg83:
Fix imhex with static non-web non-emscripten build.

Or else I have errors about missing "Built-in" plugin on start.

Actually, I think that this method can be implemented as plain 

```
isBuiltInPlugin() {
    return getPluginName() == "Built-in";
}
```

but obviously I missing some point :)

### DIFF
--- a/lib/libimhex/source/api/plugin_manager.cpp
+++ b/lib/libimhex/source/api/plugin_manager.cpp
@@ -174,7 +174,7 @@ namespace hex {
         if (m_functions.isBuiltinPluginFunction != nullptr)
             return m_functions.isBuiltinPluginFunction();
         else
-            return false;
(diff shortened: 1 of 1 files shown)


pg83:
https://github.com/WerWolv/ImHex/actions/runs/7498260312/job/20413764903#step:5:1237

I think that it is not related with my patch

WerWolv:
Hey, thanks for the PR!
Seeing this, I decided to completely get rid of the concept of "built in plugins". Since we now have a bunch of different plugins that come bundled with ImHex, we don't really need it anymore. 
The linked commit fixes the issues when statically linking plugins and also revamps the built-in plugin situation resolving your problem entirely.

PR #1501, opened 2024-01-18
fix: segfault when starting with file arguments

yretenai:
d511080814dc78ad39a63f2071003c07ee37673c introduced a regression where running imhex with arguments (i.e. `imhex blob1.bin blob2.bin`) will segfault due to a null pointer dereference.

This patch updates `getSubCommands` to follow the same control flow as `getFeatures`, where if the function returns a null pointer, it will return gracefully rather than crash.


WerWolv:
Thank you!

PR #1523, opened 2024-01-31
feat: Add progress bar to web version

iTrooz:
This PR adds a progress bar to the web version

Credits to https://github.com/WordPress/wordpress-playground/pull/46 for the original idea, although it had to be modified

Link: https://github.com/emscripten-core/emscripten/issues/16278

### DIFF
--- a/dist/web/source/index.html
+++ b/dist/web/source/index.html
@@ -72,7 +72,8 @@
             <h2 id="not_working">
                 Not loading in your Browser? <a href="https://imhex.werwolv.net">Try the native version</a>
             </h2>
-            <div style="height: 50%"></div>
+            <h2 id="progress"></h2>
+            <div style="height: 40%"></div>
         </div>
 
         <div class="loading_ripple">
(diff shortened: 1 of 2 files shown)


ivancuric:
Hey! Just wanted to ask how is this different from simply hooking into [`Module.setStatus`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_push_uncounted_main_loop_blocker)?

I'm using the following snippet to convert the returned text into a numerical progress:
```ts
export function convertEmscriptenStatusToProgress(emStatus: string): number {
  // roughly based on https://github.com/emscripten-core/emscripten/blob/1.39.11/src/shell.html#L1259
  if (emStatus === "Running...") {
    // download has completed, wasm execution has started
    return 100;
  } else if (emStatus.length === 0) {
    // empty message
    return 0;
  }

  const regExp = RegExp(/([^(]+)\((\d+(\.\d+)?)\/(\d+)\)/);
  const match = regExp.exec(emStatus);
  if (match) {
    const currentValue = parseInt(match[2]);
    const maxValue = parseInt(match[4]);
    return (currentValue * 100) / maxValue;
  } else {
    // Cannot parse emscripten status
    return NaN;
  }
}
```

PR #1538, opened 2024-02-08
tests: Add infrastructure for testing plugins

iTrooz:
This PR adds a test architecture to be able to test plugins

Main infrastructure done by @WerWolv


lib/libimhex/source/test/tests.cpp
```diff
@@ -0,0 +1,23 @@
+#include <hex/test/tests.hpp>
+
+namespace hex::test {
+    std::map<std::string, Test> Tests::s_tests;
+
+    bool initPluginInt(std::string name) {
```
WerWolv, line 6: What does the `Int` mean here? The name is kinda weird.
If it's for "Internal", maybe use "Impl" instead

WerWolv (changes requested):

lib/libimhex/source/test/tests.cpp
iTrooz: Indeed, it was short for "Internal"
Done !

PR #1542, opened 2024-02-11
PoC: Support Copy/Paste on WASM build outside the application border

exsilium:
### Problem description
WASM build does not support copy/paste beyond the application. Meaning, there's no practical way of sending text back and forth across the application border.

There are lengthy threads why this is a technical challenge in WASM/Browser world, e.g:
- https://github.com/pthom/hello_imgui/issues/3
- https://github.com/emscripten-core/emscripten/pull/19510

Implements a workaround solution as Header only C++ library, as proposed and implemented at: https://github.com/Armchair-Software/emscripten-browser-clipboard

Maybe there are cleaner ways of achieving the functionality. Definitely would like to have some discussion around this. 👀 

ℹ️ The proposed PR "works for me" on Windows, using CTRL-C/V shortcuts to copy text from and to the application. On MacOS the system shortcut to Paste is different from what ImHex has defined. This results in system Paste shortcut of command-V triggering the browser callback to synchronise the application clipboard, but no actual Paste takes place within ImHex.

If there would be a clean way to trigger the paste command, that would be wonderful (or get the context and references to write the data to the cursor, but I was unable to find a clean solution). The only proposed solutions in the referenced threads were about triggering paste event internally via Key events. This seemed wonky 🙃 , so is not currently implemented. At the moment the paste on MacOS is command+V followed by control+V.

### Additional things
This is definitely a stopgap solution before the ImGui and Emscripten take a more proper solution in enabling Copy/Paste outside the application borders. However, I feel like this is a must have capability to make the WASM build more useful, not just for trying out ImHex.

Cheers! 🍻

### DIFF
--- a/lib/third_party/imgui/custom/include/emscripten_browser_clipboard.h
+++ b/lib/third_party/imgui/custom/include/emscripten_browser_clipboard.h
@@ -1,0 +1,99 @@
+#ifndef EMSCRIPTEN_BROWSER_CLIPBOARD_H_INCLUDED
+#define EMSCRIPTEN_BROWSER_CLIPBOARD_H_INCLUDED
+
+#include <string>
+#include <emscripten.h>
+
+#define _EM_JS_INLINE(ret, c_name, js_name, params, code)                          \
+  extern "C" {                                                                     \
+  ret c_name params EM_IMPORT(js_name);                                            \
+  EMSCRIPTEN_KEEPALIVE                                                             \
+  __attribute__((section("em_js"), aligned(1))) inline char __em_js__##js_name[] = \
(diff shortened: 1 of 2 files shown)


WerWolv:
Hey!

Thank you very much for the PR! This has been bothering me as well and we couldn't find a good way to do it.
Even if might just be a temporary solution, it's definitely much better than not having it.

exsilium:
Thank you for the positive feedback, @WerWolv , granted the pipeline goes through and there are no specific comments to this PR, I'm happy if this gets merged as is. 🙇

WerWolv:
All good, thanks a lot!

PR #1544, opened 2024-02-12
Fix compiling with Zstd (at least on *nix)

regginator:
Prior to this, at least on Linux/MacOS/etc, I guess it was not possible to compile with Zstd included for `hex::dec::zstd_decompress()`:

image

Every other target lib for the decompression plugin would compile, **except for** Zstd. Additionally, the target name `zstd` caused CMake to not be able to find `Find(ZSTD).cmake` due to case-sensitivity differences between Windows and Unix-based file paths, of course.
With that said, I'm not too sure if this will break building w/ Zstd on Windows... Hopefully not :smile:

### DIFF
--- a/plugins/decompress/CMakeLists.txt
+++ b/plugins/decompress/CMakeLists.txt
@@ -32,11 +32,11 @@ add_imhex_plugin(
         ZSTD
 )
 
-find_package(zstd)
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #1584, opened 2024-03-03
Fix build on FreeBSD

nmantani:
This pull request fixes build on FreeBSD. The changes are conditioned with `#if defined(__FreeBSD__)` preprocessor  macro and they should not affect build for other operating systems.


WerWolv:
Thank you!
Would it be possible to add a define like `OS_WINDOWS` or `OS_LINUX` from cmake maybe instead of using `__FreeBSD__` directly? Something like `OS_FREEBSD` would be nice so it can be used for places like this. I guess `OS_LINUX` should still be set as well though since the two have lots in common.

nmantani:
Thank you for the review. I will rewrite it with OS_FREEBSD.

nmantani:
Hi, I have rewritten it with OS_FREEBSD.

WerWolv:
Thanks a lot!

PR #1603, opened 2024-03-19
Fix editor behavior on macos

mrexodia:
cmd+up -> ctrl+home
cmd+down -> ctrl+end
cmd+left -> home
cmd+right -> end
cmd+option+left -> ctrl+left
cmd+option+right -> ctrl+right

### DIFF
--- a/cmake/modules/FindGLFW.cmake
+++ b/cmake/modules/FindGLFW.cmake
@@ -76,6 +76,7 @@ if(TARGET glfw)
         get_target_property(GLFW_DLL_RELEASE glfw IMPORTED_LOCATION_RELEASE)
     endif()
 
+    set(GLFW_LIBRARIES "GLFW::GLFW")
     return()
 endif()
 
--- a/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
+++ b/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
@@ -1647,6 +1647,7 @@ void TextEditor::TextEditor::MoveBottom(bool aSelect) {
     } else
         mInteractiveStart = mInteractiveEnd = newPos;
     SetSelection(mInteractiveStart, mInteractiveEnd);
+    MoveEnd(aSelect);
 }
 
 void TextEditor::MoveHome(bool aSelect) {
--- a/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
+++ b/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
@@ -623,6 +623,38 @@ void TextEditor::HandleKeyboardInputs() {
     auto shift  = io.KeyShift;
     auto ctrl   = io.ConfigMacOSXBehaviors ? io.KeySuper : io.KeyCtrl;
     auto alt    = io.ConfigMacOSXBehaviors ? io.KeyCtrl : io.KeyAlt;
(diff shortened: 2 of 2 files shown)


paxcut:
I submitted somewhat similar changes to an earlier [pr](https://github.com/WerWolv/ImHex/pull/1528) that seems to agree with your version in some cases. The thing is that I don't have access to a Mac so I couldn't test them properly. The changes were based on this list created by WerWolv, the author of ImHex, from his mac text editor
```
Command + Left/Right is Home/End.
Command + Up/Down is go to start/end of file
Option + Left/Right is Word Jump Left/Right
Option + Up/Down is Home/End when the cursor is not at the start / end of the line and then just cursor up and down if it is
```
That seems to agree with your settings of
```
cmd+up -> ctrl+home
cmd+down -> ctrl+end
cmd+left -> home
cmd+right -> end
```
but the last two seem to require three key presses so it would be different:
```
cmd+option+left -> ctrl+left
cmd+option+right -> ctrl+right
```
what you mean by ctrl+key, the windows key equivalents? 
There is also a change that would affect all keyboards, not just the mac, where you removed the check to see if cursor is already at the desired position. Can you explain briefly why it is necessary and how will it change the current behavior? From your commits you describe it as to trigger selection always, but wouldnt that be an empty selection if the key is already home or end?

mrexodia:
These were the fixes I came up with while discussing it with WerWolv on Discord a few days back. I don't think the Cmd+Option+Left is accurate to the logic in the code, but if you think everything is fixed I don't want to spend more time on this.

paxcut:
I got involved in this because there were several complains about mac handling in editor and I've been working a lot on improving it, but 1) I am not or have been a Mac user and 2) I don't own a mac to test what the keys do. 
I got the cmd-options-left out of your description of the pr, that why I said "it seemed to require" because I wasn't sure. If the keys are option-left and option-right then that should match the codes that WerWolv posted a month or two ago. What I need is to make sure that Mac users feel like imhex works like other editors, but since I can't test it I rather user your code which has been tested properly.
Now I understand why you implemented those empty selections.  They are designed to deselect the previous selection if it exists. It is indeed necessary so it looks like that gets added too for all keyboards. The last bit about ctrl-end I am afraid I have completely missed. I'll look again at your code to find it because it also is needed. Thank you for taking the time to explain and I'll make sure your named is mentioned in the change descriptions.

mrexodia:
Ah my bad, I was tired when writing the description. I’ll give the latest master a try soon! No need for credit for the code though 😊

paxcut:
no worries. I do have a question though, since you set ctrl to be super, why do you set ctrl to true when you know super is true already?

mrexodia:
In my implementation I emulate the Windows keybindings when on macos. The assignment to `ctrl` isn't strictly necessary, but I think it makes the code a bit clearer to read to indicate that `Cmd + Up` becomes `Ctrl + Home`

paxcut:
I see. According to WerWolv:
''' 
Apparently Super is Command on macOS
And Alt is Option
'''
So in my code
```cpp
 auto top      = io.ConfigMacOSXBehaviors ? io.KeySuper && up : ctrl && ImGui::IsKeyPressed(ImGui::GetKeyIndex(ImGuiKey_Home));
 ```
 Also shows the same thing. The more I look at it the more it seems like we both implemented the same thing which is reassuring.

PR #1609, opened 2024-03-22
Fix 40 typos

RoboSchmied:
### Problem description
fix 40 typos

`endianess` => `endianness`


WerWolv:
HUH I did not think it was spelled that way but you're right. Thank you!

PR #1612, opened 2024-03-23
Add dependency download script for OpenSUSE Tumbleweed

Kihau:
### Testing
After installing the dependencies, I have managed to successfully compile the project as described in the building instructions:
https://github.com/WerWolv/ImHex/blob/master/dist/compiling/linux.md

I have tested the script on a fresh installations of the OpenSUSE:
- Tumbleweed Server
- Tumbleweed Generic Desktop

### Note
Dependencies added in the script are based off of scripts for other Linux distributions.  
Because of that, some download dependencies might not be necessary (or could be already included by other packages), so it is very much possible that some of them could be removed.

### DIFF
--- a/dist/get_deps_tumbleweed.sh
+++ b/dist/get_deps_tumbleweed.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env sh
+
+zypper install     \
+  cmake            \
+  ninja            \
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #1623, opened 2024-03-31
Refactor langtool

xtexChooser:
- Better argument parsing
- Allow processing all language folders at the same time
- Allow an optional reference language when translating
- Save translations on KeyboardInterrupt
- Fixes a ooold input issues by importing readline (https://github.com/kovidgoyal/kitty/issues/6560)
- Add untranslate mode to remove translations by a key regex


WerWolv:
Thanks a lot! Sorry for leaving this sit for so long

PR #1632, opened 2024-04-16
fix framerate limiter

daver32:
### Problem description
The framerate limiter doesn't work when inputs are being sent (eg mouse cursor moving over the window), because `glfwWaitEventsTimeout` returns early when it encounters an event. 

I made it sleep for the remaining time when that happens.


WerWolv:
Thanks a lot!

Issue #1645, opened 2024-04-30
[Bug] File provider pop-up dialog stacks with multiple external file changes


SparkyTD:
### Operating System

Windows

### What's the issue you encountered?

The built-in File provider detects external changes to an open file and prompts the user to reload the file with a pop-up dialog. The issue is that if the file undergoes multiple changes in a row before the user can respond to the initial pop-up, additional dialogs will stack up, requiring the user to confirm or dismiss each one individually. This can be very annoying if the file changes multiple times.

### How can the issue be reproduced?

1. Create a text file and write some text in it
2. Open the file in both ImHex and Notepad
3. In Notepad, make three different changes to the file, making sure to save after each change
4. In ImHex, the "File has been modified" dialog will appear three times in a row, and you'll have to press "Yes" or "No" on all three.

### ImHex Version

1.33.0 (master@ef99e9d)

### ImHex Build Type

- [x] Nightly or built from sources

### Installation type

MSI

### Additional context?

Here are two possible solutions:
1) Add an option to auto-reload when the window is focused, with no dialog (like HxD, I would personally prefer this one)
2) Consolidate multiple changes into a single dialog that reloads the latest change (like Notepad++)

PR #1654, opened 2024-05-07
Fix a bug that caused the file provider's monitor to trigger the change dialog multiple times in a row (#1645)

SparkyTD:
### Problem description
This PR aims to address #1645 that caused the built in file provider's change monitor to trigger the notification popup dialog multiple times in a row after multiple external file changes.

I added an additional boolean field `m_changeEventAcknowledgementPending` that tracks whether there are any pending or unacknowledged change notification dialogs to prevent further dialogs from being opened. The flag is only reset to its initial value once the user has acknowledged the first `PopupQuestion` dialog.

Since the file is reloaded only after the user clicks 'Yes', it is unnecessary to ensure that only the latest popup is acknowledged.


WerWolv:
Thank youè

PR #1658, opened 2024-05-08
Convert all tool Popups to floating windows

SparkyTD:
### Problem description
In previous versions of ImHex, all tool windows were implemented as static popups fixed in the upper left position of the hex view. This PR refactors all tool popups to use floating windows that can be dragged around by the user, or closed with a dedicated close button on the title bar. These popup also support a stylable transparency when the user is not hovering their mouse over the window.

I rewrote the logic in `ViewHexEditor::drawPopup()` to use a custom `ImGuiExt::BeginHoveringPopup` function for rendering the popup windows. This new function is an almost exact replica of the built-in `ImGui::BeginPopupModal`, except it does also displays the default window title bar with a close button.

A second custom function, `ImGuiExt::PopupTitleBarButton` was also added for rendering small icon-based buttons into the title bar of the parent popup window. This new function was used to implement an optional "Pinning" feature that individual popup implementations can specify. If a window is pinned, it won't close automatically when its main action is executed. For example, the "Select" button on the Select dialog will close the popup by default, unless the window is pinned.

### Screenshots
Popup dialogs before:
image

Popup dialogs after:

https://github.com/WerWolv/ImHex/assets/45818400/99d1a628-8ac1-40ac-9146-9062091bb0db

### Additional things
- When the user stops hovering their mouse over a popup window, it becomes semi-transparent, making it easier to see the content behind it
- This PR also introduces the `styles.imhex.popup-alpha` style, making the transparency effect configurable, including the ability to disable the effect completely by setting `popup-alpha` to `1.0`.
- Fixed a bug that caused some popup windows to ignore the Enter and the KeypadEnter keys. With this PR, all tool windows will execute their main action when the user presses either one of the two Enter keys, and will also close automatically unless the window is pinned.

### Possible changes and improvements
- Should the transparency effect be disabled if a window is pinned?
- Should the transparency factor be modifiable on the Settings/Interface page?
- A keyboard shortcut could be added for quickly pinning / unpinning the current window.
- Can the pin icon stay on the left, or should it be moved next to the close button, with a similar circular background?


lib/libimhex/source/ui/imgui_imhex_extensions.cpp
```diff
@@ -1216,6 +1216,89 @@ namespace ImGuiExt {
...
+        // Draw the label in the center
+        ImU32 text_col = GetColorU32(p_enabled || hovered ? ImGuiCol_Text : ImGuiCol_TextDisabled);
+        window->DrawList->AddText(bb.GetCenter() - ImVec2(g.FontSize * 0.45F, g.FontSize * 0.5F), text_col, label);
+
+        // Return the button press state
+        ImGui::PopClipRect();
+        return pressed;
+    }
+
+    void PopupTitleBarText(const char* text) {
```
SparkyTD, line 1284: This function adds the window title label for the custom title bars, and also takes the current cursor position into consideration. This is useful, because the pin icon isn't present on all windows, and this way the X-offset of the title text can be adjusted accordingly.

lib/libimhex/source/ui/imgui_imhex_extensions.cpp
```diff
@@ -1216,6 +1216,89 @@ namespace ImGuiExt {
...
+        bool hovered, held;
+        bool pressed = ButtonBehavior(bb, id, &hovered, &held, ImGuiButtonFlags_None);
+        if (is_clipped) {
+            ImGui::PopClipRect();
+            return pressed;
+        }
+
+        // const ImU32 bg_col = GetColorU32((held && hovered) ? ImGuiCol_ButtonActive : hovered ? ImGuiCol_ButtonHovered : ImGuiCol_Button);
+        // window->DrawList->AddCircleFilled(bb.GetCenter(), ImMax(2.0f, g.FontSize * 0.5f + 1.0f), bg_col);
+
```
SparkyTD, line 1274: I left in these two commented lines in case we want to change the pin button to have a circle background, similarly to the close button. The main issue with this is that the ICON_VS_PINNED and ICON_VS_PIN icons are slightly offset, so if you center on of them inside the circle, the other one will look slightly off-center.

SparkyTD:
I found a bug that makes popup windows unusable after they've been dragged outside of the root window.

Repro:
1. Open the Select popup (or any other popup)
2. Drag the newly spawned window outside of the root window's bounds
3. Click anywhere else, e.g. on the hex editor to close the popup
4. Open the popup again (e.g. Ctrl+E)
5. After reopening, any mouse click within the popup will immediately close it.

plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -499,37 +557,67 @@ namespace hex::plugin::builtin {
...
+            alphaIsChanged = true;
         }
 
-        static bool justOpened = true;
+        if (m_currPopup != nullptr) {
+            if (ImGui::Begin(hex::format("##{}", m_currPopup->getTitle().get()).c_str(), &open, ImGuiWindowFlags_AlwaysAutoResize | ImGuiWindowFlags_NoDocking)) {
+                if (ImGui::IsKeyPressed(ImGuiKey_Escape)) {
+                    this->closePopup();
+                } else {
+                    float titleOffset = 7 * scaling;
```
WerWolv, line 576: Instead of manually multiplying by the scaling factor, you can use the `_scaled` UDL. So `7_scaled`

SparkyTD: [02dc312](https://github.com/WerWolv/ImHex/pull/1658/commits/02dc3124653e53581eaa6716092626b2bfd15ecf)

PR #1660, opened 2024-05-09
Fix jump scroll offset

SparkyTD:
### Problem description
This PR offers two improvements:
1) When selecting / jumping to an offset that falls within the current viewport of the scroll view, the scroll offset will no longer force the selected byte to the top of the view. Instead, the scroll offset will only be changed if the selected byte is outside the current view.

2) In case a wrong offset is entered into the Select or Goto dialog (e.g. and offset beyond EoF), the dialog's button will be disabled.

For the first change, I modified the logic that recalculates the `m_scrollPosition ` based on the current byte offset.

For the second change, I added validation logic to both popups to ensure that the entered offsets are valid (using `provider->getActualSize()`). In case of the Select popup, I wrapped the button into an `ImGui::Begin/EndDisabled` to enforce the validation check.

### DIFF
--- a/plugins/builtin/source/content/views/view_hex_editor.cpp
+++ b/plugins/builtin/source/content/views/view_hex_editor.cpp
@@ -88,12 +88,14 @@ namespace hex::plugin::builtin {
                     }
                 }
 
+                bool isOffsetValid = m_newAddress <= ImHexApi::Provider::get()->getActualSize();
+
                 bool executeGoto = false;
-                if (ImGui::IsItemFocused() && ImGui::IsKeyPressed(ImGuiKey_Enter)) {
+                if (isOffsetValid && ImGui::IsItemFocused() && ImGui::IsKeyPressed(ImGuiKey_Enter)) {
                     executeGoto = true;
                 }
 
-                ImGui::BeginDisabled(!m_newAddress.has_value());
+                ImGui::BeginDisabled(!m_newAddress.has_value() || !isOffsetValid);
                 {
                     const auto label = hex::format("{} {}", "hex.builtin.view.hex_editor.menu.file.goto"_lang, m_newAddress.has_value() ? hex::format("0x{:08X}", *m_newAddress) : "???");
                     const auto buttonWidth = ImGui::GetWindowWidth() - ImGui::GetStyle().WindowPadding.x * 2;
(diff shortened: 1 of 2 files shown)


plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -163,10 +165,18 @@ namespace hex::plugin::builtin {
...
-                if (ImGui::Button("hex.builtin.view.hex_editor.select.select"_lang) || (ImGui::IsItemFocused() && (ImGui::IsKeyPressed(ImGuiKey_Enter) || ImGui::IsKeyPressed(ImGuiKey_Enter)))) {
-                    editor->setSelection(m_region.getStartAddress(), m_region.getEndAddress());
-                    editor->jumpToSelection();
+                const auto provider = ImHexApi::Provider::get();
+                bool isOffsetValid = m_region.getStartAddress() <= m_region.getEndAddress() &&
+                        m_region.getEndAddress() < provider->getActualSize();
+
+                ImGui::BeginDisabled(!isOffsetValid);
+                {
+                    if (ImGui::Button("hex.builtin.view.hex_editor.select.select"_lang) || (ImGui::IsItemFocused() && (ImGui::IsKeyPressed(ImGuiKey_Enter) || ImGui::IsKeyPressed(ImGuiKey_Enter)))) {
```
WerWolv, line 174: Were you intending to use ImGuiKey_NumpadEnter here for one of the IsKeyPressed calls?

plugins/ui/source/ui/hex_editor.cpp
```diff
@@ -837,9 +837,20 @@ namespace hex::ui {
                         m_provider->setCurrentPage(m_provider->getPageOfAddress(newSelection.address).value_or(0));
 
                         const auto pageAddress = m_provider->getCurrentPageAddress() + m_provider->getBaseAddress();
-
-                        m_scrollPosition = (newSelection.getStartAddress() - pageAddress) / m_bytesPerRow;
-                        m_scrollPosition -= m_visibleRowCount * m_jumpPivot;
+                        const auto targetRowNumber = (newSelection.getStartAddress() - pageAddress) / m_bytesPerRow;
+
+                        // Calculate the current top and bottom row numbers of the viewport
+                        int currentTopRow = m_scrollPosition;
```
WerWolv, line 843: Use `ImS64` for the variables here, otherwise very large files will not work right

plugins/ui/source/ui/hex_editor.cpp
```diff
@@ -837,9 +837,20 @@ namespace hex::ui {
...
+                        int currentBottomRow = m_scrollPosition + m_visibleRowCount - 3;
+
+                        // Check if the targetRowNumber is outside the current visible range
+                        if ((ImS64)targetRowNumber < currentTopRow) {
```
WerWolv, line 847: Use C++-style casts

plugins/builtin/source/content/views/view_hex_editor.cpp
SparkyTD: I believe my other PR addresses this

SparkyTD: Specifically this change:
https://github.com/WerWolv/ImHex/pull/1658/files#diff-9232a67f4e1d1d29e0a0a8ee80aa0d39421cdc666a0bc22460976b49f27e249dL166

plugins/ui/source/ui/hex_editor.cpp
SparkyTD: You mean for `currentTopRow` and `currentBottomRow`?

plugins/ui/source/ui/hex_editor.cpp
SparkyTD: `static_cast<ImS64>()` right?

plugins/ui/source/ui/hex_editor.cpp
WerWolv: Yeah, I basically never use `int`s to begin with, only the explicitly sized types. But the important thing here is that you use a type that can fit the value you're assigning to it (therefor 64 bit)

plugins/ui/source/ui/hex_editor.cpp
WerWolv: I'd probably do `ImS64(targetRowNumber)` but either are fine

SparkyTD: [289a479](https://github.com/WerWolv/ImHex/pull/1660/commits/289a479a1897d2d527f165e7c7994907d420abdc)

PR #1664, opened 2024-05-10
impr: Ensure that TextEditor's cursor blink state is never invisible after a keyboard event

SparkyTD:
### Problem description
By default, the cursor's blinking cycle is independent from user actions, which feels unconventional while editing pattern code. This change ensures that the blinking cycle is reset, and the cursor is made visible every time the editor receives a keyboard event like character entry, or arrow navigation.

I moved the hard-coded blink timing numbers to their dedicated static constant fields since now they are referenced from multiple different places.

I added the function `void ResetCursorBlinkTime()` to `TextEditor.cpp` that will reset the blink cycle to `currentMillis() - sCursorBlinkOnTime`, ensuring that the cursor is visible. This function is called for every keyboard and mouse click event.

### Screenshots
Before:

https://github.com/WerWolv/ImHex/assets/45818400/668c6802-79a3-450b-80d3-d6abf2ce27be

After:

https://github.com/WerWolv/ImHex/assets/45818400/ee7f60e0-a75f-416d-b86d-8d12b5cdadf2


lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -22,6 +22,9 @@ bool equals(InputIt1 first1, InputIt1 last1, InputIt2 first2, InputIt2 last2, Bi
     return first1 == last1 && first2 == last2;
 }
 
+const int TextEditor::sCursorBlinkInterval = 800;
```
WerWolv, line 25: I think there's a style variable or something like that built into ImGui already that you could use here instead. That way the timing will be consistent with all other text input fields

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -2780,6 +2796,10 @@ int TextEditor::GetPageSize() const {
     return (int)floor(height / mCharAdvance.y);
 }
 
+void TextEditor::ResetCursorBlinkTime() {
+    mStartTime = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count() - sCursorBlinkOnTime;
```
WerWolv, line 2800: You can use ImGui::GetTime() for this instead

SparkyTD: The issue is that the `std::chrono` line returns the current unix millis timestamp, and the ImHex function returns the number of seconds since application start. I think there is one more instance of `std::chrono` being used for cursor blinking timing, I'll see if I can replace that one too.

SparkyTD: I replaced all instances of that long `std::chrono` line with `ImGui::GetTime() * 1000`, see the commit below.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
SparkyTD: I'm not sure I know what you're referring to. I found `ImGui::GetCurrentContext()->InputTextState.CursorAnim`, but it stays at `0.0` no matter wat. It's probably intended for stock input fields.

WerWolv: Nevermind me, it's actually hardcoded. Would it make sense to use the same timing values as they do though?

```cpp
bool cursor_is_visible = (!g.IO.ConfigInputTextCursorBlink) || (state->CursorAnim <= 0.0f) || ImFmod(state->CursorAnim, 1.20f) <= 0.80f;
```

SparkyTD: Changed it, now it's consistent with all other inputs :)
[0046760](https://github.com/WerWolv/ImHex/pull/1664/commits/0046760b929d0d54ade5cd556a7f5fa890b51e7f)

PR #1666, opened 2024-05-10
Add a note about installing via Homebrew

tgross35:
Homebrew always has updated versions and is an easy way to install, but was not mentioned in the installation instructions. Add a note about this.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -17,6 +17,12 @@ Extract the zip file to any location on your system.
 ### macOS
 Simply use the drag-n-drop dmg package to install ImHex on your system. It's possible that you need to allow the app to run in the security settings.
 
+Alternatively, ImHex can be installed using [Homebrew](https://brew.sh/):
+
+```bash
+brew install imhex
+```
+
 ### Linux
 
 #### AppImage


WerWolv:
Thank you! 
The reason why I didn't include it was because it's not officially maintained by us. Even if there's generally no issue with it, we have no control over what's being uploaded there so we can't really give support for it. Same thing for all other package manager builds, the only one that's actually maintained by us is the one on Arch's AUR.

I'm open for discussion about this policy though.

bhenzkie (approved):

iTrooz:
If you change the text to be under "Package managers", I think this PR can be accepted
What do you think @WerWolv ? Other package managers (like Chocolatey, Flatpak or Fedora) are not more official than brew, right ?

WerWolv:
@iTrooz Isn't the packaging status enough for that though? I kinda don't want to maintain the ever changing list of packages for 500 different package managers

iTrooz:
> I kinda don't want to maintain the ever changing list of packages for 500 different package managers

That's fair, but since we list third party package managers packages like Chocolatey and Flatpak, I think it would make sense to add at least one for MacOS.

But I agree, I don't want to maintain an ever-growing list either. Maybe only the most popular ones ? (Which homebrew is part of IMO)

Issue #1376, opened 2023-10-20
[Bug] Console is writing line numbers but they are in the wrong place so they are partially or compleetely hidden.


paxcut:
### Operating System

Windows

### What's the issue you encountered?

As the image below shows, and depending on the custom font of choice and the custom font size, you may see mall pieces of green dots showing to the left of the text lines. For the default size (13 pt.) they are completely hidden but after 16 pt. they start to show. The font size for the image below is 18 pt.

![image](https://github.com/WerWolv/ImHex/assets/53811119/12ebe8ec-7388-48aa-945c-aba02f497d4a)

### How can the issue be reproduced?

Since default font can't be resized, choose some other font and try setting the font size to 14 pt and up. Next either write a pattern that just prints lines of text to console or run an existing pattern. Observe the first unidentifiable patterns of green dots to the left. On higher pt. size the dots appear as numbers, but they are not located correctly so that only 1 digit is shown and even that one not completely. They should be either fully shown or completely removed.

### ImHex Version

1.31.0 release

### ImHex Build Type

- [X] Nightly or built from sources

### Installation type

built locally using clion

PR #1669, opened 2024-05-11
fix: TextEditor line numbers were still partially rendered despite mShowLineNumbers being set to false

SparkyTD:
### Problem description
TextEditor line numbers were still partially rendered despite mShowLineNumbers being set to false

I wrapped the line no. rendering code in `if (mShowLineNumbers) { ... }`.

### Screenshots
Before:
image

After:
image


SparkyTD:
Fixes #1376

PR #1670, opened 2024-05-11
WIP: Cap texture multisampling SampleCount

FireNX70:
### Problem description
https://gitlab.freedesktop.org/mesa/mesa/-/issues/11135
It turns out LLVMpipe only supports 4x multisampling. Checking GL_MAX_SAMPLES seems like the right thing to do.

Right now, I only check GL_MAX_SAMPLES. Depending on the format, we might need to check GL_MAX_INTEGER_SAMPLES. I don't know how likely it is that you might want to use a different format in the future, glGetInternalformativ might be a safer option to retrieve the max number of samples we can use.

### Additional things
I guess I could merge the ```if```s at lines 95, 99 and 103 in imgui_imhex_extensions.cpp while we're at it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vscode/
 .idea/
+.kdev4/
 
 cmake-build-*/
 build*/
@@ -7,6 +8,7 @@ local/
 venv/
 
 *.mgc
+*.kdev4
 imgui.ini
 .DS_Store
 ./CMakeUserPresets.json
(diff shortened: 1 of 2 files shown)


FireNX70:
Also, I guess it might be cleaner if I part out the GL_MAX_SAMPLES stuff  into its own function and we just call that to initialize SampleCount.

WerWolv:
Didn't think of that, thanks a lot!

PR #1673, opened 2024-05-13
impr: Added CSV, TSV and JSON as export options for Find results

SparkyTD:
### Problem description
The default result export functionality of the Find tool is limited to only exporting data in a nonstandard text format. This PR adds support for exporting the results in CSV, TSV or JSON format. The PR also removes the old format.

I added the classes `ExportFormatter`, `ExportFormatterCsv`, `ExportFormatterTsv` and `ExportFormatterJson`, with similar implementations to the pattern data exporters.

~~I also moved the `ViewFind::Occurrence` class into `hex/helpers/types.hh`, so the exporters can access it.~~

### Screenshots
image

### Additional things
Another small change I made is moving the "{} entries found" line on the same line as the Search and Reset buttons. I think it looks cleaner this way, but if anyone disagrees, I can revert it.


SparkyTD:
The CSV/TSV exporters have been tested by finding 5 character longs strings in /dev/urandom, and all the escaping logic seems to work as it's supposed to.

As for JSON, the nlohmann library already takes care of escaping strings.

SparkyTD:
There is an issue that causes the format selector popup to stay visible behind the main window in some cases when the user clicks the parent window. The bug also exists in the original pattern export popup. I haven't been able to fix it yet.

image

WerWolv:
Thanks a lot for the work!
I'd suggest adding these exporter registration things to the Content Registry instead, that way other plugins can re-use the same infrastructure you already wrote to add their own formatters. The Occurrences struct as well because now it's being imported by basically every single file in ImHex which is probably not needed :D

The actual implementations of the formatters can stay in the builtin plugin like it is

iTrooz:
@WerWolv Do you think this is ready to be merged ?

lib/libimhex/include/hex/api/content_registry.hpp
```diff
@@ -968,7 +968,22 @@ namespace hex {
...
+
+                using ExporterCallback = std::function<std::vector<u8>(const std::vector<FindOccurrence>&, std::function<std::string(FindOccurrence)>)>;
+                struct ExporterEntry {
+                    UnlocalizedString unlocalizedName;
+                    std::string fileExtension;
+                    ExporterCallback callback;
+                };
+
                 const std::vector<Entry>& getEntries();
+                const std::vector<ExporterEntry>& getExporterEntries();
```
iTrooz, line 986: Could you document the difference between getEntries() and getExporterEntries() ?

plugins/builtin/include/content/export_formatters/export_formatter.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <array>
+#include <memory>
+
+#include <hex/api/content_registry.hpp>
+
+namespace hex::plugin::builtin::export_fmt {
+
+    using Occurrence = hex::ContentRegistry::DataFormatter::impl::FindOccurrence;
+
+    class ExportFormatter {
```
iTrooz, line 14: Could you document this class ?

iTrooz (approved):

iTrooz (changes requested):
My bad, I marked this as "approved" instead of "request changes"

plugins/builtin/include/content/export_formatters/export_formatter.hpp
iTrooz: Ok ! Thanks

iTrooz (approved):

PR #1683, opened 2024-05-16
lang: Added Hungarian translations

SparkyTD:
### Problem description
ImHex didn't support Hungarian :(

I translated ImHex to Hungarian :)

### Translation Coverage
| Plugin        | Percentage |
|---------------|------------|
| builtin       | 99%        |
| diffing       | 100%       |
| disassembler  | 100%       |
| hashes        | 95%        |
| script_loader | 100%       |
| ui            | 100%       |
| visualizers   | 100%       |
| windows       | 100%       |
| yara_rules    | 100%       |

### Additional Notes
There are four Hungarian "special" characters that fall outside the default Unicode ranges loaded by ImHex, resulting in them being replaced with the "�" character. These letters are Ő (U+0150), ő (U+0151), Ű (U+0170) and ű (U+0171), all included in the Latin Extended-A Unicode block.

The easy fix for this is to include the "Unicode Latin Extended-A" range when loading the font glyphs in [init_tasks.cpp:189](https://github.com/WerWolv/ImHex/blob/99abc4e78a82407b9e25fe1cf811e166497ace28/plugins/builtin/source/content/init_tasks.cpp#L189). This change would also unlock the full character range of Bosnian, Croatian, Czech, Estonian, Latvian, Lithuanian, Maltese, Polish, Romanian, Slovak, Slovene and Turkish. I can add the commit to this PR if maintainers are okay with it. **EDIT:** Added a commit that loads Latin Extended-A by default.

Also note that some words are longer than their English counterparts, resulting in certain UI labels overflowing from their parent containers, and being cut off. I might change some of the longer labels to be more compact in the future, but the container size limitations will have to be addressed eventually (e.g. with horizontal scrollbars).


iTrooz:
> I can add the commit to this PR if maintainers are okay with it.

Isn't this change already in this PR ?

SparkyTD:
> > The easy fix for this is to include the "Unicode Latin Extended-A" range [...]. I can add the commit to this PR if maintainers are okay with it.
> 
> Isn't this change already in this PR ?

I added the change yesterday, but I forgot to update the PR description. Everything should be ready to merge now.

WerWolv:
Thank you very much!

PR #1688, opened 2024-05-18
build: Xcode accomodating CMake setup

mentlerd:
### Problem description
This PR implements some rudimentary Xcode support for building and editing ImHex.

#### Problem 1: Xcode is a multi-configuration buildsystem
The project is already rather CMake generator independent, thus it did not need to change much to support Xcode's multi-configuration paradigm:

By default, CMake generates a `.xcodeproj` in which targets build their artifacts into the specified `<>_OUTPUT_DIRECTORY`, postfixed by the currently active configuration. To better fit the existing paradigm, I instead opted ot introduce `IMHEX_MAIN_OUTPUT_DIRECTORY`. This variable is equal to the previously used `RUNTIME_OUTPUT_DIRECTORY` when using other generators, and is changed to include a configuration specific _prefix_ when used with Xcode.

The result is different output directories when using Xcode, and no changes when using any other generator.

#### Problem 2: ImHex does not support AppleClang
To allow building the codebase with Xcode, I have introduced `IMHEX_IDE_HELPERS_OVERRIDE_XCODE_COMPILER`. Specifying this option to `ON` will force CMake to honor the user specified compiler settings, even when using the Xcode generator.

In practice this can be used together with the new "xcode" CMakePreset to build the project with mainline clang using `xcodebuild`, or Xcode itself by generating a buildsystem like so:
```
cmake --preset xcode -DCMAKE_PREFIX_PATH=/opt/homebrew/opt/llvm@17
```

This solution is of course not without flaws. The inner workings are a particularly ugly hack, and mainline clang does not implement the necessary extensions to allow Xcode to index the code. Regardless this option is useful to enable future work in terms of bundling/signing macOS applications in the "intended" way using Xcode without additional source modifications.

#### Problem 3: Vanilla CMake + Xcode = Bad developer UX 
By default, the CMake generated `.xcodeproj` is a mess. Tons of targets are scattered about, and source files are not organized beyond grouping them into a "Source Files" and "Header Files" group.

Even "Header Files" is missing, because the ImHex build system does not regard private header files of libraries as sources of a target, and Xcode does not try to guess this information.

The solution is twofold:
* Additional code has been added which organizes the targets into a neat folder structure
* Additional code was added behind a configuration flag `IMHEX_IDE_HELPERS_INTRUSIVE_IDE_TWEAKS` which automatically creates source file trees in Xcode targets, and discovers the non-declared header files via the folder convention.

### Screenshots
N/A

### Additional things

As a bonus: `IMHEX_OFFLINE_BUILD` assumes that ImHex-Patterns is cloned into the source tree. I have added an additional fallback that tries to locate it as a sibling folder of `${CMAKE_SOURCE_DIR}`, as this meshes better with my filesystem setup.

The setup was tested with `CMake 3.29.2`, `Xcode 15.2`, and `llvm@17` from homebrew.


mentlerd:
Closing to test github contributor status

CMakeLists.txt
```diff
@@ -40,6 +44,13 @@ project(imhex
         HOMEPAGE_URL    "https://imhex.werwolv.net"
 )
 
+if (XCODE)
```
WerWolv, line 47: This should probably go into build_helpers.cmake to keep the main file simple and easy to read. (Using `PARENT_SCOPE`)

mentlerd: https://github.com/WerWolv/ImHex/pull/1688/commits/207406cc7d4e7ef5998bf279febec136c11ce373

PR #1689, opened 2024-05-19
impr: Restore native macOS title bar double click gesture in borderless mode

mentlerd:
#### Problem 1
In borderless mode ImHex disables the standard macOS titlebar rendering and input processing. As a result double clicking the titlebar does not trigger the native macOS behavior set in `System Settings -> Desktop & Dock -> Double-click a window's title bar to [Zoom/Minimize/Do nothing]`.

#### Problem 2
The ImHex window shows up as blank/transparent when de-minimizing it from the dock.

#### Problem 3
Widgets experience ghost hover inputs from the past position of the cursor during live resizing.

ImGui elements consume input events in the order they are drawn. As a result by "drawing" an `InvisibleButton` over the content area of the titlebar we can catch unprocessed clicks in the titlebar area. Connecting this button's double clicks to the native window is then a trivial endeavour.

The blank windows was caused by the rendering stack clearing the GL buffer, but proceeding to draw nothing in it. I have short circuited this path.

Ghost hover inputs were squelched by consistently moving the ImGui cursor to `0, 0` during a live resize. The OS will dispatch a cursor positioning event once the resizing ends, restoring normal behavior.

### Screenshots
N/A

### Additional things
N/A

### DIFF
--- a/lib/libimhex/include/hex/helpers/utils_macos.hpp
+++ b/lib/libimhex/include/hex/helpers/utils_macos.hpp
@@ -15,6 +15,8 @@
         void setupMacosWindowStyle(GLFWwindow *window, bool borderlessWindowMode);
 
         void enumerateFontsMacos();
+    
+        void toggleWindowZoomMacos(GLFWwindow *window);
     }
 
 #endif
--- a/lib/libimhex/include/hex/helpers/utils_macos.hpp
+++ b/lib/libimhex/include/hex/helpers/utils_macos.hpp
@@ -17,6 +17,7 @@
         void enumerateFontsMacos();
     
         void macosHandleTitlebarDoubleClickGesture(GLFWwindow *window);
+        bool macosIsWindowBeingResizedByUser(GLFWwindow *window);
     }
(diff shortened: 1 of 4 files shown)


WerWolv:
Is this really a native macOS feature? None of my Windows allow for that by default. I'm using the Rectangle app though that adds gestures like that which seems to work fine on ImHex as well.
Are you using a tool like that as well maybe?

mentlerd:
> Is this really a native macOS feature? None of my Windows allow for that by default. I'm using the Rectangle app though that adds gestures like that which seems to work fine on ImHex as well. Are you using a tool like that as well maybe?

It is, you can find the setting here:
image

On the other hand I was not aware that there are multiple options here. The code now respects all possible options: https://github.com/WerWolv/ImHex/pull/1689/commits/66223e09dd8f4b2f028c01c2e1331b4739e852dd

While testing I have also found a couple of small bugs relating to window resizing/minimizing. Squashed those too. I'll update the PR description in a moment

PR #1690, opened 2024-05-19
impr: Stop rubber banding while resizing on macOS

mentlerd:
### Problem description
On macOS `glfwSetWindowSizeCallback` is invoked early during window resizing, rendering a frame in that callback leads to wonky results as the new framebuffer is swapped before the OS has the chance to actually resize the window:

https://github.com/WerWolv/ImHex/assets/1068675/46336419-3fc2-4aa1-b16f-68b0c00e3584

Window contents are redrawn only from `glfwSetWindowRefreshCallback` during resizing, fixing the issue:

https://github.com/WerWolv/ImHex/assets/1068675/3acc5d4a-b2a5-42f0-9015-5e7172a027cf

### DIFF
--- a/main/gui/source/window/window.cpp
+++ b/main/gui/source/window/window.cpp
@@ -156,6 +156,10 @@ namespace hex {
     void Window::fullFrame() {
         static u32 crashWatchdog = 0;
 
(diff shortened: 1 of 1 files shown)


WerWolv:
God damn I tried everything to get rid of that and decided that it must be Apple's composer just doing that with OpenGL canvases. Thank you!

PR #1692, opened 2024-05-20
build: Allow specifying list of plugin dependencies for IDE launch schemes

mentlerd:
### Problem description
Launching ImHex from Xcode is clunky, as the `main` target does not depend on plugins. Therefore editing a source file, and simply launching the active Xcode scheme may result in the application using a previously built version of the file just edited.

From a **build system perspective** this is perfectly logical. The main executable can be built independently of any plugins, but this is very annoying UX.

This PR adds a new cmake boolean option `IMHEX_IDE_HELPERS_MAIN_DEPENDS_ON_PLUGINS` which does as the name imply.

Plugins in the build will be marked as a dependency of the `main` target, allowing a single-button launch from the IDE without having to think twice about which plugin was just edited.

### Screenshots
N/A

### Additional things
`_tweakTargetsRecursive` was mistakenly declared as a macro. This caused issues due to an early return in the implementation prematurely interrupting `tweakTargetsForIDESupport`. 

Nothing prevents the code from working as a function, and thus it has been converted to one.

### DIFF
--- a/cmake/ide_helpers.cmake
+++ b/cmake/ide_helpers.cmake
@@ -1,10 +1,7 @@
 
 option(IMHEX_IDE_HELPERS_OVERRIDE_XCODE_COMPILER "Enable choice of compiler for Xcode builds, despite CMake's best efforts" OFF)
 option(IMHEX_IDE_HELPERS_INTRUSIVE_IDE_TWEAKS    "Enable intrusive CMake tweaks to better support IDEs with folder support" OFF)
-
-set(IMHEX_IDE_HELPERS_MAIN_TARGET_DEPENDS_ON_PLUGINS "" CACHE STRING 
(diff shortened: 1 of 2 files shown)


iTrooz:
Why not use the target `imhex_all` instead of `main` ?

iTrooz:
I'm confused. Doing `ninja main` in my terminal doesn't launch ImHex ?

mentlerd:
> I'm confused. Doing `ninja main` in my terminal doesn't launch ImHex ?

It does in Xcode. That's why the entire thing is constrained in `ide_helpers.cmake`

iTrooz:
Isn't there a way to make ImHex run when compiling the `imhex_all` task ? ? This sounds like a XCode issue rather than ImHex

PR #1700, opened 2024-05-23
feat: full syntax highlights and interactive compiler error reporting indicated using a red squiggly line under the affected symbols.

paxcut:
- Added an under-squiggled text function to imgui extensions. 
- Syntax highlighter supports identifier validation and attempts to use contextually invalid identifiers results in error denoting coloration. 
- Moved shortcuts from text editor to shortcut manager in view_pattern_editor.cpp. 
- Removed all the language definition code.
- Added a text preprocessing function to remove /t and use standard newlines. 
- Changed uses of int and unsigned to i32 and u32 or equivalent. 
- Changed theme files to use the new names for the color palette entries, but they should remain the same for the most part. - Added a full syntax highlighting theme for pattern language based on the previous Dark theme and clion's darcula which is called Clark 
- Search-replace in pattern editor retained found matches when text was edited. 
- Incorporated blinking cursor resetting to new functions that move the cursor and streamlined previous entries under mouse editing.


paxcut:
Note that this changes won't compile until the latest pattern language submodule pr that i submitted is merged and the pattern language submodule is updated in imhex to use the merged code.

paxcut:
I am turning this one into a draft as well as it is being implemented as a set of smaller PRs for the sake of easier reviewing. Once the set is complete I'll close it but I want to keep it around so I can make sure everything was included.

PR #1747, opened 2024-06-09
allow one byte ranges edits

Murmele:
### Problem description

When entering for the end value the same value as the start it is not recorgnized as valid region, because end must be strictly greater than begin. Due to the +1 in the Region constructor this is not correct, because the end is included in the range.

### Screenshots

After:
image


WerWolv:
Thank you!

Murmele:
Thanks for the fast merge!

Issue #1733, opened 2024-06-03
[Bug] Export selection to file infinite loop


kyle-sylvestre:
### Operating System

Windows

### What's the issue you encountered?

This is a test of fix #1619 
The created file contains everything except the last byte still.
The task now loops infinitely.

![Untitled](https://github.com/WerWolv/ImHex/assets/25188464/579fdf4c-6aaa-4500-a605-8bfa8f34db45)

### How can the issue be reproduced?

1. open file
2. select a range of bytes
3. click File > Export > Selection to File...
4. select output file
5. observe task stuck in an infinite loop

### ImHex Version

1.34.0

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

imhex-1.34.0-Windows-Portable-x86_64.zip

Nemoumbra:
Thank you for your input! We have found this bug too, I just didn't create an issue (hoped in would be resolved prior to the release).

Issue #1619, opened 2024-03-27
[Bug] Export selection to file omits last byte


kyle-sylvestre:
### Operating System

Windows

### What's the issue you encountered?

Export selection to file omits the last byte of the selection.

### How can the issue be reproduced?

1. highlight a section of bytes
2. export selection to file
3. open file to see it is missing the last byte

### ImHex Version

1.33.2

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

portable

paxcut:
I have tested this on a recent nightly build. I selected 21 bytes and from the export sub-menu I chose export to file and ImHex saved all 21 selected bytes and not a single one was not exported and the exported file didn't have invalid data or anything abnormal that I could see. It should be safe to close this issue.

Issue #1597, opened 2024-03-16
[Bug] Export selection to file processes for ever at full CPU load.


malespiaut:
### Operating System

Linux

### What's the issue you encountered?

I wanted to export a selection (40 bytes) to a file.

ImHex is showing the “Processing” bar at the bottom of the screen, but nothing happens while the CPU gets hot.

ImHex can't be closed.

I've killed ImHex with xkill, and it left a ImHex process running at 100% in the background.

### How can the issue be reproduced?

Select a part of a file, and choose “Export -> Selection to file” in the file menu.

### ImHex Version

1.33.1

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

Using AUR on Arch Linux, package “imhex”.

PR #1774, opened 2024-06-26
Fix off by one in "Export selection to file"

FireNX70:
### Problem description
Export selection to file would never finish and wouldn't export the last byte.

It was just a simple off by one when using the selection's end address to calculate the remaining data size. Add one to the remaining size calculation.

### Additional things
Fixes https://github.com/WerWolv/ImHex/issues/1733, https://github.com/WerWolv/ImHex/issues/1619 and https://github.com/WerWolv/ImHex/issues/1597.


WerWolv:
Thank you!

PR #1804, opened 2024-07-08
fix: Segfault when hashing regions spanning multiple MiBs

jfng:
### Problem description

Attempting to do an MD5 hash of a large region (e.g. 2 MiB, ``u8 data[0x200000]``) crashes with a segfault.

In ``hex::plugin::hashes::hashProviderRegionWithHashLib()``, ``hashFunction->TransformBytes()`` is called with an offset of 0, because it iterates over ``data`` and not the entire region.


WerWolv:
Thanks a lot!

Issue #1814, opened 2024-07-15
[Bug] Double-free on exit in font manager plugin with custom font


csnover:
### Operating System

Linux

### What's the issue you encountered?

ImHex segfaults on exit in the `FontAlias::m_fontData` destructor:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007ffff70a6b6f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007ffff70584e2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff70414ed in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007ffff7042395 in __libc_message (fmt=fmt@entry=0x7ffff71b655e "%s\n") at ../sysdeps/posix/libc_fatal.c:150
#5  0x00007ffff70b0455 in malloc_printerr (str=str@entry=0x7ffff71b94f8 "corrupted size vs. prev_size while consolidating") at ./malloc/malloc.c:5765
#6  0x00007ffff70b2394 in _int_free_merge_chunk (av=av@entry=0x7fff74000030, p=0x7fff7402b0b0, size=382560) at ./malloc/malloc.c:4688
#7  0x00007ffff70b2669 in _int_free (av=0x7fff74000030, p=<optimized out>, have_lock=<optimized out>, have_lock@entry=0) at ./malloc/malloc.c:4639
#8  0x00007ffff70b4e2f in __GI___libc_free (mem=<optimized out>) at ./malloc/malloc.c:3391
#9  0x00007fff97a17c5b in std::__new_allocator<char>::deallocate () at /usr/include/c++/13/bits/new_allocator.h:172
#10 std::allocator<unsigned char>::deallocate () at /usr/include/c++/13/bits/allocator.h:210
#11 std::allocator_traits<std::allocator<unsigned char> >::deallocate () at /usr/include/c++/13/bits/alloc_traits.h:517
#12 std::_Vector_base<unsigned char, std::allocator<unsigned char> >::_M_deallocate () at /usr/include/c++/13/bits/stl_vector.h:390
#13 std::_Vector_base<unsigned char, std::allocator<unsigned char> >::~_Vector_base () at /usr/include/c++/13/bits/stl_vector.h:369
[...]

PR #1815, opened 2024-07-15
fix: Crash on exit when using custom font from file

csnover:
### Problem description
WerWolv/ImHex#1814

1. Remove the flag that wrongly tells `ImFontAtlas` that it owns font data, and remove that parameter from the function signature entirely since now it is always `false`.
2. Rename `takeAtlas` to `getAtlas` since it no longer transfers ownership as of b652565b57bd8a9aac2779b09f51f7bac26afca2.


WerWolv:
Thanks a lot!

PR #1817, opened 2024-07-16
build: fix incorrect name case in FindmbedTLS.cmake

xndcn:
Corrected MBEDTLS_FIND_REQUIRED to mbedTLS_FIND_REQUIRED, since the package name is "mbedTLS" for case sensitive environment


WerWolv:
Thanks!

PR #1818, opened 2024-07-16
fix: Fix lzma stream end constant name

xndcn:
Corrected BZ_STREAM_END to LZMA_STREAM_END.
Also removed redundant condition of LZMA_STREAM_END for error checking.


WerWolv:
Derp, thanks a lot!

PR #1821, opened 2024-07-18
feat: Enumerate fonts on Linux using Fontconfig when available

csnover:
### Problem description

The fonts list on Linux does not show all system fonts, and does not show font names at all.

Use Fontconfig to make the list less bad if Fontconfig headers are available.

### Additional things

I like fonts.


WerWolv:
Thank you!

PR #1822, opened 2024-07-18
fix: Missing --help output & other diagnostic messages in console

csnover:
### Problem description

Typing `--help` causes ImHex to exit without outputting anything. Diagnostic messages from glib, ASan, other libraries that might have something important to say, etc. are also suppressed.

This effectively reverts 7c1e33dde6e55ef6347f3b10868a42f456c2ae6a, which was partially reverted only on Windows by code that was left commented out in f114239f51ba68c385e3d46721dd9d97ca4a6c21.

Allowing other libraries to print to stderr may make the output ‘ugly’, but lots of things print to stderr that are important for figuring out why something is bugged, like ASan and glib.

### Additional things

image

### DIFF
--- a/main/gui/source/window/linux_window.cpp
+++ b/main/gui/source/window/linux_window.cpp
@@ -14,7 +14,6 @@
     #include <wolv/utils/core.hpp>
 
     #include <nlohmann/json.hpp>
-    #include <cstdio>
     #include <sys/wait.h>
     #include <unistd.h>
 
@@ -96,11 +95,6 @@ namespace hex {
             if (std::fs::exists(path))
                 setenv("LD_LIBRARY_PATH", hex::format("{};{}", hex::getEnvironmentVariable("LD_LIBRARY_PATH").value_or(""), path.string().c_str()).c_str(), true);
         }
-
-        // Various libraries sadly directly print to stderr with no way to disable it
-        // We redirect stderr to /dev/null to prevent this
-        wolv::util::unused(freopen("/dev/null", "w", stderr));
-        setvbuf(stderr, nullptr, _IONBF, 0);
 
         // Redirect stdout to log file if we're not running in a terminal
         if (!isatty(STDOUT_FILENO)) {
--- a/main/gui/source/window/macos_window.cpp
+++ b/main/gui/source/window/macos_window.cpp
@@ -11,7 +11,6 @@
     #include <hex/helpers/logger.hpp>
     #include <hex/helpers/default_paths.hpp>
 
-    #include <cstdio>
     #include <unistd.h>
 
     #include <imgui_impl_glfw.h>
@@ -40,11 +39,6 @@ namespace hex {
             if (std::fs::exists(path))
                 setenv("LD_LIBRARY_PATH", hex::format("{};{}", hex::getEnvironmentVariable("LD_LIBRARY_PATH").value_or(""), path.string().c_str()).c_str(), true);
         }
-
-        // Various libraries sadly directly print to stderr with no way to disable it
-        // We redirect stderr to /dev/null to prevent this
-        freopen("/dev/null", "w", stderr);
-        setvbuf(stderr, nullptr, _IONBF, 0);
 
         // Redirect stdout to log file if we're not running in a terminal
         if (!isatty(STDOUT_FILENO)) {
--- a/main/gui/source/window/win_window.cpp
+++ b/main/gui/source/window/win_window.cpp
@@ -29,8 +29,6 @@
     #include <shellapi.h>
     #include <timeapi.h>
 
-    #include <cstdio>
-
 namespace hex {
 
     template<typename T>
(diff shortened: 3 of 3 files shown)


WerWolv:
The reason why I turned these off was because libmagic just spammed hundreds of lines of bogus log messages to stderr without any way to turn it off. That's just really terrible and outdated library design.

I'd argue that we should keep this for the sake of usability of the logs and instead only disable stderr in release builds where you won't be using diagnostics tools or asserts anyways.

csnover:
Do you know where in libmagic this is happening, or how I could trigger it? Looking at the source code it seems like debugging output is gated behind `MAGIC_DEBUG` flags and I don’t remember `file` ever emitting messages.

WerWolv:
When loading compiled magic files or compiling some magic source files, it would spew out hundreds of warning / error logs to the console. `MAGIC_DEBUG` isn't enabled anywhere in ImHex to my knowledge

csnover:
Oops. I see `MAGIC_CHECK` flag is being used. Emitting diagnostic messages to stderr is also its intended purpose according to magic.h:

```
#define	MAGIC_CHECK		0x0000040 /* Print warnings to stderr */
```

There are a couple of other places where messages look like they are unconditionally emitted via `file_magwarn` in libmagic when there is an error with the magic file but certainly that would not seem like a thing that would flood the stderr like `MAGIC_CHECK` (which is the flag used for `file -c`) does.

To me it seems like the thing to do for libmagic specifically could be to *temporarily* redirect stderr to collect the emitted warnings while it is running rather than suppressing all stderr output globally. Presumably if someone is using a custom magic file and there is some syntax error it would be helpful for them to see that; getting a single error from `magic_error` seems like it would be less useful. But, I don’t know how people are using this feature in practice, so maybe just stopping asking libmagic to emit warnings to stderr is the thing to do.

`--help` should emit regardless of debug or release mode; it is also possible to run release version of ImHex with e.g. debug libraries dynamically linked that emit diagnostic information to stderr, so I don’t know that it is good enough only suppress the output on a release version. It may be worthwhile to unconditionally redirect stderr messages into the normal logging system, though, so they are not absent from the log window.

csnover:
I agree that libmagic seems like a mess of a thing that probably started its life as an executable and only got half-abstracted out into a library. I am no fan of C—especially error handling in C libraries, which is almost universally awful—and also agree that in an ideal world things would not be this way. Unfortunately, the standard mechanism for emitting most diagnostic information today is still just writing to `stderr`. So, when there is a conflict between what an application author wants and what the end user or platform convention wants, who gets to win? In such cases, I try to think about the benefits versus harms to users. In this case, assuming ImHex is fixed to not ask libmagic to write a lot of stuff to `stderr`, I see it this way:

What benefit is there to users to suppress `stderr`? If a user is not looking at the terminal output, their experience is unchanged. If they have some OCD, they will be less upset that message formatting in the console is not uniform. If irrelevant diagnostics are emitted, they will not be alarmed or misled by their existence. They will not need to be aware of their own ability to redirect or filter `stderr` to suppress unwanted messages from libraries.

What harm is there to users to suppress `stderr`? It breaks platform convention and creates confusion for users who expect `stderr` to behave in the standard way. Errors from libraries that cannot unwind the stack, but that emit to `stderr` before calling `abort`, will be missing, which makes crashes harder to understand and troubleshoot. Other non-fatal diagnostic information potentially relevant to an issue will be discarded. Environment variables like `WAYLAND_DEBUG=1`, which are relevant to both debug and release builds of any application, stop working. Some (hypothetical) plugins that might use `stderr` for diagnostics and `stdout` for output will be impossible to author/use as ImHex suppresses `stderr` and emits its own diagnostics to `stdout`.

As far as why `--help` was not working: I am not sure of the answer to this. vscode’s built-in terminal may have been doing something weird (vscode’s terminal has definitely [broken other things before in weird ways](https://github.com/microsoft/vscode/issues/184713)), or maybe I was wrong about what happened and ImHex was crashing before it was able to emit the help output at all. There are some nuances of stdio that are beyond my knowledge. At least the lack of `--help` output was what clued me in that ImHex was messing with `stderr`, since some apps emit help on `stderr`.

Originally, ImHex was crashing due to a memory error, and glibc was emitting a diagnostic, but ImHex was suppressing it. I occasionally have three brain cells to rub together and eventually ran `strace` which showed the `write` call containing part of the suppressed error. (Attaching a debugger is not good enough AFAIK, as it will only trap on the `abort`, at which point the frame with the diagnostic message will usually be gone.)

So, these are my thoughts. Maybe you have a different idea about which things are most important and that’s fine, do whatever you think is best. This did cause me to waste time and energy and I like to try to make it so others don’t have to waste time and energy too.

csnover:
Probably neither here nor there at this point but vscode’s built-in terminal does seem to do something weird with file descriptors that other terminals don’t that seems to be able to cause applications to break. I tried to run flatpak wine inside of a vscode terminal, which works fine in a normal terminal, but hangs with 100% CPU forever in vscode. strace shows `fnctl(<monotonically increasing number>, F_SETFD, FD_CLOEXEC) = -1 EBADF`. I don’t plan on debugging this further but just figured I would drop a note to say I encountered some other dumb broken vscode terminal thing about FDs.

PR #1826, opened 2024-07-22
Update zh_CN.json

zhongqingsong:
### Problem description
Some Chinese translations are missing, and the order of Json's key in old documents is inconsistent with that in English. As a result, it is difficult to make a comparison during subsequent maintenance, which increases the maintenance cost of translation.

According to the English json, sync the latest text, as well as adjust the partial translation. 
Add new text, and translate to CN.

### Screenshots
None

### Additional things
None


xtexChooser (commented):
LGTM. By the way, FYI, you could use `dist/langtool.py` for translating.

WerWolv:
Merged together with #1894, thanks a lot!

PR #1833, opened 2024-07-30
fix: GLFW backend detection on Linux in fs.cpp

### LINKED ISSUE #1832: [Bug] File dialog is no longer parented properly on Linux X11 after 48fc1a7a1e90fb596662c4d59449aa011daefade
### Operating System

Linux

Previously, opening a file dialog will cause the dialog to stay on top of the ImHex main window until the dialog is completed, but now it no longer does that (you can click on the main window to bring it above the dialog).

I've tested this with the Ubuntu 22.04 build from the CI, on a reasonably fresh Ubuntu VM.

### ImHex Version

efee128c1c1741a38e82aff6fb87bc54be5f879e

### ImHex Build Type

- [X] Nightly or built from sources

### Installation type

Ubuntu 22.04 deb file

### Additional context?

I used a reasonably clean Ubuntu 22.04 VM (which uses X11).

btzy:
This is an attempt to fix the issue described in #1832, where file dialogs on Linux/X11 are no longer properly parented.

I don't really know the reason for 48fc1a7a1e90fb596662c4d59449aa011daefade and cf2e1890496773b46bef747fdce19c903d2e0d9f, but it seems I had wrongly assumed that all GLFW builds support X11 and all GLFW builds that define GLFW_WAYLAND_APP_ID support Wayland.  This is not true, and so some Wayland-only GLFW builds probably broke due to this.  There seems to be no way to determine the supported backends of the GLFW library by looking at its public headers, and so I've made the build script compile two small programs as a way to detect the supported backends of the available GLFW library.

### Problem description
This fixes #1832.

Properly detect the supported backends of the available GLFW library by compiling two test C++ programs.

### DIFF
--- a/lib/libimhex/CMakeLists.txt
+++ b/lib/libimhex/CMakeLists.txt
@@ -80,6 +80,33 @@ else()
     target_compile_definitions(libimhex PRIVATE IMHEX_PROJECT_NAME="${PROJECT_NAME}")
 endif()
 
+if (UNIX AND NOT APPLE)
+    # For Linux, we need to figure out whether our GLFW is built for X11, Wayland, or both
+    INCLUDE(CheckCXXSourceCompiles)
+    set(CMAKE_REQUIRED_INCLUDES ${GLFW_INCLUDE_DIRS})
+    set(CMAKE_REQUIRED_LIBRARIES ${GLFW_LIBRARIES})
+    check_cxx_source_compiles("
+        #define GLFW_EXPOSE_NATIVE_X11
+        #include <GLFW/glfw3.h>
+        #include <GLFW/glfw3native.h>
+        Window test(GLFWwindow* w) { return glfwGetX11Window(w); }
+        int main() {}
+    " GLFW_HAS_X11)
+    check_cxx_source_compiles("
+        #define GLFW_EXPOSE_NATIVE_WAYLAND
+        #include <GLFW/glfw3.h>
+        #include <GLFW/glfw3native.h>
+        wl_surface* test(GLFWwindow* w) { return glfwGetWaylandWindow(w); }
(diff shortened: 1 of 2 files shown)


WerWolv:
Thanks a lot!

Yeah sorry about breaking what you PRd but it caused build issues when trying to build with wayland GLFW 3.3 and I didn't have the energy to fix it properly. 
GLFW 3.4 now doesn't have separate versions for X11 and Wayland anymore and supports everything in one package but e.g Ubuntu will probably never update to that version. 

Your PR fixes the build issue for sure, however I don't think it actually makes things work better at runtime. All the Linux CI runners build using the X11 version of GLFW so the compile time check will always result in the X11 code being compiled in. If the user now installs the wayland version of GLFW on their system, ImHex will most likely either fail to start or crash when it tries to call those functions. 
The only real solution I think would be to use `dlopen` and `dlsym` to try and get the right symbols at runtime and choose the correct implementation based on that.

btzy:
> If the user now installs the wayland version of GLFW on their system, ImHex will most likely either fail to start or crash when it tries to call those functions.

Are you sure about that?  I'd be pretty surprised if a program compiled with regular GLFW (say https://packages.ubuntu.com/jammy/libglfw3) can be made to use the Wayland package (say https://packages.ubuntu.com/jammy/libglfw3-wayland) at runtime.  That would be an ABI break.

Those wanting to use Wayland on a distro that doesn't support having both backends in a single package will need to compile against `libglfw3-wayland` instead of `libglfw3`, and so the build-time detection will work.

btzy:
Closing after separate discussion. People indeed swap out glfw-x11 for glfw-wayland at load time without recompiling ImHex

Issue #1796, opened 2024-07-05
[Bug] ImHex is picked up as 'unresponsive' while using file dialogues


applecuckoo:
### Operating System

Linux

### What's the issue you encountered?

When you open a file dialogue, Ubuntu will think that ImHex is unresponsive.

### How can the issue be reproduced?

1. Open a file saving dialogue
2. Wait a few seconds
3. You'll see that there is a pop-up on the ImHex main window that says 'ImHex is not responding'

### ImHex Version

1.35.3

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

.deb file

### Additional context?

Tested on Ubuntu 24.04

WerWolv:
Hey
I'm not sure there's a real solution for this. When the file picker opens, it puts ImHex into a spin-loop so no events are happening in the background anymore and that triggers gnome to think that the app went unresponsive.

Maybe @btzy can answer this better than I can.

btzy:
Do you mind checking if this bug also occurs before https://github.com/WerWolv/ImHex/commit/dd607621d79615d8f70a2ede00c1d795c0eef693?

This needs more investigation but my guess is that you're using portals and gnome doesn't realise that the out-of-process dialog is logically part of ImHex.

WerWolv:
This has always been the case. I just assumed that's how it works when talking to the dbus on Linux

applecuckoo:
@btzy Yeah, definitely before dd60762 - I usually don't use nightlies and only tried them out specifically for this. I know this isn't intended since this is usually the only time it triggers, even GHex doesn't do this.

btzy:
I don't currently have Ubuntu 24.04 set up, but I installed the "Ubuntu 22.04 DEB x86_64" package downloaded from GitHub Actions on a mostly clean Ubuntu 22.04 minimal VM and couldn't reproduce the issue even after waiting for a minute.  I also don't see any warning pop-up when running NFDe's SDL2 test binary (which can also be downloaded from GitHub Actions if you're unwilling to set up the repo).

Does this only occur on Ubuntu 24.04?

Since GHex is part of gnome, I guess it might use GTK directly instead of portals.

btzy:
Actually it does occur on Ubuntu 22.04 too, but the warning stays under the file dialog.  The warning goes away quickly once control is returned back to ImHex, so it could be simply disappearing fast enough.

WerWolv:
[...]

PR #1834, opened 2024-07-31
fix: File dialogs on Linux causing a crash on systems where GTK uses Wayland but GLFW uses X11

btzy:
There has been reports of the file dialog crashing ImHex on Linux, including [this](https://github.com/WerWolv/ImHex/issues/1796#issuecomment-2229694812) and [this](https://github.com/WerWolv/ImHex/issues/1796#issuecomment-2248579290).  I think this occurs on systems where Wayland is the default compositor but the GLFW used in ImHex only supports X11.

The default Ubuntu 22.04 and Ubuntu 24.04 builds of ImHex use the GLFW from Ubuntu's default package repository, which supports only X11 but not Wayland.  GLFW is only built with Wayland support from Ubuntu 24.10 onwards.  This can be seen from the CI log https://github.com/WerWolv/ImHex/actions/runs/10167117281, which contain something like the lines below in the CMake configure step (they are added by #1833).
```
-- Performing Test GLFW_HAS_X11
-- Performing Test GLFW_HAS_X11 - Success
-- Performing Test GLFW_HAS_WAYLAND
-- Performing Test GLFW_HAS_WAYLAND - Failed
```
These tests detect if the GLFW package that ImHex is being built against supports X11 or Wayland (or both).

Using a build of ImHex like those above means that ImHex must use an X11 window, even when Wayland is available (e.g. with Sway).  However, the system GTK may support Wayland, and even default to Wayland with depending on the system configuration.  As the Ubuntu builds of ImHex use NFDe with the GTK backend and use the system GTK package, GTK may open a Wayland window.  An X11 window cannot be made the parent of a Wayland window, thereby causing the crash.

NFDe has been updated to force GTK to use the same compositor as the parent window, and this PR fixes ImHex by updating NFDe.

### DIFF
--- a/lib/libimhex/CMakeLists.txt
+++ b/lib/libimhex/CMakeLists.txt
@@ -80,6 +80,33 @@ else()
     target_compile_definitions(libimhex PRIVATE IMHEX_PROJECT_NAME="${PROJECT_NAME}")
 endif()
 
+if (UNIX AND NOT APPLE)
+    # For Linux, we need to figure out whether our GLFW is built for X11, Wayland, or both
+    INCLUDE(CheckCXXSourceCompiles)
+    set(CMAKE_REQUIRED_INCLUDES ${GLFW_INCLUDE_DIRS})
+    set(CMAKE_REQUIRED_LIBRARIES ${GLFW_LIBRARIES})
+    check_cxx_source_compiles("
+        #define GLFW_EXPOSE_NATIVE_X11
+        #include <GLFW/glfw3.h>
+        #include <GLFW/glfw3native.h>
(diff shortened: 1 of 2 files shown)


jumanji144:
Is this pr still a draft or can it be reviewed?

btzy:
This PR can't be merged as is, because the commit https://github.com/WerWolv/ImHex/commit/a24692b4beb59240578dec19a496ace30f974a27 has bumped the nativefiledialog version already, and the detection of X11 and Wayland is problematic in Linux due to people swapping out the `glfw` package for the `glfw-wayland` package after installation.

It would be ideal to reapply a modified commit that was reverted though, because it keeps the file dialog above the main window - I can raise a PR if there is appetite for it.

PR #1840, opened 2024-08-06
build: Find boost with name "Boost"

xtexChooser:
On many distributions like AOSC OS, Alpine Linux, Arch Linux, etc., boost should be searched with name "Boost".

This should fix the packaging issue on AOSC OS.

### Problem description
CMake fails to find system boost when `USE_SYSTEM_BOOST` is set to ON.

Find Boost.


shenlebantongying:
Yes, the original code is wrong, but there is no need to test both names.

The original code is copy-paste error. The lower case was copied from Imhex's custom cmake file for `boost`
* https://github.com/WerWolv/ImHex/blob/0263d3538e7a70db458833ad502c66366886eb02/lib/third_party/boost/regex/CMakeLists.txt#L10C13-L10C18

Both (deprecated) `FindBoost.cmake` in CMake's tree and the upstream's `BoostConfig.cmake` uses cap `Boost`.

* https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake
* https://github.com/boostorg/cmake/blob/develop/config/BoostConfig.cmake

The entire GitHub doesn't use lowercase boost for find_package.

* https://github.com/search?q="find_package(boost"

xtexChooser:
Thanks for your search! I removed the lower case completely.

WerWolv:
Thanks a lot

PR #1850, opened 2024-08-15
fix: Unable to resize window in the x-axis. 

paxcut:
The child had the same width as the parent, and it needs some minimal space so that the resizing handle can move. Also needed to change how the reset button chooses its width so that the parent shrinks with the child.

improv: defining all index sizes and using only one is not only wasteful but also confusing. Switched to a template based system for simplicity and readability.
fix: the recently added check for validity of indices had reversed logic and the wrong validity condition. Also added checking for compatibility in the sizes of all the variables that are defined to avoid crashes that were reported and whose occurrence was confirmed. Local tests confirmed the fixes resolve the issue.
fix: bizarre slow down of 3d window interactions when the axes display was removed and the 3-d visualizer window was contained entirely within ImHex window. Strange is that it seems to be related to Imgui time tiks so that if the option to show resources is removed then the 3d window will cease to move completely but if the option is on then the interactions update at the exact same time as the percent cpu display. Although The error still exists on the latest nightly from 5 days ago I am unable to reproduce it one these code changes.I don't know how it was fixed, but fixed it seems.


paxcut:
All fixes implemented here have been added as individual separate PRs so this one is no longer necessary.

Issue #1663, opened 2024-05-09
[Bug] 3d visualizer becomes extremely slow without axis disabled


mandar1jn:
### Operating System

Windows

### What's the issue you encountered?

Whenever the axis are disabled for the 3d visualizer, every rotation takes a few seconds to apply and only applies after having let go of the middle mouse button.

### How can the issue be reproduced?

open a 3d model file and run its pattern
open the 3d visualiser
disable the axis (every other option doesn't matter)
try to rotate the model. Wait a bit between interactions
enable axis again
successfully rotate the object accurately again

### ImHex Version

1.33.2 (latest nightly

### ImHex Build Type

- [X] Nightly or built from sources

### Installation type

MSI

### Additional context?

unable the model due to copyright concerns. Please contact me on discord if needed (same username)

paxcut:
To be able to reproduce this issue several conditions must be met that are not described in the description above. The conditions also affect how the issue manifests itself. The conditions are:
1) There is no text (2d) being displayed in the 3d view. For most that means turning the axes display off. For some, if running a debug build, it is impossible because the mouse coordinate display cannot be turned off.
2) The entire 3d visualizer window must fully overlap the main ImHex display.  To clarify, the issue won't be reproduced if the 3d visualizer window intersects the main ImHex window outer border or if it lies completely outside.
Under those conditions you will experience lag on the rotations and translations and scaling of the model. Just how much lag depends on  whether you have the option Extra>Settings>Interface tab=>"Show resources in footer" on or off. If it is on then the model will update when the % cpu updates to show a new value. If it is off the model updates when you hover over the icons or the reset button. If you are rotating the model with the middle click, the model will not update by simply lifting the middle button, you have to hover the items described for that to happen.

paxcut:
The reason for the bug, as will be shown soon, is that the lag occurs when there are no draw calls changing between frames despite the fact that the 3d contents is changing. The reason the 3-d changes become invisible to the optimization code should be obvious. There are no draw calls being made from the opengl code that renders the visualization. It is true that a texture is created, but it isn't being displayed using draw calls. To see this we turn to the Imgui demo and the `ShowMetricsWindow` function which allows the highlight of draw calls in real time. The reason why this wasn't done before is that the Imgui demo requires you to click open the node you want to show the draw calls of and hover it. First you can't open the node because as soon as you click on the tool's window the 3-d visualizer window closes. Even if you could open it, the focus will remain on the 3d window so hovering other windows wont register.

[...]

PR #1866, opened 2024-08-24
fix: for issue #1663 (3d visualizer becomes very slow without...)

paxcut:
### Problem description
The bug described  in #1663 was caused by an optimization of the display rendering that uses drawlists to detect changes. The changes in the 3-d visualizer don't contain drawlists when axes are turned off.

The fix is to identify the 3d visualizer among the command lists of the main window and, if found, avoid skipping frames regardless of the result of the comparison of drawlists.


WerWolv:
This will be reworked again soonish but for now it's a good think to have fixed. Thank you

PR #1872, opened 2024-08-25
fix: Textures, uv coordinates and light source not updating properly.

paxcut:
### Problem description
The only time textures would update was through the file picker. Once UV coordinates were used, disabling them didn't have any effect because the vertex array didn't invalidate the corresponding buffer array. When `shouldUpdate` is true, the light source needs to know as well, otherwise it will set the location at 0,0.

Textures were fixed by creating a member variable that holds the file name of the texture used in the last model rendering.

Instead of invalidating buffer arrays it is much simpler to define a default UV coordinate set for the case when no UV is specified. if a texture is not present then the values of UV will not be used. If there is a texture it must be a minimum of 1 pixel in size. So we choose the UV coordinates so that every vertex gets assigned the color at the 0,0 coordinate of the texture.

When `shouldUpdate` is on, we also turn `shouldUpdateLightSource` on.

Also included are some formatting changes that are purely aesthetic.


WerWolv:
Might have merged the other PRs in the wrong order, sorry. There's a few merge conflicts here now

paxcut:
No problem, I'll take a look at it now.

PR #1879, opened 2024-08-29
Some build system tweaks

BobSmun:
While setting up a dev environment using WSL2 and docker, to better isolate the tools required from those required by other projects, I made the following tweaks.
The primary aim was to improve IDE integration for the debugging process, by allowing the resulting binaries to be run directly from their build output location. Although cmake is pretty resilient in terms of tracking where it places generated outputs, for use in the later install step, I don't have any of the other environments set up (such as xcode or github workflows) to verify that I didn't break any assumptions they may have.

* Adds `.devcontainer` to the `.gitignore` - I think this directory may be vscode + remote container extension specific, so ignoring it in the same vein as ignoring e.g. `.vscode`
* Adds `.cache` and `compile_commands.json` to the `.gitignore`, as these are generated files / directories used by clangd (via vscode extension)
* Enables `CMAKE_EXPORT_COMPILE_COMMANDS` - useful for other tooling (such as clangd), but harmless on its own
* Fixed some instances of `RUNTIME_OUTPUT_DIRECTORY` that should have been `LIBRARY_OUTPUT_DIRECTORY`
* For non-xcode builds, altered `IMHEX_MAIN_OUTPUT_DIRECTORY` to be `${CMAKE_BINARY_DIR}/bin`, to separate build outputs from intermediate outputs. Not familiar with xcode, but it appears to already have had some form of separation
* Updated the build output location, of various components, to be relative to `IMHEX_MAIN_OUTPUT_DIRECTORY` - this co-locates the build outputs and allows the imhex binary to be run (and debugged) from that build location
* For non-xcode builds, at cmake configure, either symbolic link directories from the ImHex-Patterns repo clone (if `IMHEX_OFFLINE_BUILD`) or copy from the download of the patterns (if not `IMHEX_OFFLINE_BUILD`). This allows the patterns to be available to the build directory instance of imhex, and have any changes between them and the repo be mirrored (if applicable). Again, xcode appears to already have some form of 'copy to build directory' that isn't an install step

### DIFF
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -537,7 +537,7 @@ function(downloadImHexPatternsFiles dest)
             # Make the pattern files available to ImHex binaries started from the build directory
             if (NOT IMHEX_OFFLINE_BUILD)
                 file(COPY "${imhex_patterns_SOURCE_DIR}/${FOLDER}" DESTINATION "${IMHEX_MAIN_OUTPUT_DIRECTORY}" PATTERN "**/_schema.json" EXCLUDE)
-            else()
+            elseif(NOT EXISTS ${IMHEX_MAIN_OUTPUT_DIRECTORY}/${FOLDER})
                 # If it is an offline build, the imhex_patterns_SOURCE_DIR location is likely a git clone
                 # Create the directories as symbolic links, so that any changes get mirrored both ways
                 file(CREATE_LINK "${imhex_patterns_SOURCE_DIR}/${FOLDER}" "${IMHEX_MAIN_OUTPUT_DIRECTORY}/${FOLDER}" COPY_ON_ERROR SYMBOLIC)


cmake/build_helpers.cmake
```diff
@@ -387,7 +387,7 @@ function(configureProject)
         # Support Xcode's multi configuration paradigm by placing built artifacts into separate directories
         set(IMHEX_MAIN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/Configs/$<CONFIG>" PARENT_SCOPE)
     else()
-        set(IMHEX_MAIN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}" PARENT_SCOPE)
+        set(IMHEX_MAIN_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin" PARENT_SCOPE)
```
WerWolv, line 390: This should probably only be done on Linux as Windows and macOS use a different structure

cmake/build_helpers.cmake
```diff
@@ -526,8 +526,18 @@ function(downloadImHexPatternsFiles dest)
...
+        file(MAKE_DIRECTORY "${IMHEX_MAIN_OUTPUT_DIRECTORY}")
         set(PATTERNS_FOLDERS_TO_INSTALL constants encodings includes patterns magic nodes)
         foreach (FOLDER ${PATTERNS_FOLDERS_TO_INSTALL})
+            # Make the pattern files available to ImHex binaries started from the build directory
+            if (NOT IMHEX_OFFLINE_BUILD)
+                file(COPY "${imhex_patterns_SOURCE_DIR}/${FOLDER}" DESTINATION "${IMHEX_MAIN_OUTPUT_DIRECTORY}" PATTERN "**/_schema.json" EXCLUDE)
+            else()
+                # If it is an offline build, the imhex_patterns_SOURCE_DIR location is likely a git clone
+                # Create the directories as symbolic links, so that any changes get mirrored both ways
+                file(CREATE_LINK "${imhex_patterns_SOURCE_DIR}/${FOLDER}" "${IMHEX_MAIN_OUTPUT_DIRECTORY}/${FOLDER}" SYMBOLIC)
```
WerWolv, line 539: Is a symbolic link a good idea? Not all file systems support it.
I don't think the two-way mirroring is really necessary so a simple copy should suffice I think

WerWolv:
Thanks a lot!

cmake/build_helpers.cmake
BobSmun: I thought that Windows was set up to use MinGW (effectively GCC), which would also work with these changes. In terms of file layout (e.g. executable next to shared objects next to pattern files), I anticipated that they should be compatible though I didn't test the windows build yet.

I also didn't touch the XCODE output path, which I assumed was the macOS build method. I don't have a mac to test with, but it looked like it would correctly separate out the differently configured builds.

cmake/build_helpers.cmake
BobSmun: As I have been making modifications in the PatternLibrary too, I found it useful to edit in-place next to the binary. I mostly was intending to avoid changes there getting lost.

I notice that there is a `COPY_ON_ERROR` flag for `CREATE_LINK` that I seem to have missed, which should fall back to a copy if the symbolic link is not supported.

BobSmun:
Tested these changes under windows (using msys2)

* Re-added `RUNTIME_OUTPUT_DIRECTORY` where I replaced it with `LIBRARY_OUTPUT_DIRECTORY` (so both are now present), as the windows build seems to use that for shared libraries
* libpl had a specific override of the various OUTPUT_DIRECTORYs, only for windows, that was missed

BobSmun:
Whole bunch of work to fix macOS
Also fixed up unit tests (including previously disabled and broken plugin tests)
Added windows runner for unit tests, as there was one point where linux unit tests were passing when windows ones were failing
Link manually cloned ImHex-Patterns into cmake

PR #1891, opened 2024-09-05
Refactor Texteditor code to ImHex standards.

paxcut:
There is no code added or removed. All changes are to make Text editor code look more like the rest of ImHex.

- All member variable names that used a convention like in `mFieldVariable` were switched to the common ImHex convention of `m_fieldVariable`.
- All member function names that used pascal case were switched to the common ImHex convention of camel case.
- All int, unsigned int, unsigned long ... were switched to int32_t, uint32_t, uint64_t,... I used those because TextEditor has no access to the ImHex headers where the types are defined.
- Local variable names that were cryptic were given more descriptive ones.
- Other changes include a few fixes from cppCheck mostly for variable scope.

### DIFF
--- a/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
+++ b/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
@@ -51,104 +51,104 @@ public:
 
 	struct Breakpoint
 	{
-		int mLine;
-		bool mEnabled;
-		std::string mCondition;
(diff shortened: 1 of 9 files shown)


paxcut:
I tried to merge this locally and found that it is probably not a good idea to do a large refactor when there are lots of prs pending, so I'll draft this for now and wait until there has been a chance to merge the others

Issue #1895, opened 2024-09-08
[Feature] Add support zstd stream data to `hex::dec::zstd_decompress`


rockisch:
### What feature would you like to see?

Allow decompression of zstd data that was generated through streaming inside pattern scripts.

### How will this feature be useful to you and others?

It will expand the kinds of data 'zstd_decompress' can support.

### Request Type

- [X] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

### Additional context?

Even though this is a pattern script feature, internally 'zstd_decompress' calls https://github.com/WerWolv/ImHex/blob/04763eaeea0663d8deb1597a66603fa804c2ab89/plugins/decompress/source/content/pl_functions.cpp#L197 so the fix has to be made in the main repo.

PR #1898, opened 2024-09-10
impr: Add streaming decompression to zstd_decompress

rockisch:
### Problem description

https://github.com/WerWolv/ImHex/issues/1895

Added code that handles streamed zstd data. It is based around the [official documentation](http://facebook.github.io/zstd/zstd_manual.html) and the [example](https://github.com/facebook/zstd/blob/dev/examples/simple_decompression.c) provided at the main zstd repo.

The loop around the non-streamed version was also removed because I don't think it was doing anything (no `continue`s, `sourceSize` was always being set to 0).

### Additional things

To test, I generated streamed zstd data with this python script:

```py
import io; import pyzstd;
with open("data.zstd", "wb") as f:
    pyzstd.compress_stream(io.BytesIO(b'ab' * 100), f)
```

And then I ran this pattern script:
```
import std.mem;
import hex.dec;

u8 data[while(!std::mem::eof())] @ 0x00;
std::mem::Section data_sec = std::mem::create_section("data_sec");
hex::dec::zstd_decompress(data, data_sec);
```

Inspecting the section shows the correct data:
image

### DIFF
--- a/plugins/decompress/source/content/pl_functions.cpp
+++ b/plugins/decompress/source/content/pl_functions.cpp
@@ -211,25 +211,48 @@ namespace hex::plugin::decompress {
                 const u8* source = compressedData.data();
                 size_t sourceSize = compressedData.size();
 
-                do {
-                    size_t blockSize = ZSTD_getFrameContentSize(source, sourceSize);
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot!

PR #1901, opened 2024-09-13
feat: Extended bitmap visualizer to handle indexed colormaps.

paxcut:
### Problem description
Older image format use to store color values in a small lookup table so that the image could simply store an index to the table. The pattern language is not designed to handle this sort of operation which makes this extension necessary to be able to visualize images of this type.

The changes add an optional parameter to the bitmap visualizer which holds the color lookup table. If the image contains values that are outside the range of possible colors then the first color in the map is chosen. The dimensions of the image can be equal to or smaller than rows*columns depending on how the indices to the color map are stored. For example, you can use 4 bit indices which would make the image half the size (in bytes) but that limits the number of colors to 16. To store colors in sizes larger than one byte use an array of the appropriate sized integers.

### Screenshots

image


eliasdaler:
Tested locally - works great!
A file that took 7.5 seconds to process on my PC now takes just 0.25s, so huge performance gains for this particular case.

PR #1911, opened 2024-09-20
improv: Added multiline search/replace and fixed various crashes.

paxcut:
### Problem description
Previous implementation ignored everything after the first newline. Other changes include:
- Added isEmpty() function that checks if editor has no content.
- replaced asserts with default behavior to avoid unneeded crashes.
- fixed off by one error in DeleteRange()
- some crashes occurred because readily available corrective actions were not being taken. For example start/end being -1 in DeleteRange().
- At the heart of search/replace is the ability to translate from indices into the text string to line/column coordinates used for everything. To this end a function (StringIndexToCoordinates) was added to do the translation taking utf-8 chars into account. This made the recently added Utf8BytesToChars function unneeded, so it was removed.
- Removed commented out code that is not useful anymore. Also removed tooltip code which is also unused.
- Removed unused parameter wrapAround to FindNext().


lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -1084,18 +1063,17 @@ void TextEditor::Render() {
         mTopMargin = 0;
     }
 
-    static float linesAdded = 0;
-    static float pixelsAdded = 0;
-    static float savedScrollY = 0;
-    static float shiftedScrollY = 0;
+
     if (mTopMargin != oldTopMargin) {
+        static float savedScrollY = 0;
```
WerWolv, line 1068: Using `static` here will most likely not going to work correctly if there's multiple editors shown on-screen. Move these to be member variables of the TextEditor class

WerWolv (changes requested):

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
WerWolv: This is still problematic. Besides this, the PR looks good

paxcut: I posted an explanation as to why they are ok as static but now i dont see it anymore. I guess i don't really know how to use this user interface. so i'll try just this now.

paxcut: I  see if i press resolve conversation is erases the comment? thats weird. Anyway, the code in question is the one responsible for having popup menus behave like vscode so they can't be used by several editors at the same time.  Being static makes the code simpler as the variables are initialized once and then keep their values between calls. In a window use one may open the window as `Find`, expand it to `Replace`, shrink it back to `Find` and close the window. Through those event  the variables cycle through 4 values which are always the same so statics should be fine.

PR #1923, opened 2024-10-01
Fixes for breakpoints

paxcut:
WARNING: this PR won't compile unless [PR 132 from pattern language repository](https://github.com/WerWolv/PatternLanguage/pull/132)  is merged first. Some changes here are shared by at least another PR to this repository but there should not be any conflicts as the shared changes are identical.

### Problem description
fix: Editing patterns with breakpoints sets behaves unexpectedly. As a simple example, set a breakpoint and insert a blank line somewhere before the breakpoint location. The breakpoint will appear to move but in reality it hasn't. To see this set another breakpoint elsewhere in the file and the old one will be displayed where it is really located at.

The reason for this and many other problems with breakpoints is that currently ImHex keeps two set of breakpoints in text editor and in evaluator that are independent of each other, ie, changes to one don't affect the other. This PR aims at synchronizing the two sets through the per provider breakpoints that exist in view pattern editor.

It accomplishes this by making the text editor version of breakpoints the principal source of vectors and the ones in evaluator the effective version. The first allows one to modify the text around and at the breakpoint and notify others that the changes have induced changes in the breakpoint locations. The effective breakpoints allow the insertion and deletion of breakpoints.

View pattern editor is where breakpoints are updated. It receives notifications from text editor about changes and then makes sure the version in evaluator is updated with those changes. View pattern editor also manages breakpoint addition and deletion so before making changes it gets a copy of the current ones from text editor, sets the ones in evaluator, uses the evaluator functions to add or delete breakpoints and finally sets the text editor version with the new version.

### DIFF
--- a/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
+++ b/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
@@ -133,7 +133,7 @@ public:
 	using Keywords = std::unordered_set<std::string> ;
     using ErrorMarkers = std::map<Coordinates, std::pair<uint32_t ,std::string>>;
     using ErrorHoverBoxes = std::map<Coordinates, std::pair<ImVec2,ImVec2>>;
-    using Breakpoints = std::unordered_set<int32_t>;
+    using Breakpoints = std::unordered_set<uint32_t>;
     using Palette = std::array<ImU32, (uint32_t)PaletteIndex::Max>;
     using Char = uint8_t ;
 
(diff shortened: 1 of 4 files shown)


WerWolv:
Merging the other breakpoints PR seems to have caused a merge conflict here

paxcut:
very likely. I posted a comment about this problem in discord. I'll resolve them now.

paxcut:
done with the conflicts

PR #1966, opened 2024-11-23
Inline visualizers in custom data inspectors

paxcut:
### Problem description
This PR aims at making inline visualizers work on the data inspector so that more rgb encodings can be added as custom pattern language inspector rows. This was never setup to work because the inline visualizer rendering function was a private member function.

In order to be accessible from the inspector class the rendering function was made public.

### Additional things
Missing still is the tooltip to make it behave like other color entries.


plugins/builtin/source/content/views/view_data_inspector.cpp
```diff
@@ -196,8 +197,13 @@ namespace hex::plugin::builtin {
 
             try {
                 // Set up the display function using the pattern's formatter
-                auto displayFunction = [value = pattern->getFormattedValue()] {
-                    ImGui::TextUnformatted(value.c_str());
+                auto displayFunction = [pattern,value = pattern->getFormattedValue()] {
+                    auto drawer = std::make_unique<hex::ui::PatternDrawer>();
```
WerWolv, line 201: This causes a lot of unnecessary allocations every single frame. Instead, I believe the PatternDrawer shouldn't be involved at all in rendering this

WerWolv (changes requested):

plugins/ui/source/ui/pattern_drawer.cpp
```diff
@@ -384,11 +358,12 @@ namespace hex::ui {
             if (ImGui::Button(hex::format(" {}  {}", ICON_VS_EYE_WATCH, value).c_str(), ImVec2(width, ImGui::GetTextLineHeight()))) {
                 auto previousPattern = m_currVisualizedPattern;
                 m_currVisualizedPattern = &pattern;
-
+                m_lastVisualizerError = m_visualizerDrawer.getLastVisualizerError();
```
WerWolv, line 361: It doesn't look like `m_lastVisualizerError` is really needed anymore. Keeping the error message only stored in the visualizer drawer would be better

WerWolv (changes requested):

plugins/ui/source/ui/pattern_drawer.cpp
paxcut: I deleted the variable as requested and added function to clear the error on VisualizerDrawer to make it easier. I simulated a code error in my Inspector rgba5551 visualizer and I got a nice popup error of the reason why it wasn't working. Fixing the error made the visualizer work again so it looks like at a basic level it is working.

PR #1969, opened 2024-11-25
fix: 3d visualizer window is too big.

paxcut:
### Problem description
Long error messages were forcing the width of the window to span the entire screen. 

The fix was sending the long message to the log and outputting a short message to the 3d visualizer window.


WerWolv:
Thanks! Instead of shortening all the error messages though, would it make sense to maybe just wrap the error message so it doesn't overflow anymore using `ImGui::TextWrapped`? That way we can convey the full information on more complex errors while not making it look terrible

paxcut:
There are two kinds of error that the 3d visualizer may want to convey when the inputs are incorrect. One type is a short message describing what data is wrong, something like, bad indices and colors found. The other which can be potentially huge is a list of all the values that are wrong so that if there are only a few it may help the user find out whats wrong. If we truncate the message then it is not helpful.

WerWolv:
Where are we truncating them? If they get so big they go off-screen at the bottom we should make the error popup scrollable instead

paxcut:
Yes sorry, I misunderstood what you meant by wrapping. That is indeed a better solution which I'll look into implementing.

paxcut:
I added the requested code to display the messages with a scroll bar. Please review at your earliest convenience.

plugins/visualizers/source/content/pl_visualizers/3d_model.cpp
```diff
@@ -605,6 +618,20 @@ namespace hex::plugin::visualizers {
                 if (ImGuiExt::InputFilePicker("hex.visualizers.pl_visualizer.3d.texture_file"_lang, s_texturePath, {}))
                     s_shouldUpdateTexture = true;
             }
+
+            if (!s_errorMessage.empty()) {
```
WerWolv, line 622: This code will never be executed because one of the `throw` statements further up will be executed before getting here.

Also I believe instead of trying to handle the error message here separately again with a new system, why not just use the existing one and make the error message there wrapped / scrollable? Specifically in visualizer_drawer.cpp:31

WerWolv (changes requested):

plugins/visualizers/source/content/pl_visualizers/3d_model.cpp
paxcut: The throws prevent the rendering code from trying to use bad data that could crash ImHex. The throws do not stop the ui from being refreshed so that you can continue to interact with it. The error message is being added as part of the ui at the bottom of the window with the correct width and a scroll bar only if necessary. I have tested this setup with bad data designed specifically to display certain errors and it has always worked. If you fix the problem the error message disappears on the next run but will continue to be shown when you reopen the visualizer if you have made no changes.

The visualizer drawer is used only for inline visualizers which have no windows so the errors are displayed as popups. 
Using a child that contains one widget can hardly be considered a `system` though. There is nothing fancy being done except making the text red to make it stand out as an error. As opposed to the inline visualizers I my be better to show the error while we interact with  the ui.

WerWolv: > The throws prevent the rendering code from trying to use bad data that could crash ImHex.

The throws are fine. All the other visualizers use these too. But they throw a std::runtime_error together with a message and this message is then being rendered by the visualizer drawer in the visualizer popup.

> The throws do not stop the ui from being refreshed so that you can continue to interact with it.

Okay, it stops the execution the first frame and then on the next frame skips the code that could throw exceptions and then instead renders the error. That's however already all being handled by the visualizer drawer. There's no need to reimplement that part again.

> If you fix the problem the error message disappears on the next run but will continue to be shown when you reopen the visualizer if you have made no changes.

That's how the current system works already by you simply throwing the error. 

> The visualizer drawer is used only for inline visualizers which have no windows so the errors are displayed as popups.

No, the Visualizer Drawer is being used for both inline visualizers and regular visualizers. With inline visualizers, the error will be rendered inside of the value column but with regular visualizers, it's being put inside of the popup.

WerWolv: The rest of the changes are good btw, I'm really only talking about the error message rendering here

paxcut: Now that is making more sense to me and absolutely what you suggest is the way to go. I should be a simple change too (at least compared to learning how to add the scroll bar) so it should be ready soon.

paxcut: OK, I went ahead with the proposed changes and I think it should work. I say this because generating error messages from visualizers in general is not easy, or at least I wasn't able to do it for any visualizer other than 3D. I changed the code a bit to reflect the fact that the widths of the windows affected vary, so it should work regardless of their size. I tested that because 3d visualizer is resizable and if window is made to become wide enough then no scrollbar is rendered. An important test is to reduce the width of the window and make sure the error message changes accordingly which it does, albeit sluggishly due to the small difference in region available and content size.

WerWolv (approved):

PR #1973, opened 2024-11-29
Turned static variables into member variable 

paxcut:
to avoid potential problems in the future if opening  several popup windows at the same time becomes possible.


WerWolv:
Thanks!

PR #1979, opened 2024-12-02
Replaced asserts from TextEditor.cpp

paxcut:
All asserts in TextEditor.cpp were replaced with IM_ASSERTs so that if they are hit ImHex won't crash and give a nice popup message instead.


WerWolv:
Thank you!

PR #1981, opened 2024-12-03
Add constant data input to buffer node

jumanji144:
Added a hex data input field to the buffer node in the data processor.

image

### DIFF
--- a/plugins/builtin/source/content/data_processor_nodes/basic_nodes.cpp
+++ b/plugins/builtin/source/content/data_processor_nodes/basic_nodes.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <hex/api/content_registry.hpp>
 #include <hex/api/localization_manager.hpp>
 #include <hex/helpers/utils.hpp>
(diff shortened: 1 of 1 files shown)


plugins/builtin/source/content/data_processor_nodes/basic_nodes.cpp
```diff
@@ -33,9 +34,38 @@ namespace hex::plugin::builtin {
             ImGui::PushItemWidth(100_scaled);
             ImGui::InputScalar("hex.builtin.nodes.constants.buffer.size"_lang, ImGuiDataType_U32, &m_size, &StepSize, &FastStepSize);
             ImGui::PopItemWidth();
+
+            ImGui::InputTextMultiline("##buffer", m_constantString, ImVec2(150_scaled, 0), ImGuiInputTextFlags_AllowTabInput | ImGuiInputTextFlags_CharsHexadecimal);
+        }
+
+        /// Adapted from PatternLanguageBot
+        std::vector<u8> parseByteString(const std::string &string) {
+            if (string.empty()) return {};
```
C3pa, line 43: It looks like that in this project it's more common to write single-line if code block on the next line:
```c++
if (string.empty())
    return {};
```

jumanji144: Addressed

PR #1982, opened 2024-12-03
fix: HexII visualisation does not allow hex characters in the edit field

alexkar598:
### Problem description
At the moment, attempting to edit files using HexII data visualization is impossible as the editing fields only allow numeric characters but is interpreted as a hex value

I yoinked the params used for the hex visualization edit field. 

### Screenshots
image

### Additional things
Ideally this should instead accept the format used by HexII, but I have what scientists describe as a skill issue and this at least makes it possible to use HexII without switching back to hex for editing.


WerWolv:
Thank you!
That input field should probably be a `ImGui::InputText` instead with a custom parser to parse both Hex values as well as the ASCII values. As an intermediate step to make editing actually possible this is fine though

PR #1983, opened 2024-12-03
fix: context menus entries greyed out.

paxcut:
Some context menu entries that were available as shortcuts were greyed out. This PR aims to fix them and improve how context menus work for the text editor and the console. The improvements include:
- automatic focus on right click
- automatic selection on right click. If selected text is right-clicked then copy, cut and find will use the selection, if no selection is clicked but there is text were right-clicked, then the word will be selected and used. If right-clicking empty space copy and cut will be greyed out and find will start empty.
- similar functionality now exists for the console as well except the menu has fewer options due to it being read-only.
- added esc to close console context menu

### DIFF
--- a/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
+++ b/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
@@ -244,7 +244,8 @@ public:
 
 	Coordinates GetCursorPosition() const { return GetActualCursorCoordinates(); }
 	void SetCursorPosition(const Coordinates& aPosition);
-
+    bool raiseContextMenu() { return mRaiseContextMenu; }
+    void clearRaiseContextMenu() { mRaiseContextMenu = false; }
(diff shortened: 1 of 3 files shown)


lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
```diff
@@ -244,7 +244,8 @@ class TextEditor
 
 	Coordinates GetCursorPosition() const { return GetActualCursorCoordinates(); }
 	void SetCursorPosition(const Coordinates& aPosition);
-
+    bool raiseContextMenu() { return mRaiseContextMenu; }
```
WerWolv, line 247: This should probably be in UpperCamelCase to match the rest of the functions in the Text Editor

paxcut: Yes that one and the other one added. I am itching to redo the refactor PR to make text editor like the rest of ImHex in camel and snake cases and the rest of the formatting stuff. I changed those two back for now. Thanks for noticing.

PR #1984, opened 2024-12-04
feat: added goto line popup to text editors.

paxcut:
Since the popup is fairly small I opted for a straight addition parallel to the find/replace. To make code more clear the functions that create each popup were coalesced and made their interface simpler. That forced a reorganization of the data processing which translates to a larger number of changes than usual. Most of those changes are just moving some action from one function to another. 

The old method to identify popups using the size and position of the window was dropped in favor of one based on child windows and using their names for a much easier and robust identification. 

Added specialized functions to text editor to jump to a line or to given coordinates with a simple interface that simplifies older code that performed the same task.

Because this PR modifies heavily the same code as the previous PR (1983) it is also included here to make merging easier.


lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -1086,24 +1086,20 @@ void TextEditor::Render() {
...
-                    mTopMargin = mFindReplaceHandler.GetFindWindowSize().y;
-                }
-            }
-        }
-    } else {
+    if (g.NavWindow != nullptr) {
+        auto window = g.NavWindow;
+        std::string windowName = window->Name;
+        if (windowName.find("pattern_editor") != std::string::npos && (windowName.find("find_replace") != std::string::npos || windowName.find("goto_line") != std::string::npos)) {
+            std::string parentName = windowName.substr(0, windowName.find("/##pattern_editor"));
```
WerWolv, line 1096: This window name comes from ImHex and shouldn't be hardcoded

plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -231,98 +231,82 @@ namespace hex::plugin::builtin {
...
+            m_openGotoLinePopUp = false;
+            // Place the popup at the top right of the window
+            auto style = ImGui::GetStyle();
 
-            ImGui::SetNextWindowPos(windowPosForPopup);
-            ImGui::SetNextWindowSize(popupSize);
-            ImGui::OpenPopup("##pattern_editor_find_replace");
 
-            findReplaceHandler->resetMatches();
+            auto labelSize = ImGui::CalcTextSize("Goto Line: ");
```
WerWolv, line 281: This should probably use a localized string, otherwise it will only work in english

plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -950,23 +1012,36 @@ namespace hex::plugin::builtin {
...
             const auto linesToAdd = m_console->size() - lineCount;
+
+
             for (size_t i = 0; i < linesToAdd; i += 1) {
+                if (!skipNewLine)
+                    m_consoleEditor.InsertText("\n");
+                skipNewLine = false;
                 m_consoleEditor.InsertText(m_console->at(lineCount + i));
-                m_consoleEditor.InsertText("\n");
+               // if (i + 1 < linesToAdd)
```
WerWolv, line 1043: Are these commented out lines still needed?

WerWolv (changes requested):

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: Maybe you are thinking that the child window is the popup and that the parent the window created by Imhex by the window creating system in place but that's not what they are. I create popup windows and create children in them to help find them here without having to use positions and sizes. The parent here is actually the popup window.

WerWolv: I'm talking about the strings `pattern_editor` and `/##pattern_editor`. These should be called something. I'd use something like `text_editor` so it's not related to the pattern editor or ImHex anymore.

I really think the Text Editor should be 100% free of any ImHex-specific functions or terminology. It's a standalone thing

plugins/builtin/source/content/views/view_pattern_editor.cpp
paxcut: Maybe, but I doubt it. the calculation is just an approximation to make sure the window is not too big. working with localized strings should be easier. I'll see if I can figure how to calculate the size in pixels the localized string will have.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: Oh no! this thing again where i totally misunderstand the comment. I can change the name but those names are created by the view pattern editor class, not the text editor so it has the name that comes from its own name to help identify its origin to somebody who doesn't already know. Also note that there are text editor windows but there are also console editor windows. 

While I agree that the text editor is not part of ImHex there must be a limit on how costly it is to maintain the isolationism in both time and clarity of the final product. The names of `pattern editor` and `console editor` do belong to imhex, but imhex isn't likely to be using them for something else and when a version of the editor that is not for imhex is written, it won't contain the view pattern editor code so the names won't be used there.

plugins/builtin/source/content/views/view_pattern_editor.cpp
paxcut: > Are these commented out lines still needed?

No, not needed. I have removed them.

plugins/builtin/source/content/views/view_pattern_editor.cpp
paxcut: OK, it was actually quite painless because there was already a localized string with the correct value so I switched it to use that.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: Now all the windows that belong to the pattern editor can be found searching for "text editor" in their names. Thankfully, the console editor is not involved here or else choosing names would not have  been as easy as it was.

WerWolv: No you misunderstood me still. You cannot make the Text Editor depend on windows that are being created by ImHex. If you took the Text Editor and used it elsewhere, in another project or even just in another view in ImHex, these windows won't be there anymore and that portion of the Text Editor will just not work at all anymore

paxcut: That is sort of the point. When the text editor runs this part of the code there maybe windows detected here that need to be ignored, like the context menu for the pattern editor. The purpose of this code is to calculate how much space the popup is hiding from the text so that the offsets and the scroll can be adjusted to make all the text visible again. It is impossible to make that work with arbitrary windows, like if some window covers the text editor completely. The windows detected here are designed in a way that makes it possible to counteract their negative visibility effect om the text editor.

paxcut:
I made the changes requested. Please review again at your earliest convenience.

WerWolv (approved):

WerWolv:
Thanks for the changes!

Issue #1846, opened 2024-08-13
[Bug] "Edit the hex" achievement doesn't unlock


BobSmun:
### Operating System

Windows

### What's the issue you encountered?

Modifying a byte by double-clicking on it and then entering the new value does not unlock the "Edit the hex" achievement

### How can the issue be reproduced?

* open any file
* ensure that the prerequisite achievement ("You and you and you") has been completed, but "Edit the hex" has not 
* double-click on a byte in the 'Hex Editor' view
* type a different value, such that the byte shows red as modified
* single-click elsewhere in the 'Hex Editor' view to finalize modifying

Alternatively, use enter (_not_ num-pad enter) to start or finish the edit

### ImHex Version

1.35.4

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

### Additional context?

In the code, it looks like `EventPatchCreated` is subscribed to, to unlock the achievement, but a search does not show the event being posted / raised anywhere

paxcut:
I can confirm issue on dev branch 1.36.0-wip as well, but I think `EventPatchCreated ` may be for the ROM hacks achievement. If so and it is not posted maybe that one doesn't work either.

BobSmun:
`EventPatchCreated` certainly sounds like it would be intended for the 'ROM hacks' achievement, but it unlocks "hex.builtin.achievement.hex_editor.modify_byte.name", whereas the `exportIPSPatch` and `exportIPS32Patch` functions call `unlockAchievement` (with "hex.builtin.achievement.hex_editor.create_patch.name") directly

iTrooz:
Fixed by #1985

PR #1985, opened 2024-12-04
Various achievements fixes (Edit the Hex, ROM Hacks)

BioTheWolff:
### Problem description
As described in #1846:
- the `Edit the Hex` achievement doesn't unlock when it should
- the `ROM Hacks` achievement is not using event-driven architecture (the functions call `unlockAchievement` themselves)

Firstly, for the `Edit the Hex` achievement:
- replaced the old event listener on `EventPatchCreated` with a listener on `EventProviderDataModified`, which picks up bytes changes
- ensured the provider data change comes from a File provider, else unlocking the achievement wouldn't make sense
- *Note*: a discovered side effect is that the "Fill" function modifies the provider byte per byte (with a for loop)
  - there is no use in testing the size of the data change, as it is always 1 byte
  - the Fill function could probably be reworked to fill in whole regions at a time?

About the `ROM Hacks` achievement:
- implemented the new, still unused `EventPatchCreated` event.
  - signal signature is `const unsigned char *, u64, const IPSKind`: buffer pointer, buffer size, and IPS kind (IPS/IPS32)
- make use of the `::post` and `::subscribe` methods on said event to unlock the achievement
- **WARNING::behaviour change**: the event's `post` signal has been moved in the success branch of the IPS generation condition, meaning that achievement will only unlock if IPS patch export has worked. I felt it would make more sense than unlocking an achievement on an error, if there was any to raise.

### DIFF
--- a/lib/libimhex/include/hex/api/event_manager.hpp
+++ b/lib/libimhex/include/hex/api/event_manager.hpp
@@ -259,7 +259,7 @@ namespace hex {
      * @brief Called upon creation of an IPS patch.
      * As for now, the event only serves a purpose for the achievement unlock.
      */
-    EVENT_DEF(EventPatchCreated, const unsigned char *, u64, const IPSKind);
+    EVENT_DEF(EventPatchCreated, const u8*, u64, const PatchKind);
     EVENT_DEF(EventPatternEvaluating);
     EVENT_DEF(EventPatternExecuted, const std::string&);
     EVENT_DEF(EventPatternEditorChanged, const std::string&);
--- a/lib/libimhex/include/hex/api/event_manager.hpp
+++ b/lib/libimhex/include/hex/api/event_manager.hpp
@@ -11,6 +11,7 @@
 
 #include <hex/api/imhex_api.hpp>
 #include <hex/helpers/logger.hpp>
+#include <hex/helpers/patches.hpp>
 
 #include <wolv/types/type_name.hpp>
 
@@ -253,7 +254,12 @@ namespace hex {
     EVENT_DEF(EventWindowInitialized);
     EVENT_DEF(EventWindowDeinitializing, GLFWwindow *);
     EVENT_DEF(EventBookmarkCreated, ImHexApi::Bookmarks::Entry&);
-    EVENT_DEF(EventPatchCreated, u64, u8, u8);
+
+    /**
+     * @brief Called upon creation of an IPS patch.
+     * As for now, the event only serves a purpose for the achievement unlock.
+     */
+    EVENT_DEF(EventPatchCreated, const unsigned char *, u64, const IPSKind);
     EVENT_DEF(EventPatternEvaluating);
     EVENT_DEF(EventPatternExecuted, const std::string&);
     EVENT_DEF(EventPatternEditorChanged, const std::string&);
--- a/lib/libimhex/include/hex/helpers/patches.hpp
+++ b/lib/libimhex/include/hex/helpers/patches.hpp
@@ -21,6 +21,11 @@ namespace hex {
         MissingEOF
     };
 
+    enum class IPSKind {
+        IPS,
+        IPS32
+    };
+
     class Patches {
     public:
         Patches() = default;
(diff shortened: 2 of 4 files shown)


plugins/builtin/source/content/achievements.cpp
```diff
@@ -192,8 +194,15 @@ namespace hex::plugin::builtin {
                 AchievementManager::unlockAchievement("hex.builtin.achievement.hex_editor", "hex.builtin.achievement.hex_editor.create_bookmark.name");
             });
 
-            EventPatchCreated::subscribe([](u64, u8, u8) {
-                AchievementManager::unlockAchievement("hex.builtin.achievement.hex_editor", "hex.builtin.achievement.hex_editor.modify_byte.name");
+            EventProviderDataModified::subscribe([](const prv::Provider * prv, unsigned long, const unsigned long, const unsigned char *) {
+                // Warning: overlaps with the "Flood fill" achievement, since "Fill" works by writing to bytes one-by-one.
+                    // Thus, we do not check for size, that will always be equal to 1 even during a fill operation.
+                if (prv->getTypeName() == "hex.builtin.provider.file")
```
WerWolv, line 200: Should we really lock this achievement to only work with files? Like it's possible to modify the bytes in other providers like the disk provider or the gdb provider. The achievement description would still make sense there since it only talks about modifying the value of a byte

plugins/builtin/source/content/achievements.cpp
```diff
@@ -192,8 +194,15 @@ namespace hex::plugin::builtin {
                 AchievementManager::unlockAchievement("hex.builtin.achievement.hex_editor", "hex.builtin.achievement.hex_editor.create_bookmark.name");
             });
 
-            EventPatchCreated::subscribe([](u64, u8, u8) {
-                AchievementManager::unlockAchievement("hex.builtin.achievement.hex_editor", "hex.builtin.achievement.hex_editor.modify_byte.name");
+            EventProviderDataModified::subscribe([](const prv::Provider * prv, unsigned long, const unsigned long, const unsigned char *) {
```
WerWolv, line 197: Use `u64` instead of `unsigned long` and so on

lib/libimhex/include/hex/helpers/patches.hpp
```diff
@@ -21,6 +21,11 @@ namespace hex {
         MissingEOF
     };
 
+    enum class IPSKind {
```
WerWolv, line 24: I'd probably name this `PatchKind` since there might be other types of patches in the future

lib/libimhex/include/hex/api/event_manager.hpp
```diff
@@ -253,7 +254,12 @@ namespace hex {
     EVENT_DEF(EventWindowInitialized);
     EVENT_DEF(EventWindowDeinitializing, GLFWwindow *);
     EVENT_DEF(EventBookmarkCreated, ImHexApi::Bookmarks::Entry&);
-    EVENT_DEF(EventPatchCreated, u64, u8, u8);
+
+    /**
+     * @brief Called upon creation of an IPS patch.
+     * As for now, the event only serves a purpose for the achievement unlock.
+     */
+    EVENT_DEF(EventPatchCreated, const unsigned char *, u64, const IPSKind);
```
WerWolv, line 262: `const u8*`

WerWolv (changes requested):

plugins/builtin/source/content/achievements.cpp
BioTheWolff: Fair enough, I have removed the condition so we can broaden achievement unlocking.

BioTheWolff:
You're welcome :)

The fill command issue should be addressed, yeah.

Also, while working on this PR, I noticed a big point of contention in the form of the file `event_manager.hpp`.
Changing anything in it requires the rebuilding of 147 files, which poses a problem when working on large PRs regarding events.
Should this be addressed too?

WerWolv:
I actually did address that issue in Fenestra. There I separated the event manager from the event definitions and also split the events up into a few different files based on their type. It would make sense to do the same thing for ImHex as well yeah

BioTheWolff:
@WerWolv: Do you need anything else changed, or is it good to merge?

WerWolv:
No, looks good to me now. Thanks!

PR #1986, opened 2024-12-04
Fix AES ECB mode in Data processor

jumanji144:
Fix the AES ECB mode in the data processor along with some other misc fixes:
- Fixed nullpointer node not working
- Fixed crypto module incorrectly using mbedtls api
- Fixed crypto module ignoring mbedtls errors
- Fixed silently ignoring of errors in AES node

### DIFF
--- a/lib/libimhex/source/helpers/crypto.cpp
+++ b/lib/libimhex/source/helpers/crypto.cpp
@@ -1,8 +1,10 @@
+#include <algorithm>
 #include <hex/helpers/crypto.hpp>
 
 #include <hex/providers/provider.hpp>
 
 #include <wolv/utils/guards.hpp>
+#include <wolv/utils/expected.hpp>
 
 #include <mbedtls/version.h>
 #include <mbedtls/base64.h>
@@ -496,8 +498,8 @@ namespace hex::crypt {
         return encodeLeb128<i128>(value);
     }
 
-    static std::vector<u8> aes(mbedtls_cipher_type_t type, mbedtls_operation_t operation, const std::vector<u8> &key, std::array<u8, 8> nonce, std::array<u8, 8> iv, const std::vector<u8> &input) {
-        std::vector<u8> output;
+    static wolv::util::Expected<std::vector<u8>, int> aes(mbedtls_cipher_type_t type, mbedtls_operation_t operation, const std::vector<u8> &key,
+                   std::array<u8, 8> nonce, std::array<u8, 8> iv, const std::span<const u8> &input) {
 
         if (input.empty())
             return {};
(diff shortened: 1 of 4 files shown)


plugins/builtin/source/content/data_processor_nodes/decode_nodes.cpp
```diff
@@ -25,6 +29,9 @@ namespace hex::plugin::builtin {
         }
 
         void process() override {
+            const auto &mode = static_cast<crypt::AESMode>(m_mode);
```
WerWolv, line 32: I don't think this should be a reference. (How does that even compile??)

lib/libimhex/source/helpers/crypto.cpp
```diff
@@ -507,38 +509,65 @@ namespace hex::crypt {
...
 
         std::array<u8, 16> nonceCounter = { 0 };
-        std::copy(nonce.begin(), nonce.end(), nonceCounter.begin());
-        std::copy(iv.begin(), iv.end(), nonceCounter.begin() + 8);
+
+        auto mode = mbedtls_cipher_get_cipher_mode(&ctx);
+
+        // if we are in ECB mode, we don't need to set the nonce
+        if (mode != MBEDTLS_MODE_ECB) {
+            std::ranges::copy(nonce, nonceCounter.begin());
```
WerWolv, line 529: Where does the name `nonceCounter` come from? I feel like it should be called `IV` or something instead

WerWolv (changes requested):

plugins/builtin/source/content/data_processor_nodes/decode_nodes.cpp
jumanji144: I think i was just like 'make everything const reference' but i think this works because it converts the lvalue to a rvalue to that lvalue, since its just a static cast to that lvalue. might also be UB

lib/libimhex/source/helpers/crypto.cpp
jumanji144: nonceCounter was originally named by you, as shown by the blame, i didn't want to change the naming as i thought it was correctly named by you. I will address it.

WerWolv: My fault then, sorry

plugins/builtin/source/content/data_processor_nodes/decode_nodes.cpp
WerWolv: Yeah I believe `static_cast` will return a prvalue which will expire right after that line finishes so you have a dangling reference there. It compiles because rvalues can bind to const lvalue references but assignments like this don't cause lifetime extensions so it's still a bug

Issue #1781, opened 2024-06-30
[Feature] Export disassembler output to file


rupurt:
### What feature would you like to see?

When I disassemble a binary
Then I would like to export the assembly instructions to a file
So that I can view and edit it in vim

### How will this feature be useful to you and others?

1. I only need to run the disassembler once
2. I can view and edit it in other development tools

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

shazz:
my friendly vote for this one! The embedded disassembler is really nice but impossible to export or just copy/paste the result.

Thanks!

jmamma:
+1

iTrooz:
Fixed by #1987

PR #1987, opened 2024-12-04
Export disassembler results to ASM file

BioTheWolff:
### Problem description

This PR implements the feature request #1781, that suggests adding a button to export disassembled instructions into an ASM file.

This adds a button to export the current disassembled instructions to an ASM file. Said file is suffixed by an `.asm` extension if not specified at file creation.

*Note: the file is written to for every `Disassembly` item in the vector, as it was the easiest and most memory-conservative way of doing it.*

The file creation task is implemented based on IPS patch exports, so it fits the same pattern.
A `ToastError` is raised when the ASM export could not complete successfully.

Translations have been implemented for both `en_US` and `de_DE` for the two new keys:
- `hex.disassembler.view.disassembler.export`: file export button
- `hex.disassembler.view.disassembler.export.popup.error`: error popup text

### Screenshots
The button is disabled when the disassembler is working, or when the disassembly vector is empty.

Here is a complete breakdown of the visual changes:
image

### Additional things
As expected, the exporter writes every item's `mnemonic` and `operators` to the file, producing an output like this:

`example.asm`
```asm
.byte 0x7f, 0x45, 0x4c, 0x46
andeq r0, r1, r2, lsl #2
andeq r0, r0, r0
andeq r0, r0, r0
eorseq r0, lr, r3
andeq r0, r0, r1
andeq r1, r0, r0, asr #32
andeq r0, r0, r0
andeq r0, r0, r0, asr #32
```

### DIFF
--- a/plugins/disassembler/source/content/views/view_disassembler.cpp
+++ b/plugins/disassembler/source/content/views/view_disassembler.cpp
@@ -99,7 +99,7 @@ namespace hex::plugin::disasm {
                 fs::openFileBrowser(fs::DialogMode::Save, {}, [this](const std::fs::path &path) {
                     auto p = path;
                     if (p.extension() != ".asm")
-                        p.replace_filename(std::format("{}{}", p.filename().string(), ".asm"));
+                        p.replace_filename(hex::format("{}{}", p.filename().string(), ".asm"));
                     auto file = wolv::io::File(p, wolv::io::File::Mode::Create);
 
                     if (!file.isValid()) {
@@ -108,8 +108,8 @@ namespace hex::plugin::disasm {
                     }
 
                     // as disassembly code can be quite long, we prefer writing each disassembled instruction to file
-                    for (Disassembly d : m_disassembly) {
-                        file.writeString(std::format("{} {}\n", d.mnemonic, d.operators));
+                    for (const Disassembly& d : m_disassembly) {
+                        file.writeString(hex::format("{} {}\n", d.mnemonic, d.operators));
                     }
                 });
             });
--- a/plugins/disassembler/source/content/views/view_disassembler.cpp
+++ b/plugins/disassembler/source/content/views/view_disassembler.cpp
@@ -107,9 +107,16 @@ namespace hex::plugin::disasm {
                         return;
                     }
 
-                    // as disassembly code can be quite long, we prefer writing each disassembled instruction to file
+                    // As disassembly code can be quite long, we prefer writing each disassembled instruction to file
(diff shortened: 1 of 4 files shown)


iTrooz (approved):

plugins/disassembler/source/content/views/view_disassembler.cpp
```diff
@@ -92,6 +93,29 @@ namespace hex::plugin::disasm {
...
+                    auto file = wolv::io::File(p, wolv::io::File::Mode::Create);
+
+                    if (!file.isValid()) {
+                        ui::ToastError::open("hex.disassembler.view.disassembler.export.popup.error"_lang);
+                        return;
+                    }
+
+                    // as disassembly code can be quite long, we prefer writing each disassembled instruction to file
+                    for (const Disassembly& d : m_disassembly) {
+                        file.writeString(hex::format("{} {}\n", d.mnemonic, d.operators));
```
WerWolv, line 112: This might lead to additional white space at the end of lines if the instruction doesn't have any operators. Not sure if this matters for any assemblers but I figured I'd mention it

plugins/disassembler/source/content/views/view_disassembler.cpp
```diff
@@ -92,6 +93,29 @@ namespace hex::plugin::disasm {
...
+                    if (p.extension() != ".asm")
+                        p.replace_filename(hex::format("{}{}", p.filename().string(), ".asm"));
+                    auto file = wolv::io::File(p, wolv::io::File::Mode::Create);
+
+                    if (!file.isValid()) {
+                        ui::ToastError::open("hex.disassembler.view.disassembler.export.popup.error"_lang);
+                        return;
+                    }
+
+                    // as disassembly code can be quite long, we prefer writing each disassembled instruction to file
```
WerWolv, line 110: As a note for future PRs, comments should generally be capitalized at the beginning

WerWolv:
Thank you very much! That's definitely a feature that has been requested quite a lot over the years

plugins/disassembler/source/content/views/view_disassembler.cpp
BioTheWolff: You're right. I have fixed this by adding a simple condition to check if the `operators` string is empty before writing to file.
I have also added a condition to completely ignore the instruction if the `mnemonic` was somehow empty. This should never happen anyways. Please let me know if I should remove said condition.

WerWolv (approved):

PR #1992, opened 2024-12-08
Fix crash on UNDO/REDO shortcut

BohdanQQ:
Fix crash on UNDO/REDO shortcut press when in "title screen"

### Problem description
ImHex crashes when (by default CTRL + Z/Y) undo/redo is pressed when on "title"/"starting" screen (no file open, tested on Windows release build and Linux [WSL] from-source build).

This is due to the shortcut's callback being called even if the `provider` is `nullptr`. (see `createEditMenu` function).
Theoretically, this is prevented by the `enabledCallback` function passsed to `addMenuItem`. In this case, though, 
`addMenuItem` correctly propagates `enabledCallback` to menu item creation but does not pass `enabledCallback` to
shortcut creation. Thus, when handling shortcuts, `enabledCallback` is not used at all and the shortcut's callback 
can be called in contradiction with its preconditions. (specified by `enabledCallback`)

The implementation  wraps the callback in a check that decides whether the shortcut is enabled or not. 

(see changed files)
```c++
        auto callbackIfEnabled  = [enabledCallback, function]{ if (enabledCallback()) { function(); } };
```

This function is then passed along instead of the `function` (shortcut's callback).

Alternatively, we can check for `nullptr` in the callback directly. This would require modification of `createEditMenu`'s contents.
(I did not choose this implementation because I do not think it addresses the root of the issue).

### Screenshots
None

### Additional things
I'm not sure how big of a deal it is but I am unsure whether I can capture (`[enabledCallback, function]`) by reference or not.


WerWolv:
Thanks a lot!

Issue #1904, opened 2024-09-16
[Feature] Pasting text in hex editor


EGNL:
### What feature would you like to see?

I would like to see being able to paste text in files with ImHex's hex editor. Currently, it is only possible to paste bytes into files.

### How will this feature be useful to you and others?

For certain people, this feature removes the extra step of converting your text to bytes.

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

BioTheWolff:
I have looked into it and found a few ways to implement it.

One way would be to implement a different kind of paste (bytes or text) depending on the type of region you are focusing on (bytes, or their representation in the right column). However, I found two flaws to that:
- Currently, the selected region event does not provide information of whether you are selecting bytes or their representation (also, both selections are directly tied to one another);
- This would have to be described through documentation, because there is little to no room in the UI to properly explain the difference in pasting behaviour.

Another way would be to calculate (which is to say, literally guess) whether the clipboard contains plaintext or an ImHex-formatted array of bytes.
This seems to me like a bad idea, because there are many ways to interpret a given string, and the function could fail in delivering the expected result, without any way for the user to override this behaviour.

Finally, we could create a new, specialized keybind in the likes of "Paste all as text" (that could be tied to `Ctrl + Alt + Shift + V`).
It would allow users to voluntarily paste as plaintext, which would cause ImHex to transform text to its bytes array equivalent before writing to the region.
This also solves many edge cases, i.e. copying bytes and pasting them as text that could work flawlessly as it's an explicit user-requested action.

@WerWolv What are your thoughts on this?

paxcut:
Why not use the mouse location to decide where you are pasting to. If the mouse is over the column then paste as ASCII. If the mouse is over the hex values then paste as binary. This would be similar to how Imhex is able to edit the data as ASCII (by selecting an offset on the ASCII column) or as binary (by selecting an offset of the hex editor proper).

iTrooz:
Assignee: @BioTheWolff

PR #1998, opened 2024-12-10
Add capability to paste as string literal in files

BioTheWolff:
### Problem description
As suggested in #1904, ImHex could benefit from having the capability of pasting text directly into the editor.
This also complements the "Copy as... ASCII string" capability already implemented in the software.

A new shortcut called `Paste all as string` (to resemble the naming of the `ASCII string` copy option) now allows to paste plaintext directly.
This shortcut is mapped to the `CTRL + ALT + SHIFT + V` keybind.

Internally, a new flag called `asPlainText` has been added to the `pasteBytes` function, to minimise code changes and streamline code readability.
The buffer is a simple type cast of the clipboard, without any modification applied, which is then handed to the provider's `write` function.

### Screenshots
The new shortcut is visible in the menu, just below the other paste options:
image

### DIFF
--- a/plugins/builtin/romfs/lang/en_US.json
+++ b/plugins/builtin/romfs/lang/en_US.json
@@ -818,6 +818,7 @@
         "hex.builtin.view.hex_editor.menu.edit.jump_to.curr_pattern": "Current Pattern",
         "hex.builtin.view.hex_editor.menu.edit.open_in_new_provider": "Open selection view...",
         "hex.builtin.view.hex_editor.menu.edit.paste": "Paste",
+        "hex.builtin.view.hex_editor.menu.edit.paste_as": "Paste...",
         "hex.builtin.view.hex_editor.menu.edit.paste_all": "Paste all",
         "hex.builtin.view.hex_editor.menu.edit.paste_all_string": "Paste all as string",
         "hex.builtin.view.hex_editor.menu.edit.remove": "Remove...",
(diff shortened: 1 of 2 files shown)


iTrooz:
Would you mind checking if other hex editors have a different keybind ? I would prefer to align with user expectations from other editors

iTrooz:
Also, wouldn't be better to add the option to "Paste as string", rather than "Paste all as string" ? AFAIK, "Paste" should be preferred to "Paste all" in general

BioTheWolff:
> Also, wouldn't it be better to add the option "Paste as string", rather than "Paste all as string" ? AFAIK, "Paste" should be preferred to "Paste all" in general

That seems fair enough.

> Would you mind checking if other hex editors have a different keybind ? I would prefer to align with user expectations from other editors

I have looked at one or two editors, and I don't see nearly as much paste options as we have in ImHex.

Another suggestion from AxCut ([context here](https://discord.com/channels/789833418631675954/789840633414025246/1316062059057909803)) was to create a submenu for Paste options.
Maybe `Paste`, `Paste all` and `Paste as string` could be grouped in a single submenu called `Paste...`
This could also suggest removing the keybind for the new `Paste as string` option, as it seems pretty rarely used over pasting bytes anyways.

What are your thoughts on this @iTrooz?

iTrooz:
> This could also suggest removing the keybind for the new Paste as string option, as it seems pretty rarely used over pasting bytes anyways.

Idk tbh, but I'm not against it


> Another suggestion from AxCut was to create a submenu for Paste options.

The regular "paste" option seems to useful to be put in a submenu. But you can keep the "Paste" entry and use a "Paste ..." submenu, just like the current "Copy" and "Copy as" buttons

BioTheWolff:
I have implemented the suggested changes. You can find them here:
image

WerWolv:
Thank youè

PR #2008, opened 2024-12-17
fix: more tabs to deal with.

paxcut:
This time it may be for good. 

As requested I added the preprocessing functionality into the pattern editor. I had to duplicate a few functions and update the c++ library to version 20. but now I can make sure the pattern editor doesn't see a tab again. I also removed the preprocessor from where it was before because it is not needed anymore. The changes were tested using a file that used tabs for all its white space. The file was pasted into the pattern editor and imported as well. I both cases no crashes occurred and the files had no tabs on them.


WerWolv:
Appreciate it, thanks a lot!

PR #2009, opened 2024-12-17
improv: VScode like scrolling

paxcut:
I have implemented a fix that makes the scrollbars behave like they do in VScode. Vertically you can scroll past the end of the file until only the last line can be seen at the top. The horizontal bar behaves as if every line was the same length which is the length of the longest line in the file. Not only this creates a better user experience, but it also fixes the annoying flicker that occurs when scrolling through large files. Also, I have switched all the old draw calls to render text to regular TextUnformatted calls which adds extra stability to the resulting display.

To implement the behavior I added a dummy widget with the desired dimensions.

### DIFF
--- a/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
+++ b/lib/third_party/imgui/ColorTextEditor/include/TextEditor.h
@@ -559,6 +559,7 @@ private:
 	int GetCharacterColumn(int aLine, int aIndex) const;
 	int GetLineCharacterCount(int aLine) const;
     int Utf8CharsToBytes(const Coordinates &aCoordinates) const;
+    int GetLongestLineLength() const;
     unsigned long long GetLineByteCount(int aLine) const;
 	int GetStringCharacterCount(std::string str) const;
 	int GetLineMaxColumn(int aLine) const;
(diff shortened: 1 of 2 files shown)


paxcut:
No I haven't but it seems to verify my experience. When I had finally gotten a flicker free pattern editor I turned the textUnformatted back to draw calls and the flickers started immediately.  It is in fact the speed of the draw calls that is in part responsible for the superimposition of scroll values that get drawn quickly due to the draw calls. The calls to textUnformatted have an effect of stabilization so that changes dont occur too quickly.

As for the calculation of the longest line I should mention that I always test these changes with my monster pattern of 1000's of lines and they suffer no ill effect. Having said that this is only a first approximation to resolving many issues with scrolling and the text editor that I will be committing slowly but steadily so as to not overwhelm with changes. I will be improving the improvements as I go along but right now it makes more sense to submit this first step and see if it really helps with the annoiyng scroll jumping.

WerWolv:
Okay thank you. Let me know when it's ready for merging

paxcut:
Here is a video of the scrolling in action.The PR is ready for merging now. Further improvements will be added if necessary but I don't think they will be. There are improvements that take priority now because they have a detrimental effect on the user experience far greater than some minor slow down that I haven't seen in my slow computer and big files.


https://github.com/user-attachments/assets/ca4516c9-cc35-4b06-891d-1b28cb1db94b

paxcut:
thanks, this is the same file on the current editor using draw calls. As hard as I try I can't get it to jump like that when using textunformatted. Faster is not always better i gues. 

https://github.com/user-attachments/assets/0e5057b5-0925-415d-a26d-f728144f4243

PR #2012, opened 2024-12-17
Updated main -> gui -> source -> main.cpp

Isaadqurashi:
Exception handling was added to catch any runtime errors.
Safe handling of the Linux distribution information was implemented to avoid potential crashes when the value is unavailable...
Improved logging for the Linux distribution to handle empty versions and missing distro data.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,7 +76,7 @@ jobs:
         . venv/bin/activate
         pip3 install gcovr
         cd build
-        gcovr --gcov-executable /usr/bin/gcov-14 -e '.*yara_rules.*' -e '.*third_party.*' -e '.*external.*' -r .. --xml coverage_report.xml --verbose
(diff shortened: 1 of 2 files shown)


WerWolv:
Thanks for the PR! I'm generally okay with the changes. However, as with the try-catch block in window.cpp, it should be #if-defined out in debug builds as it will make it very annoying to debug uncaught exceptions.

Additionally, instead of just logging to the console, it might be nice to display an OS popup using `hex::nativeErrorMessage` that shows the error message to the user

Isaadqurashi:
Thank you for the feedback! 
I understand your concern about debugging uncaught exceptions in debug builds. I'll make sure to conditionally include the `try-catch` block using `#if defined` to ensure it's excluded in debug builds, making debugging more straightforward.

Regarding the logging mechanism I agree that displaying an OS popup using `hex::nativeErrorMessage` would provide a better user experience by immediately notifying the user of the error. I'll update the implementation accordingly...

However my finals are approaching, so I might need some time to make these adjustments. I'll ensure the changes are pushed as soon as possible after my exams. Please let me know if there's anything else you'd like me to consider in the meantime.

WerWolv:
No rush! I'll keep the the PR own until you're done :) Thanks a lot

PR #2016, opened 2024-12-18
fix: console either printed no newlines or printed extra ones.

paxcut:
The reason for the error were the text preprocessor added to deal with tabs was created to process entire files, not just lines or words. In that context it cleaned trailing new lines if any existed but in the new role that's not only unwanted but erroneous.

### Problem description
After that was fixed the console started to add two empty lines between each output line. When splitting a string using new lines you need to not add a new line to the last line created. Even if the text ends in an end line, the split screen code is set to not discard empty lines.

The fixes are straightforward.


WerWolv:
Thank you, noticed that today too ^^

Issue #2033, opened 2024-12-27
[Bug] `hex::dec::lzma_decompress` reports an error when decompressing a small buffer


AlexGuo1998:
### Operating System

Windows

### What's the issue you encountered?

When decompressing a short lzma-compressed buffer, `hex::dec::lzma_decompress` reports an error, but the data is written to the section with an incorrect size of 100 bytes.

### How can the issue be reproduced?

1. Generate an `xz` compressed file with the following Python code:
```python3
import lzma

decompressed = b'123'
compressed = lzma.compress(decompressed, format=lzma.FORMAT_XZ, preset=1)
with open('123.xz', 'wb') as f:
    f.write(compressed)
```
The hexdump of the result `123.xz` file is:
```text
FD 37 7A 58 5A 00 00 04 E6 D6 B4 46 02 00 21 01 10 00 00 00 A8 70 8E 86 01 00 02 31 32 33 00 00 61 C5 1C 07 44 28 23 30 00 01 1B 03 0B 2F B9 10 1F B6 F3 7D 01 00 00 00 00 04 59 5A
```

2. Load the file into ImHex, apply the following pattern:
```rust
import std.mem;
import std.io;
import hex.dec;

u8 compressed[std::mem::size()] @ 0;
std::mem::Section decompressed = std::mem::create_section("decompressed");
bool ok = hex::dec::lzma_decompress(compressed, decompressed);
std::print("ok={}", ok);
```

3. `ok=false` is printed to the console. There is a "decompressed" section of 100 bytes, of which the first 3 bytes are `31 32 33`("123") and others `00`.

The expected result should be: `ok=true`, and the "decompressed" section contains only the first 3 bytes.

### ImHex Version

1.36.0

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

### Additional context?

The corresponding code is here: https://github.com/WerWolv/ImHex/blob/6005af15955eb98e9c0a888c7e46a16b1f74f296/plugins/decompress/source/content/pl_functions.cpp#L151-L204

After some debugging, it seems like first `lzma_code()` returns `LZMA_MEMLIMIT_ERROR`(6), but after adjusting the limit, it returns `LZMA_STREAM_END`(1), and is incorrectly treated as an error.

---

To fix this once and for all, noticing earlier at `lzma_auto_decoder()`

[...]

PR #2044, opened 2025-01-02
fix: Set LZMA decompressor memory limit to 1GiB fixed

AlexGuo1998:
### Problem description

See #2033 (`hex::dec::lzma_decompress` reports an error when decompressing a small buffer).

Set the LZMA decompressor memory limit to 1GiB fixed. Print a warning when exceeded, and abort with returning `false`.

### Screenshots

Normal result when decompressing a small buffer

image

Warning message when `memlimit` is exceeded: (Set to 64B here for a demo)
> W: lzma_decompress memory usage 1114168 bytes would exceed the limit (64 bytes), aborting

image

### Additional things

Is the warning wording OK? I'm not a native English speaker so please change it if you want to.

### DIFF
--- a/plugins/decompress/source/content/pl_functions.cpp
+++ b/plugins/decompress/source/content/pl_functions.cpp
@@ -155,7 +155,8 @@ namespace hex::plugin::decompress {
                 auto &section = evaluator->getSection(params[1].toUnsigned());
 
                 lzma_stream stream = LZMA_STREAM_INIT;
-                if (lzma_auto_decoder(&stream, 0x10000, LZMA_IGNORE_CHECK) != Z_OK) {
+                constexpr int64_t memlimit = 0x40000000;  // 1GiB
+                if (lzma_auto_decoder(&stream, memlimit, LZMA_IGNORE_CHECK) != LZMA_OK) {
                     return false;
                 }
 
@@ -179,8 +180,8 @@ namespace hex::plugin::decompress {
 
                     if (res == LZMA_MEMLIMIT_ERROR) {
                         auto usage = lzma_memusage(&stream);
-                        lzma_memlimit_set(&stream, usage);
-                        res = lzma_code(&stream, LZMA_RUN);
+                        evaluator->getConsole().log(pl::core::LogConsole::Level::Warning, fmt::format("lzma_decompress memory usage {} bytes would exceed the limit ({} bytes), aborting", usage, memlimit));
+                        return false;
                     }
 
                     if (res != LZMA_OK)


paxcut:
The changes proposed here don't fix any errors and only limits the sizes of files that can be decompressed. The issue mentioned misrepresents incorrect usage of the decompressor and claims that it results in an error. It is impossible to compress three bytes using lzma. 100 bytes is about the smallest size where you can see some decrease in size although it will be very small. The function does not return ok in that case to reflect that fact as well as the size of the decompressed file being bigger than that of the compressed file. Using normal file sizes you will never see the issues reported that this PR is trying to fix.

Currently it is possible to decompress files bigger than 1 GB. By removing the call to lzma_code() after adjusting the limit you are preventing that from happening. That is not a fixed limit for lzma. It just tries to use that at first and needs to have the size returned to adjust it so that the exact size can be obtained. I don't think it is possible to implement lzma with just one call to lzma_code() that can decompress arbitrarily large files. 1 gb is not that big for file sizes that need compression so imposing that limit and making it impossible to decompress bigger files doesn't seem all that safe to me.

AlexGuo1998:
> The changes proposed here don't fix any errors and only limits the sizes of files that can be decompressed. The issue mentioned misrepresents incorrect usage of the decompressor and claims that it results in an error. 

The memory limit is for limiting the internal allocation of the LZMA decompressor, not for limiting the output file size.

> It is impossible to compress three bytes using lzma. 100 bytes is about the smallest size where you can see some decrease in size although it will be very small.

See the referenced issue. If you compress a 3 bytes buffer it sure will expand multiple times, but the result is still a valid LZMA-compressed file. See this hexdump for example:

```test
FD 37 7A 58 5A 00 00 04 E6 D6 B4 46 02 00 21 01 10 00 00 00 A8 70 8E 86 01 00 02 31 32 33 00 00 61 C5 1C 07 44 28 23 30 00 01 1B 03 0B 2F B9 10 1F B6 F3 7D 01 00 00 00 00 04 59 5A
```

Save this file yourself and use `unxz` on it. This LZMA-compressed file (using the `.xz` container format) is 60 bytes long, but decompresses to just 3 bytes `31 32 33` or "123".

> Currently it is possible to decompress files bigger than 1 GB. By removing the call to lzma_code() after adjusting the limit you are preventing that from happening.

In fact, this commit decompresses files bigger than 1GiB just fine.

2GiB file decompressed

paxcut:
How can you call a file that is 60 bytes  a compressed lzma file for a file that is 3 bytes? Compressed files that are bigger than the decompressed file is the result of an invalid compression. The call to lzma_code not being ok is only for those invalid compressions. When files are truly compressed the function return ok and the size is what is expected. for files that were originally smaller than 100 bytes lzma  is not able to compress them. There is no practical advantage to being able to decompress files that are bigger than the files they are supposed to be compressing.

I guess I misspoke. You are correct in that it is not the file size but the internal allocation. You are still limiting the sizes of the files that can be decompressed. You are also always allocating 1 gb of internal memory to decompress files that don't need so much.

The current code has no such limitations. The initial call to lzma_code that returns the memory error is only used to allow the following function calls to determine the exact amount of memory that needs to be allocated to decompress any file of any size.
```cpp
   auto usage = lzma_memusage(&stream);
  lzma_memlimit_set(&stream, usage);
  res = lzma_code(&stream, LZMA_RUN);
```
The first line determines how much memory is needed to decompress the file and the second sets the maximum memory usage limits to that exact value to ensure that the third call obtains the uncompressed file with the smallest possible memory usage. By removing the last two lines  your PR reverts the code to the previous state where it wasn't able to adjust the memory use making it impossible to decompress files that require more memory than 1 gb. It offers no gains over the current code. lzma doesn't need 1 gb of internal memory to decompress most files. 

The current code will only use the exact amount needed, no more no less. If it needs more than 1 gb it will not only use more than 1 gb, but if it needs less it will also use less. Fixing it to 1 gb is both unnecessary and wasteful plus it limits the files that can decompress. Moreover the reason given to justify the change is to enable the decompression of files that are not smaller in size than the original ones a situation that is not going to occur in practice because nobody compresses files that are smaller than 100 bytes.  The files that requires more than 1gb of memory to be decompressed properly will certainly be compressed to sizes smaller than the originals but the changes in this PR  makes their decompression impossible while it makes the decompression of invalid ones seem like they can also compressed.

AlexGuo1998:
> How can you call a file that is 60 bytes a compressed lzma file for a file that is 3 bytes? Compressed files that are bigger than the decompressed file is the result of an invalid compression.

Because such files exist in the wild.

> The call to lzma_code not being ok is only for those invalid compressions. When files are truly compressed the function return ok and the size is what is expected.

This is not the case here. Again, `unxz` exits with code 0 (OK) for those files.

> for files that were originally smaller than 100 bytes lzma is not able to compress them. There is no practical advantage to being able to decompress files that are bigger than the files they are supposed to be compressing.

This is not something we should bother thinking when hacking others' binary files.

> You are still limiting the sizes of the files that can be decompressed.

This situation should be very very rare. To achieve a 1 GiB decompression memory usage, you have to set a custom compressor filter chain when compressing, which is not common. Such files is troublesome to use because it requires 1 GiB RAM to decompress. And the decompressed file needs to be more than 1 GiB for that dictionary size to be useful.

In this situation, you can always just decompress it with another application and load it decompressed.

> You are also always allocating 1 gb of internal memory to decompress files that don't need so much.

This is just a _limit_ not an _allocation_.

If you use `UINT64_MAX` as the limit, the application won't crash with the address space exhausted. In fact, this is suggested way to disable the limiter according to the [`liblzma` doc](https://tukaani.org/xz/liblzma-api/container_8h.html#a21cbebf2771617bb1e956385cfb353e3).

WerWolv:
Changes look good to me, thanks a lot 👍If we'll ever run into some common binaries that trigger the memory limit, we can still increase it in the future I believe.

paxcut:
A while back lzma didn't work at all. I made some changes that made it work except for files that are 100 bytes or smaller. Those small files are being used as a reason to remove the changes that made lzma work without explaining why they need to be removed. lzma has a system to avoid needing to set a limit and that system is currently in use. why is using the dynamic limit setting even have to do with what happens to 100 byte files? The more I think about it the less sense it makes.

PR #2050, opened 2025-01-04
build: Make MacOS arm64 build use clang 19

iTrooz:
it also updated ubuntu to 24.04


iTrooz:
Related: https://github.com/iTrooz/macos-crosscompile/pull/1

Issue #2055, opened 2025-01-06
[Bug] Unable to Remove Characters, Copy and Paste Characters in Bookmark Comments


Croculus:
### Operating System

Windows

### What's the issue you encountered?

Whenever I make a comment on a bookmark, I am unable to delete characters from my comment (via backspace), nor can I copy/paste text into the comment. 

### How can the issue be reproduced?

Seems to be applicable for any bookmark I create across the file I am working on. To recreate, make a new bookmark, type some text, and try to use the backspace key.

### ImHex Version

1.36.2

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

MSI

WerWolv:
Hey, thanks for the report! 
We removed the built-in shortcuts of the text editor with the latest release and made them into configurable ones but we apparently forgot about the text editors in the bookmarks.

@paxcut?

paxcut:
Bookmarks use the text editor? had no idea. what else uses it?

WerWolv:
They do yeah. The only other thing is the demangler I believe but that is read only

paxcut:
I thought the only place that wasn't write-only was the pattern editor. The demangler, the console and the http requests are all write-only so they don't need shortcuts.I'll add them to the bookmarks but I wonder why is the text editor being used there at all.

PR #2056, opened 2025-01-06
fix: Bookmark comment shortcuts (issue #2055)

paxcut:
Fixes for #2055.
 
Added the registrations for the editor shortcuts and reorganized the registration calls into events and handlers.

Needed a way to select which of the possibly multiple editors sees the effects of the keyboard so a pointer to the current text editor in use was added to the view bookmarks class. The pointer may be empty so the shortcuts need to check it before using it.

To detect which of the editors needs to see the effects of the shortcuts a function ws added to text editor that tells if window has focus.

To avoid errors when editors become read only, asserts were exchanged with do-nothing early returns on the text editor.

To avoid using the default language coloring (HLSL), the colorize enabling variable was set to false.

A suggestion by cppcheck simplified the logic of a conditional on view bookmarks.

On view pattern editor some icons were added to console menu entries.


WerWolv:
I don't think duplicating these shortcut definitions everywhere where the text editor is used is a good idea... Maybe it would make sense to add a new helper class to the ui plugin that just wraps the Text Editor with the things we need and adds the shortcuts there. That way we can just use that new class everywhere and have the shortcuts work and affect the currently focused editor

paxcut:
Putting the burden of identifying who is calling it on the shortcut seems wrong. In the simplest cases you want basic navigation only like the console. Rather than a class that merges the shortcuts with the editor it may make more sense to abstract the concept of a shortcut even further to something like copySortcut, pasteSortCut, cutShortcut, etc...  that can be called from each editor user without having to specify the keys. That way you don't duplicate as much code while leaving the burden of identification to the class that using the editor like it currently does.

paxcut:
How does the hex editor define its shortcuts in ViewHexEditor when it is used outside the main hex editor (like in sections or the hex visualizer)? Is it possible to so something similar so that the bookmarks use the shortcuts of ViewPatternEditor?

paxcut:
I see how. they don't have any shortcuts either which is the same problem that the bookmarks had. There has to be a better way to handle shortcuts on a more general perspective so that we don't have entire portions of ImHex lack the functionality they should have

PR #2062, opened 2025-01-08
fix: unreported crash on 3d-visualizer

paxcut:
The recently added data checks allowed invalid sized vertex arrays to be used as if they were valid making ImHex crash. 

Moved all the error messages into localization strings for translation.


WerWolv:
Thank you! I usually use libfmt format strings for translatable strings that contain variables. That allows translators to insert those variables exactly where they want them and aren't limited to the order things are written in english. That will also reduce the number of entries you need in the language file

PR #2074, opened 2025-01-18
Draft: install include folders of plugins to sdk

### LINKED ISSUE #2068: [Bug] ImHex SDK does not contain headers for internal plugins
### Operating System

Linux

After doing make install the SDK does not contain the headers / possibility to link against internal plugins like ui.

cd build && cmake .. && make -j
sudo make install
find /usr/local/share/sdk/ -name "pattern_drawer.hpp"

### ImHex Version

992f18b94ba7e395369d8dbb7e1d722f0ba6d80b

### ImHex Build Type

- [X] Nightly or built from sources

### Installation type

make install

### Additional context?

_No response_

ev1313:
Partially fixes #2068 .

This assumes the headers are in "include" for all plugins - this is necessary since the INCLUDES argument of add_imhex_plugin may contain paths to third party libs, whose headers should not be copied. 
To fix this I think it is necessary to add a second argument like "PRIVATE_INCLUDES", which is included, but not installed.

With this it possible to append the ui plugin for example:
```
INCLUDES
        $ENV{IMHEX_SDK_PATH}/lib/plugins/ui/
```
and to link against it:
```    
LIBRARIES
        /usr/local/lib/imhex/plugins/ui.hexpluglib
```

In a follow-up in the CMake Template for plugins imho there should be fixed, that:

 - you can include plugin includes relative to the SDK Path
 - you can link plugins relative to the lib path + without the hexpluglib or hexplug extension

### DIFF
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -506,16 +506,16 @@ function(downloadImHexPatternsFiles dest)
 
         # Maybe patterns are cloned to a subdirectory
         if (NOT EXISTS ${imhex_patterns_SOURCE_DIR})
-            set(imhex_patterns_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ImHex-Patterns")        
+            set(imhex_patterns_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ImHex-Patterns")
         endif()
 
         # Or a sibling directory
         if (NOT EXISTS ${imhex_patterns_SOURCE_DIR})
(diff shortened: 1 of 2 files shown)


WerWolv:
Thank you! I did a quick and dirty fix for it yesterday but the way you did it is better.

One nit-pick, not all plugin includes should be included. We make the distinction between regular plugins (which are cmake module libraries) and library plugins (which are regular shared libraries). You can only ever link against the library plugins anyways so only those should be included. These plugins have the `LIBRARY_PLUGIN` set when defining them so you can check for `IMHEX_PLUGIN_LIBRARY_PLUGIN` in the `add_imhex_plugin` function

PR #2092, opened 2025-01-29
feat: implement subpixel rendering

paxcut:
Proof of concept for implementing subpixel processing in ImGui. This is work in progress, and it is bound to have problems.

What it does:
1) Uses freetype own subpixel processing implementation to build a 32-bit color atlas for the default font only (no icons, no unifont) . 2) Avoids pixel perfect font conversion when possible. 3) Self contained, no ImGui source code changes.
4) Results in much improved legibility of fonts rendered on low dpi LCD screens that use horizontal RGB pixel layouts (no BRG or OLED or CRT if they even exist anymore)

What it doesn't:
1) Fancy class based interface. The code is barely the minimum needed to show it can work. 2) Dual source color blending. That needs to be implemented in shader code, so it needs to change ImGui source code although minimally. This will result in some characters appearing dimmer than others. Easily fixed with small fragment and vertex shaders. 3) subpixel positioning. If characters are very thin they will look colored, or they can be moved to improve legibility. 4) deal with detection of fringe cases including rare pixel layouts, non LCD screens, Mac-OS not handling subpixel rendering and any other deviation from the standard LCD. 5) tries to be efficient in speed or memory use. Font Atlases will be 4 times the size they were before, but there are no noticeable delays in font loading in the examples I have tried.

Any comments and code improvements are welcome.


lib/third_party/imgui/backend/source/imgui_impl_opengl3.cpp
```diff
@@ -419,14 +419,47 @@ void    ImGui_ImplOpenGL3_NewFrame()
...
+bool isColorDark(ImU32 color) {
+    float r = (color         & 0xFF) / 255.0f;
+    float g = ((color >> 8)  & 0xFF) / 255.0f;
+    float b = ((color >> 16) & 0xFF) / 255.0f;
+    float luminance = 0.2126f * r + 0.7152f * g + 0.0722f * b;
+    return luminance < 0.5f;
+}
+
+void UseFontShaders(const ImDrawList *parent_list, const ImDrawCmd *cmd) {
+    useFontShaders = !useFontShaders;
```
WerWolv, line 434: I'd use two callbacks probably, one that enables the font shader and one that disables it instead of having one that toggles it. That way, when reading the code, you know exactly when the shader is active and when it isn't

plugins/fonts/include/font_atlas.hpp
```diff
@@ -256,4 +283,152 @@ namespace hex::fonts {
         std::list<std::vector<u8>> m_fontData;
     };
 
+
+    class Bitmap {
```
WerWolv, line 287: I'd probably move these classes into their own files since it's quite a bit of code and not immediately related to the font atlas

plugins/fonts/include/font_atlas.hpp
```diff
@@ -256,4 +283,152 @@ namespace hex::fonts {
...
+            m_height = height;
+            m_pitch = pitch;
+            m_data = std::move(data);
+        }
+        ImU32 crossCorrelate( ImU8 *input, ImU32 inLen, ImU32 skipY);
+        void lcdFilter();
+        //void blend( Bitmap &bitmap, ImU32 x, ImU32 y);
+    };
+
+    class RGBA {
```
WerWolv, line 315: Try to match the style of the rest of ImHex here. For example remove extra new lines and spaces, use camelCase for the functions, etc

plugins/fonts/source/font_loader.cpp
```diff
@@ -10,23 +14,144 @@
...
+#include <freetype/freetype.h>
+#include "imgui_impl_opengl3_loader.h"
 
 #include <font_atlas.hpp>
 
 namespace hex::fonts {
 
+    bool BuildSubPixelAtlas(FontAtlas *fontAtlas) {
+        FT_Library ft = nullptr;
+        if (FT_Init_FreeType(&ft) != 0) {
```
WerWolv, line 26: Isn't freetype initialized by ImGui already?

plugins/fonts/source/font_loader.cpp
```diff
@@ -10,23 +14,144 @@
...
+            fontName = "proggy";
+        } else {
+            ImVector<ImS32> rect_ids;
+            std::map<ImS32, Bitmap> bitmapLCD;
+            ImU32 fontCount = io.Fonts->ConfigData.Size;
+            for (ImU32 i = 0; i < fontCount; i++) {
+                fontName = io.Fonts->ConfigData[i].Name;
+
+                std::ranges::transform(fontName.begin(), fontName.end(), fontName.begin(), [](unsigned char c) { return std::tolower(c); });
+                if (fontName.find("proggy") != std::string::npos || fontName.find("unifont") != std::string::npos) {
```
WerWolv, line 48: Is there maybe a better way to detect if we should skip fonts instead of just matching for names? e.g what happens if you'd load a pixel perfect TTF font?

plugins/fonts/source/library_fonts.cpp
```diff
@@ -79,6 +79,112 @@ namespace hex::fonts {
         return true;
     }
 
+    ImU32 Bitmap::crossCorrelate(ImU8 *input, ImU32 inLen, ImU32 skipY) {
```
WerWolv, line 82: This also doesn't really belong here. Put that in its own file

plugins/fonts/source/font_loader.cpp
paxcut: There is no real way to characterize a font as being pixel perfect that I know of other than inspecting it. Proggy clean and unifont are not bitmap fonts. They are scalable and described using contours like any other true type font. If you want to use some pixel perfect font and you know what its native size is, you just wouldn't use subpixel rendering and use `none` for antialising. Those two need to be listed because they are part of the standard 4 fonts we load for each choice of font.

WerWolv: Okay in that case I think it would be better that we specify if they need anti-aliasing or not when registering them, e.g when calling `registerFont` have it be an argument there. That way it's more easily expandable

paxcut: The problem with that is that at this point we are only using the font  configuration data stored in the atlas and we don't have access to the font map returned by getFonts(). I load the names of the fonts on the previous function call just so I can filter the fonts that cannot be resized and don't need subpixel rendering.

plugins/fonts/source/library_fonts.cpp
paxcut: I had a lot of code that was leftover from earlier trials of different things that needed cleaning up. I have removed most of the unused stuff and will be committing those changes when we resolve the issue with the disabled text in the settings of the fonts.

lib/third_party/imgui/backend/source/imgui_impl_opengl3.cpp
paxcut: Made the change and committed. Please review it again at own convenience.

plugins/fonts/include/font_atlas.hpp
paxcut: Moved code to their own files and deleted old and unused code.

plugins/fonts/include/font_atlas.hpp
paxcut: Yes, sorry about that. Some parts of the code started out as part of Imgui source code so it was written to match imgui style. I changed it back to use imhex style as suggested.

plugins/fonts/source/font_loader.cpp
paxcut: Still working on this. I need to find some way to pass the information about which internal fonts cannot be resized so they wont be processed for subpixel rendering. maybe we can use a name like `unresizableFont` to all the fonts that need to be skipped instead of their explicit names since the names are erased as soon as the texture is registered anyway.

paxcut: I implemented a system where fonts now include a field that specifies if they are scalable which can be optionally set when registering the font. Fonts that are not scalable cannot be anti-aliased or sub-pixel rendered as that requires scaling the font. Currently only one font (unifont) being registered uses this property.

PR #2133, opened 2025-02-20
fix: Console editor missed printed output.

paxcut:
This change is to fix a bug reported in discord by berkus and Naheulf about the console missing output lines. The bug was caused by using SetText which replaces the existing text with the text in the argument.

To fix it use InsertText which puts the text at the current cursor position that was already set to the end of the current contents.

Code was tested with pattern used to reproduce the bug and seemed to work when evaluated repeatedly.


Naheulf:
Commit 8c13d0096e839ba35a1f00b5d5a3b2b099cf610f fix build actions.  So I guess you can rebase your PR.

paxcut:
Ok, I rebased it and is trying to build again. I don't expect any problems since this is probably one of the simplest PRs ever.

PR #2151, opened 2025-02-28
build: Find regex component of Boost explicitly

xtexx:
'Boost::regex' could not be found unless regex is specified explicitly on some systems.

cf. https://github.com/AOSC-Dev/aosc-os-abbs/pull/9798
cc. @chenx97

when building ImHex with system Boost.

Add `COMPONENETS regex` to require regex library explicitly.


xtexx:
The CI build failure is not related to this PR.

WerWolv:
Thank you! Yup, looks like Fedora updated their cmake, I'll take care of it

PR #2162, opened 2025-03-09
feat: Added horizontal scroll bar for hex edit

Xsakura1314:
### Overview
I have implemented a horizontal scrollbar for the hex editor, which allows displaying columns that are outside the window when there are multiple columns. It also supports address jumping and displays the corresponding column.

I control the content display in the Byte Column by recording the offset value of the horizontal scrollbar.

### Screenshots
PixPin_2025-03-09_20-12-02_1741522328880_0

### Additional things
As you can see, there is currently a flickering issue in the ASCII column when scrolling to the left. I found that disabling the `ImGuiTableFlags_SizingFixedFit` flag resolves this issue. However, I noticed that when adding columns, the flickering does not occur, which is quite confusing. I have tried to analyze this but haven't made any progress. If you have any insights or suggestions, please let me know.

### DIFF
--- a/plugins/ui/include/ui/hex_editor.hpp
+++ b/plugins/ui/include/ui/hex_editor.hpp
@@ -108,6 +108,7 @@ namespace hex::ui {
         void drawFooter(const ImVec2 &size);
         void drawTooltip(u64 address, const u8 *data, size_t size) const;
         void drawScrollbar(ImVec2 characterSize);
+        void drawHorizontalScrollbar(ImVec2 characterSize);
         void drawMinimap(ImVec2 characterSize);
 
         void handleSelection(u64 address, u32 bytesPerCell, const u8 *data, bool cellHovered);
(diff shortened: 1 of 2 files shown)


paxcut:
the top row with the byte column number don't seem to scroll with the bar.

paxcut:
if we call the top row of numbers that indicate the column byte index the x axis and the addresses on the left the y axis then I think that both should always be visible no matter where you scroll.

 When you  scroll up and down then the y axis should be the only one changing. If you scroll left and right then the x axis should be the only one changing. there isn't a case when you fix one or the other except when  there is no scroll bar in that direction, ie, the whole window fits.
 
Another thing to consider is the fact that you are including the ascii column in the scrolling unit so that it is not possible to see the ascii values for the left most hex values at the same time. It seems that a better choice would be to always show both and have them slide together with the scroll bar.

Xsakura1314:
Fixing the ASCII column is an interesting feature, and I am considering implementing it in the future.

Currently, I have changed the way the address bar is displayed. When the scrollbar moves, the row address value remains fixed, while the column address value changes.

However, the most frustrating issue for me right now is the flickering of the ASCII column. If anyone has the expertise, I would greatly appreciate help in resolving this issue.

PixPin_2025-03-10_18-47-39

WerWolv:
Hey! Thanks a lot for the PR

The flickering ASCII column I noticed already too. It's there even without your PR.
_I believe_ that it's because the number of columns changes, it needs to reflow the table and recalculate all the column widths. That takes a frame so on the first frame, the width is close to zero so it doesn't render anything.
One fix might be to pre-calculate the new width and setting the column width manually

PR #2171, opened 2025-03-16
i18n: Added french translation

GekySan:
Hello, this pull request adds a complete french translation.


WerWolv:
Hi! Great work! Thank you very much

@iTrooz If you have any time, could you do a quick check over the translations? Your French is significantly better than mine :D

Naheulf:
I can help for that too. (I'm French), will do it this week.

Naheulf:
I started the review this morning and I found an inconsistency in the source (English) strings. Let's take these two English 'translations':
> "hex.builtin.achievement.starting_out.docs.desc": "Open the documentation by selecting Help -> Documentation from the menu bar.",

and
> "hex.builtin.achievement.patterns.load_existing.desc": "Load a pattern that has been created by someone else by using the 'File -> Import' menu.",

The latter does use quotes around menu path, but the former don't. Should the French translation match the source strings, or should they use only the quotes pattern? (I think the non-quotes pattern shouldn't be used.)

iTrooz:
It is unfortunately not feasible for me to review everything properly, but it seems good

We can ship the proposed translation as-is, and wait for users to potentially point out things.

iTrooz (approved):

PR #2175, opened 2025-03-20
TCP Client / Server Enhancements

GoomiiV2:
### Problem description
Messages in the TCP tool were only visible as a text string in the messages list and unable to be opened in the hex view.
There was also no way to send data other than as a typed in text string in the input field.

* Added the ability to double-click on a message in the messages list to open it in a provider tab.
* Added a new button to the client tab to send the current provider to the connected server.
* Updated the list look to show the message index in a column

### Screenshots

https://github.com/user-attachments/assets/198d78c7-7c94-4f48-ac8c-3581e2ac12e4


WerWolv:
Thanks a lot! Looks great

Issue #2183, opened 2025-03-26
[Bug] SIGABRT in `TextEditor::MoveHome`


Mrmaxmeier:
### Operating System

Linux

### What's the issue you encountered?

I'm seeing a crash while editing patterns in ImHex's built-in editor.

<details>
    <summary>
    Backtrace
    </summary>

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=0x6, no_tid=no_tid@entry=0x0)
    at pthread_kill.c:44
#1  0x00007ffff6ca56d3 in __pthread_kill_internal (threadid=<optimized out>, signo=0x6) at pthread_kill.c:89
#2  0x00007ffff6c4bba0 in __GI_raise (sig=sig@entry=0x6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff6c33582 in __GI_abort () at abort.c:73
#4  0x00007ffff6ed3f70 in std::__glibcxx_assert_fail (
    file=file@entry=0x7ffff7b78108 "/usr/include/c++/14.2.1/bits/stl_vector.h", line=line@entry=0x46a, 
    function=function@entry=0x7ffff7baa588 "constexpr std::vector< <template-parameter-1-1>, <template-parameter-1-2> >::reference std::vector< <template-parameter-1-1>, <template-parameter-1-2> >::operator[](size_type) [with _Tp = TextEditor::Glyph; _Alloc = std::allocator<TextEditor::Glyph>; reference = TextEditor::Glyph&; size_type = long unsigned int]", condition=condition@entry=0x7ffff7b66079 "__n < this->size()")
    at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/assert_fail.cc:41
#5  0x00007ffff7aa1510 in std::vector<TextEditor::Glyph, std::allocator<TextEditor::Glyph> >::operator[] (
    this=<optimized out>, __n=<optimized out>) at /usr/include/c++/14.2.1/bits/stl_vector.h:1130
#6  TextEditor::MoveHome (this=0x7fffdc4313d8, aSelect=0x1)
    at /usr/src/debug/imhex/ImHex/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp:1875
#7  0x00007ffff73eb042 in std::function<void()>::operator() (this=0x7fffdc431370)
    at /usr/include/c++/14.2.1/bits/std_function.h:591
[...]

PR #2184, opened 2025-03-26
fix: fixes issue #2183.

paxcut:
Fixes issue #2183 Find details there but basically  I forgot to check for empty lines before trying to find first non-space char.


lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -1884,8 +1884,10 @@ void TextEditor::MoveHome(bool aSelect) {
     auto oldPos = mState.mCursorPosition;
     auto &line = mLines[mState.mCursorPosition.mLine];
     auto home=0;
-    while (isspace(line[home].mChar))
-        home++;
+    if (line.size() > 0) {
```
Mrmaxmeier, line 1887: This doesn't fix the case where the line is non-empty but only has spaces. Something like `while(home < line.size() && line[home]..)` is probably the right fix here. :slightly_smiling_face:

Mrmaxmeier (changes requested):

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: What do you mean by not fixing that case? I tested that and it doesnt crash anymore. The way I understood the request to make the home key skip initial spaces is that the right answer should be  the end of the string of spaces at the beginning of the line regardless of the  existence of any characters following them. When the last space is found the cursor is placed after it and before the next line so that should work fine. The Home key can become equivalent to the end key in special cases like this just like the end key becomes equivalent to the home key in an empty line.

paxcut: I cant test the latest version because Im in the process of compiling 150 files and it can take a while on my slow computer. While it is true that it seems like it should crash, I tested a line of spaces on an older version and it didn't crash so i assume it won't either in this one, but until i can debug it I can't really say why or even if it still works.

paxcut: Yes, you were right, but going out of bounds for a non-empty vector doesn't crash ImHex for some reason so i thought it was finding a new line or end of line char  but there is nothing like that. I added the code you posted to the fix to commit in a bit when I finish what I'm currently doing. thanks for the help.

Mrmaxmeier (approved):

paxcut:
The changes in this PR are superseded by the changes in PR 2193 that implements all the changes proposed here.

PR #2193, opened 2025-04-01
impr: Fix word bound selection and 'MoveHome' behavior in the pattern editor

SparkyTD:
### Problem description
This PR addresses two small issues regarding the cursor in the pattern editor (TextEditor.cpp):

1. It was not possible to move the cursor to the start of a line, if it contained leading white space characters. With my fix, the editor will behave more like other code editors. Pressing Home once will jump to the first non-whitespace character (as it did before), but pressing it again will jump to column 0. Subsequent presses will alternate between the two positions.
2. When expanding a selection with Ctrl+Shift+{Left/Right}, the new selection position would skip an additional word, resulting in the cursor landing "inside" the selection. This PR fixes this bug.

1. To fix the first issue, I simply added a condition in `TextEditor::MoveHome` to check if the cursor is already on the first meaningful character, or on one of the whitespaces preceding it, in which case the jump offset is set to 0. If we're already on column 0, then jump forwards to the first non-whitespace character.
2. This bug was happening because the word boundary jump calculations were essentially happening twice. Once in `TextEditor::MoveLeft`/`TextEditor::MoveRight`, and then a second time in `TextEditor::SetSelection`, leading to the selection skipping an additional word. I fixed this by replacing the ternary operator with just `SelectionMode::Normal`.


bojidar-bg (approved):
Tested locally, didn't spot any bugs in the changed code.

The changes to MoveRight/MoveLeft were also suggested in #2195 in a stroke of great minds thinking at the same time :sweat_smile: :joy:

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -1886,6 +1886,11 @@ void TextEditor::MoveHome(bool aSelect) {
     auto home=0;
     while (isspace(line[home].mChar))
         home++;
+
+    if (home >= oldPos.mColumn && oldPos.mColumn != 0) {
+        home = 0;
+    }
+
```
paxcut, line 1893: The code for the while loop is missing the check for home being less than the line length to prevent crashes when pressing home on empty lines. With that addition then the fix here is not as efficient as it could be since it first counts all the leading spaces of the line when it can stop counting if the number of spaces reaches the current cursor position as indicated by the home >= oldPos.column condition. 

A more efficient implementation would be to use the cursor column as the limit for the while loop when it is not zero and the line length otherwise. Then to check if the cursor must be moved to zero it suffices to check that the home variable is equal to the column value.

As an example, consider a line with 20 leading spaces and the cursor located at the second space.  Then the while loop only executes one iteration before stopping and the cursor is set to the zero position instead of having to execute 20 times unnecessarily.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
```diff
@@ -1846,7 +1846,7 @@ void TextEditor::MoveRight(int aAmount, bool aSelect, bool aWordMode) {
         }
     } else
         mInteractiveStart = mInteractiveEnd = mState.mCursorPosition;
-    SetSelection(mInteractiveStart, mInteractiveEnd, aSelect && aWordMode ? SelectionMode::Word : SelectionMode::Normal);
+    SetSelection(mInteractiveStart, mInteractiveEnd, SelectionMode::Normal);
```
paxcut, line 1849: By making what would be the only call to setselection that uses word mode use normal mode, you are already making the code that uses word mode in setselection unreachable, so removing it altogether seems more appropriate than keeping it. In fact I would remove the third parameter from set selection since it should have never had one. To see how line mode and word mode are handled look at the syntax highlighting pr in the mouse processing function.

paxcut (changes requested):
The code being edited is incomplete because it is missing the changes to avoid crashing on empty lines.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
SparkyTD: I implemented a fix and I am no longer able to reproduce a crash on empty lines.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
SparkyTD: I did look into this when I originally made this change, and the reason why I didn't delete the word mode code is because there are still some references to it in other parts of the code. For example in `HandleMouseInputs()`, double-click and ctrl-click selections are still being handled this way.

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: There is no need to check for both line.size() and limit because limit is always smaller than or equal line.size(). Only check for limit.

SparkyTD: Oh apologies, I missed that. I force-pushed the fix onto my latest commit, I hope it's okay.

paxcut: Two more optimizations could be added. 

Early return if line is empty since the cursor wont move in that case. 

Also the check for home=0 can be simplified using  
```cpp
if (home == oldPos.mColumn)
        home = 0;
 ```
 The reason for this is that we know that the line is not empty so if the cursor was at the origin then limit is the line size so it must be at least 1 or bigger. if length is 1 and the char is non space then the cursor wont move. if it is size one and the char is a space then home will be bigger than oldpos so the cursor will go to the end of the line. If home equals old pos then that means that we are at either the first non space leading char or a space leading char before it so it gets send to the origin correctly. The code wastested and no problems were found.

Comment (SparkyTD) [reply lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp:1898]: Amended my latest commit again. Should we add the early return to `TextEditor::MoveEnd` too? We would only be skipping a few branches and not loops, but it might be good for consistency and robustness.

File: lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp (hunk 2 above)
Comment (paxcut) [reply lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp:1849]:
```
To see how line mode and word mode are handled look at the syntax highlighting pr in the mouse processing function.
```
Not for example, all code that uses modes for selection other than default is now in the mouse processing function. I was able to remove the third parameter and implement the selections at the places where they were being called by choosing interactive start and end accordingly.  I am including the function here for reference. There will be some minor differences that should be easy to translate to the code thats available in the PR like the finding of the start and end of word, etc...
```cpp
void TextEditor::HandleMouseInputs() {
    ImGuiIO &io = ImGui::GetIO();
    auto shift  = io.KeyShift;
    auto ctrl   = io.ConfigMacOSXBehaviors ? io.KeyAlt : io.KeyCtrl;
    auto alt    = io.ConfigMacOSXBehaviors ? io.KeyCtrl : io.KeyAlt;

    if (ImGui::IsWindowHovered()) {
        if (!alt) {
            auto click         = ImGui::IsMouseClicked(0);
            auto doubleClick   = ImGui::IsMouseDoubleClicked(0);
            auto rightClick    = ImGui::IsMouseClicked(1);
            auto t     = ImGui::GetTime();
            auto tripleClick   = click && !doubleClick && (mLastClick != -1.0f && (t - mLastClick) < io.MouseDoubleClickTime);
            bool resetBlinking = false;
            /*
            Left mouse button triple click
            */

            if (tripleClick) {
                if (!ctrl) {
                    mState.mCursorPosition = ScreenPosToCoordinates(ImGui::GetMousePos());
                    auto line = mState.mCursorPosition.mLine;
                    mState.mSelectionStart = Coordinates(line, 0);
                    mState.mSelectionEnd = Coordinates(line, GetLineMaxColumn(line));
                }

                mLastClick = -1.0f;
                resetBlinking=true;
            }

            /*
            Left mouse button double click
            */

            else if (doubleClick) {
                if (!ctrl) {
                    mState.mCursorPosition = ScreenPosToCoordinates(ImGui::GetMousePos());
                    mState.mSelectionStart = FindWordStart(mState.mCursorPosition);
                    mState.mSelectionEnd = FindWordEnd(mState.mCursorPosition);
                }

                mLastClick = (float)ImGui::GetTime();
                resetBlinking=true;
            }

            /*
            Left mouse button click
            */
            else if (click) {
                if (ctrl) {
                    mState.mCursorPosition = mInteractiveStart = mInteractiveEnd = ScreenPosToCoordinates(ImGui::GetMousePos());
                    SelectWordUnderCursor();
                } else if (shift) {
                    mInteractiveEnd = ScreenPosToCoordinates(ImGui::GetMousePos());
                    mState.mCursorPosition = mInteractiveEnd;
                    SetSelection(mInteractiveStart, mInteractiveEnd);
                } else {
                    mState.mCursorPosition = mInteractiveStart = mInteractiveEnd = ScreenPosToCoordinates(ImGui::GetMousePos());
                    SetSelection(mInteractiveStart, mInteractiveEnd);
                }
                ResetCursorBlinkTime();
              
                EnsureCursorVisible();
                mLastClick = (float)ImGui::GetTime();
            } else if (rightClick) {
                auto cursorPosition = ScreenPosToCoordinates(ImGui::GetMousePos());

                if (!HasSelection() || mState.mSelectionStart > cursorPosition || cursorPosition > mState.mSelectionEnd) {
                    mState.mCursorPosition = mInteractiveStart = mInteractiveEnd = cursorPosition;
                    SetSelection(mInteractiveStart, mInteractiveEnd);
                }
                ResetCursorBlinkTime();
                mRaiseContextMenu = true;
            }
            // Mouse left button dragging (=> update selection)
            else if (ImGui::IsMouseDragging(0) && ImGui::IsMouseDown(0)) {
                io.WantCaptureMouse    = true;
                mState.mCursorPosition = mInteractiveEnd = ScreenPosToCoordinates(ImGui::GetMousePos());
                SetSelection(mInteractiveStart, mInteractiveEnd);
                resetBlinking=true;
            }
            if (resetBlinking)
                ResetCursorBlinkTime();
        }
    }
}
```

lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
paxcut: The early return here was added to simplify the check for moving home to zero since the crash on empty lines was already fixed and the check to move home to zero that you had did work . In general I tend to either leave other functions alone or completely optimize them as much as I can. 

The text editor is such that the devs only fixed the most visible bugs and refused to work on it to make improvements stating that the code was horrible and jard to work with. While I agree with that I also think it is not hopeless and that it is not impossible to make the code good by changing it slowly. 

If you look at the code with intent to find places where it should be changed you'll find yourself in the mother of rabbit holes. It is good that you are making these changes so that there will be no conflicts when I merge code that looks identical to the one you wrote. 

Since the syntax highlight code review is taking longer than anticipated I am rewriting it in small steps that I should be able to merge. If the code above to remove the last argument to setselection is too much just leave it as it is and I will make the changes after merging your pr.

paxcut (approved):
this looks good now

PR #2195, opened 2025-04-04
fix: Correct behavior of Ctrl-Shift-Left/Right to be closer to other editors

bojidar-bg:
### Problem description

Currently, when using <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd><-</kbd>/<kbd>-></kbd> to select a whole word (either ahead or behind the cursor), ImHex has this very weird behavior (inherited from https://github.com/BalazsJako/ImGuiColorTextEdit), where it would select an extra word on the side, as you can see in the Before recording below. This was extremely jarring for me when writing a pattern script (as I regularly use <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>-></kbd> to select a word and instantly start typing over it), so I opted to jump in the code and fix it.

`MoveRight` and `MoveLeft` use `SetSelection(..., SelectionMode::Word)` whenever both `Shift` and `Ctrl` are pressed. However, the rest of the `MoveRight`/`MoveLeft` code already takes care of setting up the selection between the previous end of the selection and the new cursor location of the selection, as well as of moving the cursor to the correct location when selecting a word.
Meanwhile, `SelectionMode::Word` tells `SetSelection` to expand the selection on both sides so they are aligned to word boundaries.
In the end, we get a double bug: we can end up selecting stuff before the cursor when moving to the right (which, if you try in any other editor, is not how it works), and we can end up selecting stuff even further away from the cursor when moving to the left.

I searched the code for `MoveRight` and `MoveLeft`, and can confirm that `aSelect && aWordMode` is true only when they are called by the handlers for the Shift-Ctrl-Left/Right shortcuts, so this change should not impact anything else.

### Screenshots

Before:

[Screencast_20250404_021828.webm](https://github.com/user-attachments/assets/81bc0514-35ad-423b-bb6b-d0f7aa7d06a3)

After:

[Screencast_20250404_022455.webm](https://github.com/user-attachments/assets/5dbcdb48-6db3-454c-b303-395ad03e3f73)

### Additional things

The pattern language was pretty cool to use, kudos! :sparkles:

### DIFF
--- a/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
+++ b/lib/third_party/imgui/ColorTextEditor/source/TextEditor.cpp
@@ -1790,7 +1790,7 @@ void TextEditor::MoveLeft(int aAmount, bool aSelect, bool aWordMode) {
         }
     } else
         mInteractiveStart = mInteractiveEnd = mState.mCursorPosition;
-    SetSelection(mInteractiveStart, mInteractiveEnd, aSelect && aWordMode ? SelectionMode::Word : SelectionMode::Normal);
+    SetSelection(mInteractiveStart, mInteractiveEnd);
 
     EnsureCursorVisible();
 }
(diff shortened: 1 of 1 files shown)


SparkyTD:
My last PR also addresses this issue: #2193

bojidar-bg:
Oh. Amazing timing :joy: I'll close this one, so we can focus on yours (:

PR #2218, opened 2025-04-26
build: Move nlohmann-json to build dependencies

notpeelz:
nlohmann-json is a headers-only library, so there's no runtime requirement for it.


WerWolv:
Thank you!

PR #2228, opened 2025-05-05
Implement skipping sequences of repeating bytes

agarmash:
This PR implements a neat little feature I missed - the ability to jump to the next/previous differing byte, skipping the chunk of repeating bytes. Very useful when you analyze a raw flash dump and want to skip the large sections of `0x00`s/`0xFF`s.

Some implementation details worth validating:
 - I wasn't sure what is the correct place to put the new menu entries into. The possible candidates were `File -> Go to address...` and `Edit -> Follow selection`. I chose the former, although the latter may be a better fit since it already states that the action is related to the selection. Overall, it may be a good moment to refine these menu entries in general.
 - I didn't add any tests since I'm not sure what is the project's policy for those. Please let me know if I need to add some!
 - I added the machine-generated translations for the new menu entries which may be considered a questionable thing. Please let me know if you're unhappy with those, I'll drop the commit.
 
 Also, thanks for such a nice tool, I use it a lot and was glad to build a new feature for it!


paxcut:
> This PR implements a neat little feature I missed - the ability to jump to the next/previous differing byte, skipping the chunk of repeating bytes. Very useful when you analyze a raw flash dump and want to skip the large sections of 0x00s/0xFFs.

Is this feature implemented in some hex editor you used before? I couldn't find anything like that to compare against. Your implementation seems like it only skips repeating bytes but a more general approach would skip words, double words, etc so if say the repeating pattern was 0xAABB you could also skip until the next different word,

> I wasn't sure what is the correct place to put the new menu entries into. The possible candidates were File -> Go to address... and Edit -> Follow selection. I chose the former, although the latter may be a better fit since it already states that the action is related to the selection. Overall, it may be a good moment to refine these menu entries in general.

Neither one of those seems like they apply. While it is true that the cursor position is being changed, there are lots of ways to move the cursor that don't necessarily belong in one of those two. What is important is the way in which the cursor is moved. In this case it is by searching, so the feature should be in the File>find menu as a separate tab of the existing hex and ascii options. The code should also be using the existing search code for the hex editor. This has the additional benefits of not having to create new menu entries and also the need to duplicate all the code that deals with providers or searching.

> I didn't add any tests since I'm not sure what is the project's policy for those. Please let me know if I need to add some!
I added the machine-generated translations for the new menu entries which may be considered a questionable thing. Please let me know if you're unhappy with those, I'll drop the commit.

If there are tests for the search feature in the hex editor thats where additional tests for the first difference should be. If there aren't then don't worry about adding one. I sorry but for translations we require that native speakers translate all the text shown in ImHex and we don't allow AI translations of any kind.

agarmash:
> Is this feature implemented in some hex editor you used before? I couldn't find anything like that to compare against. Your implementation seems like it only skips repeating bytes but a more general approach would skip words, double words, etc so if say the repeating pattern was 0xAABB you could also skip until the next different word

AFAIK no, I never saw anything similar in other editors either. As for the implementation extension - I see how skipping over a repeating pattern may be useful (skipping over the same pixels in a raw bitmap for example), but the only practical usecase I have at the moment is to only skip over a single repeating byte. To make life easier for both me and reviewers, I suggest implementing a single byte skip in this PR, and extend it in the following one.

> In this case it is by searching, so the feature should be in the File>find menu as a separate tab of the existing hex and ascii options.

Not sure if it would provide a good user experience (requiring to open a separate Search window to do the skip). Also, I'm not sure if it would be possible to assign hotkeys for the skip action in this case.

Instead of placing the new menu items in "Go to" menu, how about adding the one called "Skip until" next to it? I see it as good enough option.

> I sorry but for translations we require that native speakers translate all the text shown in ImHex and we don't allow AI translations of any kind.

No problem at all, I'll drop those!

agarmash:
Just updated the implementation, now the UI looks like this:
Screenshot_20250710_165131

As for the following:
> The code should also be using the existing search code for the hex editor.

I checked the existing implementation, but in a nutshell it's just a wrapper around the `std::search`: https://github.com/WerWolv/ImHex/blob/master/plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_find.cpp#L231
I don't need a full-blown `std::search` here, so I decided to keep the simple custom implementation.

WerWolv:
Thank you! Looks good to me :)

agarmash:
@WerWolv nice, glad to hear!

How shall we proceed with the PR then? Are you expecting some other people to review it?

paxcut:
In my opinion this would be much cleaner if it was a checkbox in the normal search so that it works on hex values and strings. The current search code can be easily adapted to have a custom comparison of NotEquals. You would get more functionality with a lot less code duplication  and without further crowding the menus. But, hey, its just my opinion.

paxcut:
> I checked the existing implementation, but in a nutshell it's just a wrapper around the std::search: https://github.com/WerWolv/ImHex/blob/master/plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_find.cpp#L231
I don't need a full-blown std::search here, so I decided to keep the simple custom implementation.

The point is that there is no need to reinvent the wheel. We have a search function and you are implementing a search function. So we end up with two search functions that do the same thing, not to mention  the duplication of the provider code and the reading of the data and pretty much everything. you are just adding a `not of` option to the search. If we could use regular expressions when searching for sequences, there would be no need to even add an option for it during search. Another option would be to add special notation like `!` to denote match anything that is not this value.

agarmash:
@paxcut Let's consider the example I had when I decided to build this feature - skipping a long chunk of `0xFF`s in the raw flash dump. With the current implementation, it's sufficient to set the cursor on any of the `0xFF`s and press `Ctrl+]` to skip forward or `Ctrl+[` to skip back.

The option you are suggesting is indeed more flexible, but if I understood it correctly, I would have to open the Search window, put `0xFF` there, select "Not equal", and press Search. And each time I want to make a skip, I need to do this all over again. That feels clunky for a more-or-less basic navigation feature, and it makes the search itself harder to use.

What are you thoughts?

(I hope my message doesn't sound impolite or pushy - English is my second language, so I may not fully know all the nuances that set the tone!)

paxcut:
If I want to find the next example of the byte that is under the hex editor cursor there is no shortcut that will do that.You have to open the search type the byte and click the arrow.  Does that mean that we need to add a new function and menus to find the next byte that matches the one under the cursor? In my opinion that would be a mistake. 
For example, if the search window had the values of the current cursor already on the search field and if checking the `Not Of` checkbox triggered the search then there would be only two actions (ctr-s and click No Of) instead of one which is not terrible.

Also adding a shortcut for the action doesn't mean that it also needs a menu entry. There are lots of shortcuts with no menu entries. You can add a comment in the search window with the shortcut to the next unmatched byte so once users find it they no longer need to go through all the steps needed to perform that action when using the search window.

agarmash:
> If I want to find the next example of the byte that is under the hex editor cursor there is no shortcut that will do that.You have to open the search type the byte and click the arrow.

That's true, but the usecases a little bit different. For finding the next occurrence of a given byte, the search function works flawlessly. For skipping the repeating sequences, not so much.

Take a look at the following example. The arrow below the values represents the cursor position after each invocation of "next differing byte":

```
1.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
^

2.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
            ^

3.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
               ^

4.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
                  ^

5.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
                     ^

6.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
                        ^

7.
FF FF FF FF DE AD BE EF 00 00 00 00 00 FF FF FF FF FF
                                       ^
```

The key point here is that the value under the cursor gets picked automatically as you progress. You don't have to paste it each time in the search input field. While this feature conceptually acts like a search, from the usage point of view it's closer to navigation rather than search.

paxcut:
From what I gather your main complain about my objections is that you say it would take many steps to do the same operation if you had to go through the search window and you illustrate that point with an example that shows how you move from on point to the next using ctrl+[ and ctrl+]. This illustrates how you don't have to type the value of the byte and you seem to imply that it is only possible to do that under your implementation.  The ctrl+[ and ctrl+] keys should always work exactly as they do now so why and where would you have to type the value of the byte? 

You are also not considering that if you are adding code to the search function itself that you can also implement the feature that  if no value is entered in the field the one under the cursor will be used for searching. Then even when you open the search window you still don't have to type anything.

Since using the shortcut works the same for both yours and my implementation lets consider using the menus. As you have now implemented it takes three clicks to go to the next not of byte every time you want to go to a new one. Using the search window it would take 4 the first time but only one click after that.

agarmash:
> You are also not considering that if you are adding code to the search function itself that you can also implement the feature that if no value is entered in the field the one under the cursor will be used for searching. Then even when you open the search window you still don't have to type anything.

That's another way to do it, yes. However, is it really intuitive? To use it, one have to know that:
1. The value under the cursor will be used as a default value for searching;
2. The search value changes every time the cursor moves;
3. All of the above in combination with "not equal" option would allow them to skip the sequences of repeating bytes.

While this would technically work, this would end up being an obscure feature nobody knows about. It's also unclear now to make it possible with hotkeys, and how the users would learn about those hotkeys. And generally, the whole dance of 'opening search, putting nothing in the input field, but selecting the "not equal"' feels like something I personally would be reluctant to do compared to a dedicated menu entry / hotkey.

> Since using the shortcut works the same for both yours and my implementation lets consider using the menus.

Sorry, I'm not sure that I understood you

agarmash:
Hey everyone, how shall we proceed with this PR?

WerWolv:
Just tested this out and it works great, thanks a lot!
One thing to consider, maybe we should add some upper bound of the number of bytes that get searched so ImHex doesn't end up getting stuck in a near-infinite loop. I'm not sure what a good upper bound would be though to not cripple the feature

agarmash:
@WerWolv Thanks for testing and merging!

About the upper bound - literally this morning I was analyzing a 64 GB flash dump and felt the struggle of the blocked main thread myself 😅
I was planning to mark this PR as draft and explore the option of wrapping the skip functionality with `Task`, so instead I'll simply do it in a separate PR.

Besides that, I uncovered some other rough edges related to analyzing large files, will try to address those as soon as I can

Naheulf:
If you want some default limits I think you can use 1024 repetitions of the sequence or 1MB of data, whatever came first. And if you can also add a way to override those limits this will be even better. (Because you will always find a case that requires bigger limits)

agarmash:
I'm not a fan of adding the skip limit since it defeats the whole purpose of the feature - to skip the chunk of repeated bytes and be sure that you haven't missed anything in the process.

In my example I was skipping ~1 GB holes in the flash dump, and it was absolutely delightful. Moreover, some people use hex editors to inspect the raw contents of storage media, so it's not that uncommon to have the holes of tens to hundreds of gigabytes. So for this feature I'd rather wrap it with the `Task`, so it executes on a background thread and can be cancelled at any moment.

PR #2231, opened 2025-05-08
Added support for AppleClang compiler id

alfishe:
Added support for AppleClang compiler id on Macs. Tested on latest OS and Clang - works well

### Problem description
ImHex build process stops due to failed compiler id check on Macs

Just added AppleClang to a list of allowed compilers

### Screenshots
N/A

### Additional things
N/A


WerWolv:
Thank you for the PR, this is intended however.
AppleClang is generally multiple years behind upstream Clang which makes it a pain in the ass to support when using modern features. I know Apple updated their clang version very recently, unfortunately they only do that pretty rarely so even if things work now, they might not in a few months. That's why that check exists.
Officially supported are only GCC and LLVM Clang.

Issue #2221, opened 2025-04-30
[Bug] ImHex crashes when analysing any PE file


shewitt-au:
### Operating System

Windows

### What's the issue you encountered?

ImHex crashes when analysing any PE file.
I also tested it on a pattern file I wrote for Commodore 64 Basic files. No crash there.

### How can the issue be reproduced?

Load a PE file.
For some reason to auto-detect doesn't work anymore, so select the PE file pattern.
The application crashes.

### ImHex Version

1.37.4

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

WinGet

paxcut:
This simply cannot be true because part of the unit tests includes a PE file (pe.hexpat.exe)  that is ran against the PE pattern on every build and the unit test does not fail. I have confirmed this by manually loading the PE executable file used as a test and running the PE pattern on it without any crashes. 

Please use the attached PE executable (unzip it before opening it in ImHex) to check for crashes. If they occur then you either are using an old version of the PE pattern which needs to be updated via the content store or there are local modifications in the PE pattern file that are responsible for the crashes. If no crashes occur then provide or name a specific PE executable file that consistently crashes ImHex when the PE pattern is run on it.

[pe.hexpat.exe.zip](https://github.com/user-attachments/files/19996837/pe.hexpat.exe.zip)

shewitt-au:
I believe ImHex sends crash dumps? It's true alright. It would seem something in my environment is a catlyst. I'm setting things up to debug it. The MSYS2 environment is new to me.

shewitt-au:
Yes. That exe also crashes.

paxcut:
Are you sure you are using the latest version of the PE pattern from the Content Store? If I recall correctly it has been updated recently a few times so make sure you click update if the button doesn't say delete.

shewitt-au:
All I have now is. I'm trying to set up a better debugging enviromnet,

It's an access violation
rax=308389ef44dd5132
    
ui!ZNK3hex2ui13PatternDrawer12sortPatternsEPK19ImGuiTableSortSpecsPKN2pl4ptrn7PatternES9_+0x348:
00007ffb`d003ecf2 488b00          mov     rax,qword ptr [rax] ds:308389ef`44dd5132=????????????????

And a stack:
ui!ZNK3hex2ui13PatternDrawer12sortPatternsEPK19ImGuiTab
[...]

PR #2240, opened 2025-05-12
Strict weak

shewitt-au:
### Problem description
[Bug] ImHex crashes when analysing any PE file #2221

This is a fix for the ImHex bug "ImHex crashes when analysing any PE file #2221".  The fix requires changes to the Pattern Language and the ImHex UI. Revision would be wise. We want to avoid collateral damage. It's a big code base and I'm new to it and the compilers/build-systems used. And the bug is complex and low-level. I suspect this will fix other random crashes. The problem is caused by two issues:
 - The std::sort algorithm conjuring up garbage due to the sorting criteria not being a strict weak ordering. See  [this](https://github.com/Voultapher/sort-research-rs/blob/main/writeup/sort_safety/text.md) link. 
 - We sort shared_ptr&lt;ptrn::Pattern&gt; by pointer value, and the object is a clone.  In essence we're changing the values as we're sorting.


shewitt-au:
Mistakenly included additional file.

PR #2242, opened 2025-05-12
This is a fix for the ImHex bug "ImHex crashes when analysing any PE file #2221"

shewitt-au:
### Problem description
This is a fix for the ImHex bug "ImHex crashes when analysing any PE file #2221"

This is a fix for the ImHex bug "ImHex crashes when analysing any PE file #2221".  The fix requires changes to the Pattern Language and the ImHex UI. Revision would be wise. We want to avoid collateral damage. It's a big code base and I'm new to it and the compilers/build-systems used. And the bug is complex and low-level. I suspect this will fix other random crashes. The problem is caused by two issues:
 - The std::sort algorithm conjuring up garbage due to the sorting criteria not being a strict weak ordering. See  [this](https://github.com/Voultapher/sort-research-rs/blob/main/writeup/sort_safety/text.md) link. 
 - We sort shared_ptr&lt;ptrn::Pattern&gt; by pointer value, and the object is a clone.  In essence we're changing the values as we're sorting.
 
### IMPORTANT
I'm not sure how "plugins/builtin/source/content/data_formatters.cpp" got into the PR. Been trying for an hour to rectify. I'm not a Git expert (the last time I used source control seriously SourceSafe was a thing) please ignore that file. It's a fix for another PR I submitted. I suspect I stuffed up the branching and merging.


paxcut:
Git is pretty easy to use once you learn the fundamentals. You start by forking the repo you want to contribute to and then cloning the fork into your computer hard drive. the fork is your origin and you do all the work in your local clone. You interact with git by entering the local folder in the command line and issuing git commends.

In order to have changes done to local files being included in a Pull Request you need to issue three git commands. the first one is to git add the files that have been changed. this is known as staging the files.All the files that have been added can be  committed using git commit. To make the files go into your fork you need to push the commits using git push. Then in the web browser page of your repo you will get a message stating that there are files changes and it will ask you if you want to create a Pull Request using them.

At any moment you can issue git status command that will tell you information about your local clone. things like the files that have changed, the files that were added and it will inform you of the commands to issue to do certain operations. Here is an example
```
$ git status
On branch Syntaxhighlights
Your branch is up to date with 'origin/Syntaxhighlights'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   plugins/ui/source/ui/hex_editor.cpp

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        cmake-build/

no changes added to commit (use "git add" and/or "git commit -a")
```
this is already telling you how to discard changes using git restore. If you want to keep the changes, then just fon't add the file before you commit and it wont be included.

WerWolv:
I can take care of the merging process later. The "unwanted" changes still look good to me ^^

plugins/ui/source/ui/pattern_drawer.cpp
```diff
@@ -1139,7 +1139,7 @@ namespace hex::ui {
     }
 
     bool PatternDrawer::sortPatterns(const ImGuiTableSortSpecs* sortSpecs, const pl::ptrn::Pattern * left, const pl::ptrn::Pattern * right) const {
-        auto result = std::strong_ordering::less;
```
shewitt-au, line 1142: This is important. The only value can can be unconditionally returned from a strict weak compare function is false. Returning anything but std::strong_ordering::equal unditionally will violate the strick weak ordering.

PR #2264, opened 2025-05-20
Faster pattern picker popup

shewitt-au:
### Problem description
I’ve found ImHex to be impressively zippy. I’ve no doubt that’s intentional. Why use the GPU to render the UI otherwise?

Opening the pattern file selection popup, however, takes ages (on my system at least). On a release build just over 7 seconds (much worse in debug builds, around 50 seconds!). Enough time for one to wonder if it’s working.

The reason is that ImHex has to preprocess every pattern file to extract the pattern's descriptions.
This PR uses a standalone parser designed just extract the description, and only loads the initial portion of the file when doing so, assuming the description pragma will be near the beginning.

Pretty much as described above. I won’t blame you if you object to the handwritten parser however. Noxiously fiddly problematic things. That said, some effort at least has been expended to match the behaviour of pattern_language’s parsing.

There's a magic number.

Even so, I suggest you try it out and see the difference. As more and more patterns are added the problem will only get worse and worse.

This is my second attempt at this PR. Rouge files made it into my last attempt. Still leanring how to use git properly. Sorry.

### DIFF
--- a/plugins/builtin/CMakeLists.txt
+++ b/plugins/builtin/CMakeLists.txt
@@ -50,6 +50,7 @@ add_imhex_plugin(
         source/content/data_information_sections.cpp
 
         source/content/helpers/demangle.cpp
+        source/content/helpers/pattern_description_parser.cpp
 
         source/content/data_processor_nodes/basic_nodes.cpp
         source/content/data_processor_nodes/control_nodes.cpp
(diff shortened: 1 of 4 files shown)


shewitt-au:
I've reopened this one. Turns out I can't live without it. So much faster. I apply it to every build I make for debugging when I have to open patterns for example.

paxcut:
it is pretty slow for me too. it is not the auto loader it is when you click on import pattern on the menu

shewitt-au:
And the with the cutom parser


https://github.com/user-attachments/assets/be6a9e28-eeb3-46e0-be33-946aab827336

shewitt-au:
App hangs loading fonts on splash screen.

shewitt-au:
Ok. Killed the process and it started up the second time. It's slower than with the custom parset but still much faster than before both our changes. The code is a lot cleaner. I think it's a good compromise .

PR #2276, opened 2025-05-26
Easier pattern drawer node expansion/selection

shewitt-au:
### Problem description
In the "Pattern Data" view, trying to select an expandable item in name column is hard. It expands or collapses the node but does not select it. It's frustrating, even though you can simply click in one of the other columns.

I separated the two functions. Clicking on the arrow expands or contracts the node and clicking elsewhere selects.

### Caveat
I’ve heard the terms “immediate mode” and “retained mode”, but never knew what they were. I was born in the 70s! My first look at an immediate mode UI was a shock to say the least. I think what I’ve done is sound, but what would I know? I didn't even realise the "Im" in "ImGui" stood for "Immediate Mode". I just thought it was cute. My changes do improve useability however.

There are too many failing checks for my liking. I can't see how they could be releatd to my changes however. Seems related to new stack tracing code.


shewitt-au:
My mouse has two buttons. You should not have to use two hands to select/expand a node. No other software on any OS I know of requires this.

WerWolv:
I actually quite like how it feels with your changes. Though I think it would be good if we could somehow make the hover highlighting only show up when hovering over the arrow. It feels weird that the name gets highlighted too now

shewitt-au:
On a side note, given my ignorance of IM stuff, I'd be interested in hearing why you chose that path.

WerWolv:
Which path do you mean?
Immediate Mode I chose because I wanted to try it out and it turned out to make writing UI quite pleasant and fast. 
The pattern data tree I did this way because the ImGui Demo told me to

shewitt-au:
Yeah, I meant the immediate mode UI stuff. It mixes the rendering and user interaction logic, something that, before seeing it, I would have thought a bad idea. I can see how it would be particularly well suited to game engines even from this initial exposure however.

plugins/ui/source/ui/pattern_drawer.cpp
```diff
@@ -456,16 +456,28 @@ namespace hex::ui {
...
+                    retVal = ImGui::TreeNodeEx(this->getDisplayName(pattern).c_str(), ImGuiTreeNodeFlags_SpanFullWidth | ImGuiTreeNodeFlags_DefaultOpen | ImGuiTreeNodeFlags_OpenOnArrow);
+                    break;
                 case Flattened:
-                    return ImGui::TreeNodeEx(this->getDisplayName(pattern).c_str(), ImGuiTreeNodeFlags_SpanFullWidth | ImGuiTreeNodeFlags_Leaf | ImGuiTreeNodeFlags_NoTreePushOnOpen);
+                    retVal = ImGui::TreeNodeEx(this->getDisplayName(pattern).c_str(), ImGuiTreeNodeFlags_SpanFullWidth | ImGuiTreeNodeFlags_Leaf | ImGuiTreeNodeFlags_NoTreePushOnOpen);
+                    break;
             }
+
+            if (!ImGui::IsItemToggledOpen() && ImGui::IsItemClicked()) {
+                m_selectionCallback(&pattern);
```
shewitt-au, line 475: This code is copied form `PatternDrawer::makeSelectable`.  It probably should be moved into a member function which would be called at both sites.

shewitt-au:
Do you mean that if hovering over the arrow, only the arrow should be selected, and similarly for name cell text?

paxcut:
Because you are changing the behavior of what it means to click in certain places you need to also change how the app reacts to the hovering of places. 
In the code before your changes, hovering over the name filed highlighted the entire name because clicking it would open the node. In contrast, hovering over other columns would highlight the entire line because clicking would select the entry in the row.
Fast forward to the new code and now hovering over the name should highlight the action that will be taken when clicking it. Accordingly the entire line minus the arrow should be highlighted. when hovering over the arrow then only the arrow should be highlighted because clicking it opens the node.

shewitt-au:
Cursor

The arrow shows the cursor. What I can interact with, as things stand before this PR, in this case, is not what can be interacted with. It's the "entity" pointed at.

paxcut:
the entire row is broken into two mutually exclusive sets. Each set is the places where clicking achieves a certain action.
Before the pr:  The two sets were the name column and the rest of the line. so hovering over any part of them highlights the entire set.
Currently: the two sets are the arrow and the rest of the line.

paxcut:
the only rule that must be followed is that when you hover over a portion of the user interface that can be interacted by clicking to achieve some result, then you must highlight the entire set of places where clicking achieves the same result.

paxcut:
That way you are teaching users how to interact with the ui without having to spell every interaction

shewitt-au:
Thers's logic to that.

paxcut:
it is also a universal theme across all apps. when you hover over any field the part that highlights is always the one that clicking anywhere on the highlighted part results on the same operation. Here is an example that is well known.

https://github.com/user-attachments/assets/16002e75-41d3-4419-9074-40602f4c80f9

shewitt-au:
OK. I think I’ve fixed the hover highlighting issue. Dear ImGui put up a fight. The solution is not as clean as I’d like, but my options were limited. It seems to work well. There are issues I’ve come across, but they are preexisting. For example, a TreeNode are seems to are a small border area around it that affects hit detection in edge cases. The aggregator nodes, when selected, don’t highlight the subset but instead the entire aggregated pattern,

shewitt-au:
Seems to have been replicated in a more elegant way.

WerWolv:
Sorry I checked out your commit and played around with some things but wasn't really done yet.
I then started working on other stuff and that code accidentally ended up in there. I wasn't trying to make you work for nothing, the code is still pretty much entirely yours

shewitt-au:
@WerWolv 
"the code is still pretty much entirely yours"

Nah, it's not man. The idea was mine, but your code is way more elegant than mine was.

shewitt-au:
@WerWolv 
"the code is still pretty much entirely yours"

Nah, it's not man. The idea was mine, but your code is way more elegant than mine was.

You were right to rewrite it.

PR #2284, opened 2025-05-31
issue-2282: Client changes

shewitt-au:

### DIFF
--- a/plugins/builtin/source/content/pl_builtin_types.cpp
+++ b/plugins/builtin/source/content/pl_builtin_types.cpp
@@ -27,7 +27,7 @@ namespace hex::plugin::builtin {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternEncodedString(*this));
+            return std::shared_ptr<Pattern>(new PatternEncodedString(*this));
         }
 
         [[nodiscard]] std::string getFormattedName() const override {
--- a/plugins/builtin/source/content/views/view_pattern_editor.cpp
+++ b/plugins/builtin/source/content/views/view_pattern_editor.cpp
@@ -1735,7 +1735,7 @@ namespace hex::plugin::builtin {
                         const pl::ptrn::Pattern *entry = pattern;
                         while (entry != nullptr) {
                             pathSegments.push_back(entry->getVariableName());
-                            entry = entry->getParent();
+                            entry = entry->getParent().get();
                         }
 
                         for (const auto &segment : pathSegments | std::views::reverse) {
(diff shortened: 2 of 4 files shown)


paxcut:
I tested the msi installer and it does seem to fix the problem. I ran the test case that created the crash and now I can both open the nodes and hover over the array without crashes or issues

shewitt-au:
Ok. I have updated both PRs.

I rolled back a lot of the code I was working on as it was getting too error prone. I have added a [utility](https://github.com/shewitt-au/std-enable_shared_from_this-and-constructors) to pattern_language to make the two stage creation easier and to allows gradual/on-demand adoption: `lib/include/pl/helpers/construct_shared_object.hpp`

I have not patched any of the patterns to use two stage creation yet.

Another issue is I suspect we’ll get `std::shared_ptr` cycles, and thus memory leaks, because the parent pointer is a` std::shared_ptr` and one can only assume the parent has `std::shared_ptr`'s to its children. I think the parent pointer should be a weak_ptr.

Builds fine for me. The client side GitHub tests fail however. Not sure why. Perphaps it's building without the matching pattern_language changes?

shewitt-au:
OK. Seems good for me so far. I've added enough two-stage constructions to fix all the bugs I've encountered.  I’m sure there’s more. I've also made the parent pointer weak.  A framework is (hopefully) in place to easily fix issues as they arise.

paxcut:
I wanted to test your changes but your builds are failing. The build includes both the ImHex and pattern language changes and the first error can be traced bask to this [line](https://github.com/WerWolv/PatternLanguage/blob/3ec6ae20ac2b5ded5829eefaa660f29b67ff1229/lib/source/pl/core/evaluator.cpp#L355) in your Imhex repository.
```cpp
    std::shared_ptr<construct_shared_object<pl::ptrn::Pattern> Evaluator>Variable(const std::string &name, const ast::ASTNodeTypeDecl *type, const std::optional<Token::Literal> &value, bool outVariable, bool reference, bool templateVariable, bool constant) {
 ```
 I am not sure where that comes from but if you follow the links in your Imhex repository for the pattern language linked there that what you'll find in line 355. I don't see any commits that have that change so you must be linking some other commit as a submodule.

shewitt-au:
@paxcut 
I put in a fix. Tests running now. I'm confused though. "so you must be linking some other commit as a submodule." How do I check for that?

shewitt-au:
@paxcut 

Failed again. Had another crack. Tests running. They are not failng so quickly. Looks like you were spot on. Thanks for pointing out the root cause. With my git knowledge I would not have figured that out.

shewitt-au:
Ok. Only two tests failed, and they don't look like they are caused by my changes.

shewitt-au:
Ok. This PR (and its pattern-side brethren) is now close to where I want it to be for now.

The two-stage construction may be contentious, and I get that, but it was used in four pattens to fix (hopefully) real bugs. 

`<pl/helpers/construct_shared_object.hpp>` is not intended to be an end onto itself. But, for now at least, it has served its purpose. And, to be honest, it was interesting to write. I have never used C++ concepts before. Even from this brief foray I’m blown away.

Firstly, it is a debugging tool. It has made it easier for me to update `Pattern` code which was causing bugs due to using `shared_from_this` in constructors. Assuming all the creation code has been updated to use `construct_shared_object` I can make the construction two-stage by simply adding a `post_construct` method.

It allowed me to find `Pattern` creation that bypasses `construct_shared_object` by making the `Pattern`’s constructions `protected` and using the `BEFRIEND_SHARED_OBJECT_CREATOR` macro to allow `construct_shared_object` to still create it. This will not pick up such bypasses in a Pattern class’s code itself however.

I think perhaps we need think about using the factory patten for `Pattern` creation in future.

But, that said, I’ve not has any crashes with importing patterns and associated activity with these changes. I'm sure someone will rectity that.

PR #2288, opened 2025-06-02
fix: Added protection against simlinks in the tasks.cpp::deleteOldFil…

1kurs1:
This PR improves the safety and reliability of the deleteOldFiles() function by:

Preventing symlink attacks

Uses std::filesystem::canonical() to resolve all paths, eliminating ./.. and symlinks
Verifies each file is actually inside the target directory before deletion
Safer file operations
Explicitly skips non-regular files (directories/symlinks)


WerWolv:
Thank you!

PR #2297, opened 2025-06-06
fix: fixes for various errors related to shared_from_this.

paxcut:
A while back there were some changes to the pattern language library that changed the way shared_pointers are created using shared_from_this(). Unfortunatelly the changes were not complete and various bugs were created among them 2234, json type not working, unable to export files, static arrays of bitfields,... The cause of the errors was that in class Pattern the member m_parent was left as a raw pointer and it needs to be handled by shared pointers. Also there were some cases in which share pointers were needed but unique pointers were used instead. Both cause crashes when shared_from_this is used on pointers that are not managed by shared_ptr. Another source of errors were infinite loops of clone and reference that caused stack overflow. The fixes include making m_parent a weak pointer, turning unique pointers into shared pointers and moving codefrom the copy constructors into clone to break the infinite loops.These changes are the bare minimum needed to bring the pattern language back to the full functionality that it had before shared_from_this was introduced or at least thats the hope.

### DIFF
--- a/plugins/builtin/source/content/pl_builtin_types.cpp
+++ b/plugins/builtin/source/content/pl_builtin_types.cpp
@@ -27,7 +27,8 @@ namespace hex::plugin::builtin {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternEncodedString(*this));
+            return std::make_shared<PatternEncodedString>(*this);
+           // return std::unique_ptr<Pattern>(new PatternEncodedString(*this));
         }
 
         [[nodiscard]] std::string getFormattedName() const override {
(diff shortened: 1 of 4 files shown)


paxcut:
these and the associated pattern language changes are the final changes for this pr. It aims at fixing the following bugs:

1)  crashes when using static arrays of bitfields. The problem was that shared_from_this messes up dynamic cast  of getParent result and makes it segv. The only way to fix it that I can find is to make m_parent a weak pointer which changes getParent and setParent. For getParent the return value is kept the same so that no further modifications are necessary.  More about  SetParent later.

2)  various ways to get invalid weak pointers. The easiest one to fix are the variables being set to unique pointers that try to use shared_from_this(). Not all unique pointers were made shared because we might need to use unique pointers too and the ones that werent changed don't seem to cause any problems. Another case of bad weak pointers are the ones where shared_from_this is called from a constructor.  There are 4 or 5 copy constructors that call setParent for each of the children. SetParent needs to use a reference to the parent to update the children, but this cannot happen inside the constructor because reference is just a call to shared_from_this. Therefore the problem can be solved by taking setParent calls out of the constructors and moving them into clone() which is the only place that calls the copy constructors directly.

With those two set of changes all the problems that were being reported like exporting patterns and custom inspector functions seems to run without problems. There are still some issues with references and attributes returning wrong values, but that's another pr.

PR #2298, opened 2025-06-07
MbedTLS update requirement

winterheart:
### Problem description

MbedTLS 3.6 (LTS) includes cmake configuration modules that simplifies integration with it and makes custom find module redundant.


iTrooz:
@WerWolv Do you have anything against this being merged, on principle ?

If not, I'd like to take care of it

iTrooz:
Unfortunately this is a recent version, which is not supported on all platforms that we target (e.g. Ubuntu 24.04), so closing

Maybe in a few years !

PR #2325, opened 2025-07-10
Update ImGui to v1.92.1

WerWolv:
Update ImGui to the latest version and rewrite most of the font handling code


lib/libimhex/source/ui/imgui_imhex_extensions.cpp
```diff
@@ -1019,7 +1019,7 @@ namespace ImGuiExt {
         std::string drawString;
         auto textEnd = text + strlen(text);
         for (auto wrapPos = text; wrapPos != textEnd;) {
-            wrapPos = ImGui::GetFont()->CalcWordWrapPositionA(1, wrapPos, textEnd, availableSpace.x * 0.8F);
+            wrapPos = ImGui::GetFont()->CalcWordWrapPosition(1, wrapPos, textEnd, availableSpace.x * 0.8F);
```
ocornut, line 1022: This is incorrect FYI

API breaking changelog says
```cpp
- Fonts: changed ImFont::CalcWordWrapPositionA() to ImFont::CalcWordWrapPosition():
   - old: const char* CalcWordWrapPositionA(float scale, const char* text, ....);
   - new: const char* CalcWordWrapPosition (float size,  const char* text, ....);
```
Notice the first parameter. Should probably be `GetFontSize()`.

WerWolv: Oh yes you're right, thanks a lot!

Issue #2330, opened 2025-07-13
Typo in screenshot_descriptions.json


levelad:
There is a typo in "ImHex / plugins / builtin / romfs / assets / screenshot_descriptions.json"

I don't want to fork it just for this tiny change.

"... a audio player..." -> "... a**n** audio player..."

See https://www.grammar.com/a-vs-an-when-to-use

PR #2333, opened 2025-07-14
Fix 2330

C3pa:
Fix a typo in `screenshot_descriptions.json` reported in #2330. While I was presented with the contributing doc, I came across another typo that's fixed in the second commit.


WerWolv:
Thank you!

PR #2353, opened 2025-07-24
impr: scroll to beginning of pattern when clicked, not the end

lawm:
When a pattern is larger than the visible area, the Hex editor will scroll to the end of the pattern when it is clicked. This change makes it scroll to the beginning instead, which is more intuitive for users who expect to see the start of the pattern when they click on it.

### Problem description
When a pattern item is larger than the Hex editor visible area (if pattern item is 40 rows, but Hex editor is 20 rows tall), it scrolls to the end of the pattern item.  I expect to see the beginning of the pattern item.

Change cursor position to m_selectionStart instead of the end in setSelection().

### Additional things
I don't know what else this will affect.


lawm:
This doesn't work well. 
jump pivot also needs to be updated

PR #2377, opened 2025-08-04
feat: Add EventViewClosed event

aoqia194:
Where one door opens, another one closes (said someone) and in this situation i mean view, because there's no event for closing a view only opening one, so please add this :c

### Problem description
Simply adds an **EventViewClosed** to match the pre-existing **EventViewOpened**. Also fixes the issue where EventViewOpened wouldn't trigger for windows that were popups (even if they are technically a View.

I mirrored the current functionality of EventViewOpened for closing.


WerWolv:
Thanks!

Issue #2233, opened 2025-05-09
[Bug] Cannot start on FreeBSD


alexax66:
### Operating System

FreeBSD 14.2

### What's the issue you encountered?

I successful built ImHex on FreeBSD, but it can't start:
```
[14:05:23] [INFO]  [main | Main]               Welcome to ImHex 1.38.0.WIP-Debug!
[14:05:23] [INFO]  [main | Main]               Compiled using commit master@d33fad9
[14:05:23] [INFO]  [main | Main]               Running on FreeBSD 14.2-RELEASE-p2 FreeBSD 14.2-RELEASE-p2 n230744-26127ca9fef GENERIC (amd64)
[14:05:23] [INFO]  [main | Main]               FreeBSD distribution: GhostBSD 25.01-R14.2p2. Version: 25.01
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  18 (X_ChangeProperty)
  Resource id in failed request:  0x5800007
  Serial number of failed request:  123
  Current serial number in output stream:  132

```

### How can the issue be reproduced?

Need some fixes for building:

1. ImHex/lib/external/libwolv/libs/io/source/io/file_unix.cpp

```
diff --git a/libs/io/source/io/file_unix.cpp b/libs/io/source/io/file_unix.cpp
index c15fe79..a6467f4 100644
--- a/libs/io/source/io/file_unix.cpp
+++ b/libs/io/source/io/file_unix.cpp
@@ -8,6 +8,10 @@
 #if defined(OS_MACOS)
     #include <sys/types.h>
     #include <sys/event.h>
+#elif defined(OS_FREEBSD)
+    #include <sys/types.h>
+    #include <sys/event.h>
+    #include <poll.h>
 #elif defined(OS_LINUX)
     #include <sys/types.h>
     #include <sys/inotify.h>
@@ -194,7 +198,7 @@ namespace wolv::io {
     }


-    #if defined(OS_MACOS)
+    #if defined(OS_MACOS) || defined(OS_FREEBSD)
         void ChangeTracker::trackImpl(const bool &stopped, const std::fs::path &path, const std::function<void()> &callback) {
             int queue = kqueue();
             if (queue == -1)

```
 

2. ImHex/main/gui/source/messaging/linux.cpp

```
diff --git a/main/gui/source/messaging/linux.cpp b/main/gui/source/messaging/linux.cpp
index a25c0f5b..44ec2a87 100644
[...]

Issue #2215, opened 2025-04-14
[Bug] X Error of failed request:  BadWindow (invalid Window parameter)


marinjurjevic:
### Operating System

Linux

### What's the issue you encountered?

I've installed package from AUR. Tried to run it and I get following:

```shell
$> imhex                                                                                                                                                                                                       
[22:06:22] [INFO]  [main | Main]               Welcome to ImHex 1.37.4!
[22:06:22] [INFO]  [main | Main]               Compiled using commit master@ed24964
[22:06:22] [INFO]  [main | Main]               Running on Linux 5.15.179-1-MANJARO #1 SMP PREEMPT Thu Mar 13 13:45:22 UTC 2025 (x86_64)
[22:06:22] [INFO]  [main | Main]               Linux distribution: Manjaro Linux. Version: None
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  18 (X_ChangeProperty)
  Resource id in failed request:  0x5400007
  Serial number of failed request:  129
  Current serial number in output stream:  138

```

### How can the issue be reproduced?

Launch from terminal or launcher

### ImHex Version

 1.37.4

### ImHex Build Type

- [x] Nightly or built from sources

### Installation type

AUR package (1.37.4-1)

### Additional context?

I'm using manjaro with i3 window manager. I've run ImGui applications in the past without issue so I'm not sure if this is related to ImGui or ImHex

screenfetch:
```shell

 ██████████████████  ████████ 
 ██████████████████  ████████     OS: Manjaro 25.0.0 Zetar
 ██████████████████  ████████     Kernel: x86_64 Linux 5.15.179-1-MANJARO
 ██████████████████  ████████     Uptime: 1h 29m
 ████████            ████████     Packages: 2112
 ████████  ████████  ████████     Shell: zsh 5.9
 ████████  ████████  ████████     Resolution: 4920x1920
 ████████  ████████  ████████     WM: i3
 ████████  ████████  ████████     GTK Theme: Nordic [GTK2/3]
 ████████  ████████  ████████     Icon Theme: Papirus-Dark
 ████████  ████████  ████████     Font: Noto Sans 10
 ████████  ████████  ████████     Disk: 1.9T / 3.1T (60%)
 ████████  ████████  ████████     CPU: Intel Core i7-7700K @ 8x 4.5GHz [40.0°C]
 ████████  ████████  ████████     GPU: NVIDIA GeForce GTX 1070
[...]

PR #2390, opened 2025-08-08
fix: Initialise splash window with correct size

iwanders:
With an initial Window size of 1 by 1, the program aborts during launch on my Debian 12 (bookworm) system running the Mate Desktop environment on X11.

The exception from X is the following:
```
$ ./imhex
[18:36:33] [INFO]  [main | Main]                Welcome to ImHex 1.38.0.WIP!
[18:36:33] [INFO]  [main | Main]                Compiled using commit master@59c6d30
[18:36:33] [INFO]  [main | Main]                Running on Linux 6.1.0-37-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.140-1 (2025-05-22) (x86_64)
[18:36:33] [INFO]  [main | Main]                Linux distribution: Debian GNU/Linux 12 (bookworm). Version: 12
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  18 (X_ChangeProperty)
  Resource id in failed request:  0x5400009
  Serial number of failed request:  152
  Current serial number in output stream:  163
```

The culprit is in `hex::init::WindowSplash::initGLFW()`, specifically the `glfwCreateWindow` call with a window size of 1 by 1, that's the last call in the ImHex codebase, after that it's libglfw and libX11.

The issue can be worked around by inhibiting the splash screen using the `IMHEX_SKIP_SPLASH_SCREEN=1` environment variable, setting the window to be a size larger than 1 by 1 also seems to resolve the situation.

My hunch is that an underflow occurs somewhere with the smaller values, but I did no further investigation into analysing the root cause.

----
Above the line is the commit message.

### Problem description
See https://github.com/WerWolv/ImHex/issues/2215, which I think is the same problem, and commit description above. Edit; I did another search in the issues, I wouldn't be surprised if #2233 is a duplicate of the 2215 issue.

Please see the commit message.

Analysis was done by setting an X11 error handler, making that throw and running it in a debugger. This is available in this dev commit; 4a24e16a2ef100a9417e572f6b16c4a95711e94e.

### Screenshots
No apparent glitches if I make the splash window stick around with `return std::nullopt;` in the `loop()` function. I don't know if there's an issue at the beginning, but given that the window is set to transparent before it is created and resized I don't expect any.

Here's a compulsory screenshot of the splash screen over an editor;

image

Yes! I don't use ImHex often, but it's really useful when I do, thank you!

### DIFF
--- a/main/gui/source/init/splash_window.cpp
+++ b/main/gui/source/init/splash_window.cpp
@@ -470,7 +470,7 @@ namespace hex::init {
         glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_API);
 
         // Create the splash screen window
-        m_window = glfwCreateWindow(1, 1, "Starting ImHex...", nullptr, nullptr);
+        m_window = glfwCreateWindow(WindowSize.x, WindowSize.y, "Starting ImHex...", nullptr, nullptr);
         if (m_window == nullptr) {
             hex::nativeErrorMessage(fmt::format(
                 "Failed to create GLFW window: [{}] {}.\n"


WerWolv:
Thank you! 
Interesting how it breaks with lower values, might be worth reporting to either the GLFW or the X11 people if it's a problem on their side. Glad the workaround for it is so simple though, I don't think this should cause any trouble here

iwanders:
> might be worth reporting to either the GLFW or the X11 people if it's a problem on their side

That's a good idea, I think it is in GLFW and filed https://github.com/glfw/glfw/pull/2754 for it.

I wonder if something changed recently around the scaling factors on more recent installations, I got;
```
width: 1, height: 1
_glfw.x11.contentScaleX: 0.958333, _glfw.x11.contentScaleY: 0.958333
width: 0, height: 0
```
So only after the multiplication of the scaling factor did it change to be invalid dimensions.

PR #2440, opened 2025-09-10
Rework command line argument parsing

Zman350x:
The current CLI argument parsing did not work as documented and had a number of issues related to multi-flag (subcommand) parsing. I've reworked the logic in such a way that should maintain full compatibility with any existing scripts/use-cases but with added functionality.


WerWolv:
Pushed some small style fixes but the PR as a whole looks very clean to me and works great. Thank you very much!

PR #2445, opened 2025-09-13
Add keyboard shortcuts to jump to the prev/next differences in diff view

Zman350x:
Adds keyboard shortcuts (currently `n` and `N`) in the diffing plugin view to jump to the next/prev difference in the list.

IMPORTANT NOTE: Depends on changes made in a library submodule. [This PR](https://github.com/WerWolv/libwolv/pull/34) must be accepted first.

### DIFF
--- a/lib/third_party/imgui/backend/CMakeLists.txt
+++ b/lib/third_party/imgui/backend/CMakeLists.txt
@@ -19,6 +19,12 @@ if (NOT IMHEX_EXTERNAL_PLUGIN_BUILD)
     find_package(OpenGL REQUIRED)
     find_package(Freetype REQUIRED)
 
+    if (UNIX AND NOT APPLE AND NOT EMSCRIPTEN)
+        find_package(X11)
+        target_include_directories(imgui_backend PUBLIC ${X11_INCLUDE_DIR})
+        target_link_libraries(imgui_backend PUBLIC ${X11_LIBRARIES})
+    endif()
+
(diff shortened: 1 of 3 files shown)


WerWolv:
Looks good to me! Thank you

Some things I noticed:
- The shortcuts (N and SHIFT + N) are currently the only ones in ImHex that don't use some sort of modifier (CTRL or ALT) by default. I know this comes from from Vim and various other Linux tools and feels great if you're used to these kind of shortcuts. Though for ImHex I think it would make more sense to have something like CTRL + Left/Right or similar. They can of course still be remapped through the settings
- There's currently no menu entries for these shortcuts so they're not super easy to discover unless you know about that functionality. Maybe we should add them as view-specific menu entries to the File menu. Thoughts on that?
- When going forward or backward onto a change that's only one byte long using Myer's diffing algorithm, too many bytes are being selected sometimes. Not sure if that's an issue with the algorithm or your code though

Zman350x:
Default keybinds changed (to CTRL+left/right), and it's all been added to menus, as opposed to just being shortcuts.
As for the Meyer stuff, see my message in Discord.
All lgtm, but let me know if there's anything else you want changed!

WerWolv:
Thank you! Merged the libwolv PR

Zman350x:
Submodule pointed at correct libwolv commit. Ready to merge!

WerWolv:
Thanks a lot again!

PR #2449, opened 2025-09-14
impr: Allow up to 16384 filter items

nebkat:
### Problem description
We use ImHex to read a custom data logging format that contains multiple event types and need the ability to filter them in bulk (for visualizing).

This limit was introduced in https://github.com/WerWolv/ImHex/commit/126868c251c6a6b5b86d11331710a2c7f1a854c7, I presume as a reasonable medium value, but at least on our machines there are seemingly no performance issues even at 1024 limit.


WerWolv:
Thanks for the PR!

The limit is there because we can't clip / optimize the display because it's a tree. So depending on what kind of formatting you do for your types, it will quickly get very slow. 
That said, you can override pretty much any slider settings and set them outside of their ranges by CTRL-Clicking on the slider and entering a value manually - including this one. I believe that would be a better way here to not send the wrong message to "regular" users that will expect the entire range to work decently.

Let me know if that works for you. Otherwise we can evaluate increasing the upper value but I'd like to make sure that even the max value works well for the majority of use-cases

nebkat:
Oh perfect, agreed re default, thanks for the quick response!

PR #2473, opened 2025-10-02
fix(cli): add test and fix for argv readonly handling

Zenvila:
## Summary
This PR introduces a failing test for the `argv` bug related to readonly handling in the CLI,
and provides a fix to ensure proper argument parsing.

## Changes
- Added new test case in `tests/cli/test_cli_readonly.cpp` to reproduce the bug.
- Implemented fix in CLI initialization to correctly handle readonly `argv`.
- Updated CMake configuration to include the new test.

## Motivation
Previously, the CLI would incorrectly handle readonly cases with `argv`, 
leading to unexpected behavior. This PR ensures that the issue is reproducible 
via a test and then properly resolved in the code.

## Testing
- Verified the new test fails before the fix.
- Confirmed the test passes after the fix.
- All existing tests still pass.

## Related
Fixes #<issue-number-if-any>

### DIFF
--- a/lib/libimhex/CMakeLists.txt
+++ b/lib/libimhex/CMakeLists.txt
@@ -57,6 +57,7 @@ set(LIBIMHEX_SOURCES
         source/ui/banner.cpp
 
         source/subcommands/subcommands.cpp
+        source/init/cli.cpp
 )
 
 if (APPLE)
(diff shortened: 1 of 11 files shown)


Zenvila:
Kindly review the request

Zman350x:
Is there a reason this CLI flag was given its own parser instead of just being registered as a subcommand the same way every other CLI flag is handled? And unless I'm reading your PR wrong, you're implementing a whole new feature here (read-only mode), not just adding a test for an existing feature, so why is this marked as a test/fix instead of a feature?

PR #2492, opened 2025-10-30
fix: add newly created projects to "Recent"

iTrooz:
### Problem description

Projects weren't being saved as recent when a new project was saved. They were only added as recent when re-opening the project

I also save projects as recent when saving them (I don't make a difference between saving existing and new projects)


paxcut (approved):
Thank you for this. it was bothering me and had it in my todo list for a long time. lease merge when you are ready

PR #2493, opened 2025-10-31
Feature: add hooks to let Views to get notified when they are opened or closed

davidgiven:
This is a trivial change which adds virtual methods to View, `onOpen()` and `onClose()`, which are called when the view is opened or closed.

This information is already tracked inside the View, but not exposed via the API. There is `didWindowJustOpen()` and `didWindowJustClose()`, but these fetch and then reset the flag, so they can't be used more than once in a frame (and are sometimes called by the frame, meaning the flag has already been consumed by the time the View's draw callback gets called).

The use case here is that I have a View which needs to do some work every time it's shown.


davidgiven:
Thanks!

PR #2533, opened 2025-12-03
Fix out-of-bounds access in UTF-8 substring extraction for Line parts

AkiSakurai:
This can happen during undo/redo if there are wide characters in the pattern editor.

### DIFF
--- a/plugins/ui/source/ui/text_editor/support.cpp
+++ b/plugins/ui/source/ui/text_editor/support.cpp
@@ -217,11 +217,11 @@ namespace hex::ui {
             return m_flags.substr(start, length);
         if (part == LinePart::Utf8) {
             u64 utf8Start = 0;
-            for (u64 utf8Index = 0; utf8Index < start; ++utf8Index) {
+            for (u64 utf8Index = 0; utf8Index < start && utf8Start < m_chars.size(); ++utf8Index) {
                 utf8Start += TextEditor::utf8CharLength(m_chars[utf8Start]);
             }
             u64 utf8Length = 0;
-            for (u64 utf8Index = 0; utf8Index < length; ++utf8Index) {
+            for (u64 utf8Index = 0; utf8Index < length && utf8Start + utf8Length < m_chars.size(); ++utf8Index) {
                 utf8Length += TextEditor::utf8CharLength(m_chars[utf8Start + utf8Length]);
             }
             return m_chars.substr(utf8Start, utf8Length);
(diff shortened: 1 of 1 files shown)


paxcut:
Do you have a way to reproduce the problem? I tried using the following comment which contains 1,2 and 3 byte chars
```cpp
// 解析 codesys 通讯协议， 协议数据通过 CANopenMagicUltimate 采集。
```
removing half of it and undoing/redoing didn't cause any problems. As long as the strings are properly encoded the code should work as is. Is it possible that you are using  improperly encoded or non-utf8 encoded strings? Is this causing a crash? Currently the utf8 support is not very robust and has may not be handling invalid sequences properly.

AkiSakurai:
⌥a ⌥a ⌥a del del ⌘z ⌘z
I just tried again (v1.37). It didn't happen on the stable version, but on the [nightly version](https://github.com/WerWolv/ImHex/releases/tag/nightly) (v1.38), it crashes. I am running on macOS.

WerWolv:
Thank you!
I can try on my mac tonight if there's something specific there

paxcut:
I dont have a mac so i don't know if this applies to windows too. I copy and pasted what you posted and deleted it and undo redo didn't crash. Pasting the chars on hex editor shows that some are multi-byte and using utf8 decoding column shows the same characters shown on your post so i think it may be a mac os thing. Did you try the strings I posted? 
image

AkiSakurai:
Sorry, I did not express it clearly enough. The crash happens in the pattern script editor, not the hex editor. The sequence I posted is a physical keystroke sequence to reproduce the problem: first type 3 "å" characters, then delete 2 characters, then undo twice, and then it will crash.

paxcut:
Yes, I am testing the pattern editor for crashes and using the hex editor to identify the encodings. I'll try with the instructions but i have no way to type utff9 chars directly so i have to paste them instead.

paxcut:
The code for that utf8 glyph is 0xC3 0xA5. I pasted 3 of them (no spaces) in the pattern editor, deleted the last two using backspace and ctrl-z twice to undo and no crash occurred.

paxcut:
It isn't crashing but not because there is no problems, it just doesn't care about the array bounds on windows debug versions. I did find the problem and it is due to using the wrong metric to calculate the substring ends. At the second undo m_chars is a 2 character, 4 byte string. The substring command asks for subs(2,-1,utf8) which means start at the third utf8 char and select until the end if the string. so the substring is indeed empty or it should be. Inside the substring code you'll see this
```cpp
        if (length == (u64) -1 || start + length >= m_chars.size())
            length = m_chars.size() - start;
```
so length which is supposed to become zero uses size() which is 4 and then the error occurs. The use of -1 to select the end of the string needs to be done differently for utf8 and for chars and that's really what the problem is.

paxcut:
This is what substr looks like with the fixes. I had to also create a cont maxColumn function so i could use it here:
There were similar problems in several other places that I also fixed. Basically you can use length() or size() when counting utf8 chars. in that case you need to use maxColumn() which is cached for speed.

```cpp
 std::string Line::substr(u64 start, u64 length, LinePart part) const {
       
        if (part != LinePart::Utf8) {
            if (start >= m_chars.size() || m_colors.size() != m_chars.size() || m_flags.size() != m_chars.size())
                return "";
            if (length == (u64) -1 || start + length >= m_chars.size())
                length = m_chars.size() - start;
            if (length == 0)
                return "";

            if (part == LinePart::Chars)
                return m_chars.substr(start, length);
            if (part == LinePart::Colors)
                return m_colors.substr(start, length);
            if (part == LinePart::Flags)
                return m_flags.substr(start, length);
        } else {
            if (start >= (u64) maxColumn())
                return "";
            if (length == (u64) -1 || start + length >= (u64) maxColumn())
                length = maxColumn() - start;
            if (length == 0)
                return "";

            u64 utf8Start = 0;
            for (u64 utf8Index = 0; utf8Index < start; ++utf8Index) 
                utf8Start += TextEditor::utf8CharLength(m_chars[utf8Start]);
            
            u64 utf8Length = 0;
            for (u64 utf8Index = 0; utf8Index < length; ++utf8Index) 
                utf8Length += TextEditor::utf8CharLength(m_chars[utf8Start + utf8Length]);
            
            return m_chars.substr(utf8Start, utf8Length);
        }
        return "";
    }
```

AkiSakurai:
Close, as there is a better solution.

PR #2597, opened 2026-01-01
build: add IMHEX_ENABLE_UPDATER option

xtexx:
### Problem description

- Many Linux distributions don't like self-updaters.
- The updater does not support all architectures and fails builds on archs like PowerPC, LoongArch, RISC-V, etc.

Add an `IMHEX_ENABLE_UPDATER` option to control that.


WerWolv:
Hey!

The reason why I didn't make it configurable like that is because I believe this isn't a decision a package maintainer should be making for their users but users should do it themselves (by disabling the setting).

I'd be more inclined to just make the updater build and return a proper error message on architectures that aren't supported.

Additionally, just making the updater not be built will still make the update button show up in the UI, clicking on it will just not do anything though which I don't think is great UX

xtexx:
Hey!

I still prefer an option to disable updating.

Users who install ImHex through system package managers should always update ImHex through the package manager. For distributions, self-updating leads to inconsistency between package databases and actual files. For distributions whose `/usr` is read-only (or cannot be written by normal users), updating always fails.

The updated version:
- Added an option to disable automatic updating.
- Make `imhex-updater` build on unsupported architectures.

Issue #2592, opened 2025-12-30
[Feature] Format time based on locale


eirnym:
### What feature would you like to see?

Please support system locale and possibly add settings to override time formatting.

### How will this feature be useful to you and others?

I have Polish locale and I don't (and I don't want to) understand US time formatting. 

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

WerWolv:
Hey! Which time formatting are you talking about exactly? The only one that's fixed I believe is the one in the data inspector which uses pretty standard formatting I believe.

eirnym:
That's the only time field formatted in the default UI.

WerWolv:
How does the Polish one differ there? I feel `Weekday, Day.Month.Year Hour:Minute:Second` is pretty standard and not US time formatting.

The alternative I would be open to consider is ISO 8601 format in the form of `2025-12-30T17:56:20Z` though I'm not sure that's really _more_ readable than what we currently have

eirnym:
date(1) printed `wto 30 gru 22:35:02 2025 CET`. But for a hex editor  ISO 8601 is the best neutral alternative.

shewitt-au:
I've been looking at this. MingW64's `std::locale` class is broken on Windows. A simple call to `std::locale("")` raises an exception. It tries to find POSIX-style locale names on Windows. I could hack around it, or we could use Boost.Locale.

eirnym:
Empty string is not a valid locale. Use `C` instead.

shewitt-au:
> Empty string is not a valid locale. Use `C` instead.

ISO C++ working draft N2356 for locales:
> explicit locale(const char* std_name);
> Effects: Constructs a locale using standard C locale names…
> Notes: The set of valid string argument values is "C", "", and any implementation‑defined values.

Somewhat old, but I believe still the case.

But it's a moot point. MingW support for locales is limited to say the least.

eirnym:
ISO-8601 format is the best alternative for this, and strftime (and all any other ways to format it) supports it in one way or another. It will be way better than US format anyway for people who don't live there.

shewitt-au:
The code I've been working on gives this for Polish:

<img width="1004" height="159" alt="Image" src="https://github.com/user-attachments/assets/f2a67def-911a-4b26-b28c-59135ce42e84" />

This is from Windows.

And Chinese:

[...]

PR #2677, opened 2026-03-09
F2592 locale date time

shewitt-au:
### Problem description
Dates and times currently have a fixed format and do not respect the selected language. MinGW’s (Windows) `std::locale` class is broken and basically useless. This PR implements, along with the libwolv PR linked below, date and time formatting using the Win32 API on Windows and the C runtime library on other OSes.

The motivating feature request can be found here:
[Format time based on locale #2592](https://github.com/WerWolv/ImHex/issues/2592
)

The corresponding libwolv changes can be found here:
[libwolv](https://github.com/WerWolv/libwolv/pull/37)

The corresponding PatternLanguage changes can be found here:
[PatternLanguage](https://github.com/WerWolv/PatternLanguage/pull/227/changes)

The corresponding ImHex-Patterns changes:
[ImHex-Patterns](https://github.com/WerWolv/ImHex-Patterns/pull/510)

For the most part, ImHex has been changed to use a new function in libwolv for date and time formatting. A notification was also added to update the Data Inspector when the language is changed. Some changes were made to fix issues in the existing code that make sense even without the broader changes:
- The time formatting code was using the local time zone instead of GMT.
- `time_t` was being treated as unsigned, but it should be signed.
- When there was a formatting issue with a `time_t` "Invalid" was displayed. No `time_t` is invalid. The issue is with our ability to format it. I have changed the message to "Can't format" for now.
- time32_t was only present on Windows. Now it's always available.
- DOSDate & DOSTime not the correct size. That messed up endian conversions and created issues when they were placed at the end of a file.

### Linux
On Linux only dates and times for which a locale is installed (`sudo locale-gen`) will be formatted correctly.


shewitt-au:
@WerWolv 

I have added locale filtering as I had started but not finished it. The locale drop down is now filtered to only show locales  with the selected language. I do not plan on spending any more time on this. Seems pointless given the total lack of feedback.

Anyway, here's a (somewhat clumsy) demonstration:

https://github.com/user-attachments/assets/8183e1df-631e-4d21-99a5-ec31cc703b1f

PR #2698, opened 2026-04-04
fix: buffer slice last byte out of range

### LINKED ISSUE #1880: [Bug] Slice buffer node range excludes "To" index
### Operating System

Windows

The Slice buffer node outputs buffer addresses of [From:To-1] instead of [From:to]. Apart from not working as expected, this also makes it impossible to get the last byte from a buffer in a slice operation.

In the data processor create a Slice buffer node and pass a buffer through it. Observe that in the output the data starts at the From value index position, and ends one byte shy of the index for the To position. Also observe that it doesn't seem possible to get the last byte out of the slice regardless of values, since a "To" value greater than the length of a buffer errors out as out of range.

### ImHex Version

1.35.4

### ImHex Build Type

### Installation type

Windows install via MSI

### Additional context?

_No response_

3l0w:
### Problem description
Fixes: #1880
The slice buffer node 'to' variable cannot be set to capture the last byte, throwing `'to' input out of range`.

The upper bound check was an greater or equals than, so switching it to a greater than fixes the issue.

### DIFF
--- a/plugins/builtin/source/content/data_processor_nodes/other_nodes.cpp
+++ b/plugins/builtin/source/content/data_processor_nodes/other_nodes.cpp
@@ -193,7 +193,7 @@ namespace hex::plugin::builtin {
 
             if (from < 0 || static_cast<u128>(from) >= input.size())
                 throwNodeError("'from' input out of range");
-            if (to < 0 || static_cast<u128>(to) >= input.size())
+            if (to < 0 || static_cast<u128>(to) > input.size())
                 throwNodeError("'to' input out of range");
             if (to <= from)
                 throwNodeError("'to' input needs to be greater than 'from' input");


paxcut:
I guess we have two choices. We can use C-like indices of 0..N-1 to slice N elements. ImHex uses an iterator like system where 0..N selects the first N elements so that the end number is one past the end of the buffer.
The fix proposed here follows ImHex's model.To implement a C like index system use: 
```cpp
this->setBufferOnOutput(3, std::vector(input.begin() + u64(from), input.begin() + u64(to + 1)));
```
and leave the error condition for `to` as it was.

I think either one is fine but I wanted to make sure that the choice is deliberate and not accidental due to the bug.

PR #2718, opened 2026-04-21
Fix nullptr deref when opening ImHex without a provider on frame 1

neptuwunium:
### Problem description
Commit https://github.com/WerWolv/ImHex/commit/f813656e8ccfb2a5166ba824130b9bd7d210d3e5 introduced a nullptr dereference that happens on the first frame of a provider being opened.

```
[12:03:47] [INFO]  [main | Main]                Switching to workspace 'Default'
Process 89715 stopped
* thread #1, name = 'ImHex 🔍', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x0)
    frame #0: 0x00007fffcdd43958 ui.hexpluglib`::drawEditor() at hex_editor.cpp:648:46
   645 	                ImGui::TableSetupScrollFreeze(0, 2);
   646 	
   647 	                // Row address column
-> 648 	                u64 maxAddress = m_provider->getActualSize();
    	                                             ^
   649 	                if (maxAddress > 0)
   650 	                    maxAddress--;
   651 	                if ((m_scrollPosition + m_visibleRowCount) * bytesPerRow < maxAddress)
(lldb) bt
* thread #1, name = 'ImHex 🔍', stop reason = signal SIGSEGV: address not mapped to object (fault address=0x0)
  * frame #0: 0x00007fffcdd43958 ui.hexpluglib`::drawEditor() at hex_editor.cpp:648:46
    frame #1: 0x00007fffcdd4954b ui.hexpluglib`::draw() at hex_editor.cpp:1409:15
    frame #2: 0x00007fffdc01eb21 diffing.hexplug`::drawDiffColumn() at view_diff.cpp:64:30
    frame #3: 0x00007fffdc01d56d diffing.hexplug`::drawContent() at view_diff.cpp:271:28
    frame #4: 0x00007ffff7af7413 libimhex.so.1.39.0.WIP`::draw() at view.cpp:201:23
    frame #5: 0x0000555555593e7d imhex`::frame() at window.cpp:759:23
    frame #6: 0x000055555559237d imhex`hex::Window::fullFrame() at window.cpp:200:19
    frame #7: 0x00005555555948b5 imhex`::loop() at window.cpp:297:19
    frame #8: 0x00005555555c211c imhex`::runImHex() at desktop.cpp:54:28
    frame #9: 0x00005555555836a6 imhex`main at main.cpp:73:12
    frame #10: 0x00007ffff682f43b libc.so.6`___lldb_unnamed_symbol_2b3c0 + 123
    frame #11: 0x00007ffff682f4eb libc.so.6`__libc_start_main + 139
    frame #12: 0x00005555555831d5 imhex`_start + 37
```

If nullptr check guard.

### DIFF
--- a/plugins/ui/source/ui/hex_editor.cpp
+++ b/plugins/ui/source/ui/hex_editor.cpp
@@ -645,16 +645,20 @@ namespace hex::ui {
                 ImGui::TableSetupScrollFreeze(0, 2);
 
                 // Row address column
-                u64 maxAddress = m_provider->getActualSize();
-                if (maxAddress > 0)
-                    maxAddress--;
-                if ((m_scrollPosition + m_visibleRowCount) * bytesPerRow < maxAddress)
-                    maxAddress = (m_scrollPosition + m_visibleRowCount) * bytesPerRow;
+                u64 maxAddress = std::numeric_limits<u64>::max();
+                
+                if (m_provider != nullptr) {
+                    maxAddress = m_provider->getActualSize();
+                    if (maxAddress > 0)
+                        maxAddress--;
+                    if ((m_scrollPosition + m_visibleRowCount) * bytesPerRow < maxAddress)
+                        maxAddress = (m_scrollPosition + m_visibleRowCount) * bytesPerRow;
 
-                if (maxAddress + m_provider->getCurrentPageAddress() < std::numeric_limits<u64>::max() - m_provider->getBaseAddress())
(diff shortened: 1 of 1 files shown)


paxcut:
Nice catch. I can't believe I missed this one considering all the other checks for null providers elsewhere. Probably thought t was handled by the check above but it really isn't. I tested the fix locally and it fixed the crash reported in issue #2719 which happens to be a test case for this pr too. Unless there are objections i am ready to merge this.

PR #2720, opened 2026-04-22
Search for included patterns in user provided folders

neptuwunium:
### Problem description
Folders tab in settings specifies "Specify additional **search paths for patterns**, scripts, Yara rules and more"

However this is false. I don't think the argument is used for scripts or YARA rules, either. The vector is only read in getDataPaths, which is only accessed by getConfigPaths, and extractBundledFiles.

Also, relative paths do not work for file-backed patterns.

Added `ImHexApi::System::getAdditionalFolderPaths()` to the calls that looks for pattern paths.

### DIFF
--- a/lib/libimhex/source/api/content_registry.cpp
+++ b/lib/libimhex/source/api/content_registry.cpp
@@ -724,7 +724,7 @@ namespace hex {
                 );
             }
 
-            runtime.setIncludePaths(paths::PatternsInclude.read() | paths::Patterns.read());
+            runtime.setIncludePaths(paths::PatternsInclude.read() | paths::Patterns.read() | ImHexApi::System::getAdditionalFolderPaths());
 
             for (const auto &[ns, name, paramCount, callback, dangerous] : impl::getFunctions()) {
                 if (dangerous)
--- a/plugins/builtin/source/content/views/view_pattern_editor.cpp
+++ b/plugins/builtin/source/content/views/view_pattern_editor.cpp
@@ -2802,7 +2802,7 @@ namespace hex::plugin::builtin {
         auto provider = ImHexApi::Provider::get();
         if (provider == nullptr)
             return;
-        const auto basePaths = paths::Patterns.read();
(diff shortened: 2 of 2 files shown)


paxcut:
I don't have this problem. I added a folder in the folders path and in it I created a sub-folder called patterns. I then wrote a small test hexpat file which was the found when I used Open Patterns in ImHex. Checking under the debugger the folder added in the folders tab was listed along the other normal folders in basePaths  and all of the ended in a patterns sub-folder.   

Often people misinterpret the documentation thinking they can store patterns and other types of files directly in the folders added to the folders tab but reading it carefully does mention that each type of file needs a sub-folder of that name for Imhex to be able to find their contents.

neptuwunium:
> Often people misinterpret the documentation thinking they can store patterns and other types of files directly in the folders added to the folders tab but reading it carefully does mention that each type of file needs a sub-folder of that name for Imhex to be able to find their contents.

I think that's less an issue with the docs and more an issue with where the docs are. The description text does not indicate nor lead to the docs. (The docs site also just generally frustrating to look anything up because the search is slow while it immediately suggests using AI to search which is even more useless.)

Closing this, but the tooltip should be updated to reflect this behavior.

neptuwunium:
Actually I just quickly looked through the docs, and there's no mention of the folder system at all.

https://docs.werwolv.net/imhex/misc/settings <-- this is a stub page

https://docs.werwolv.net/pattern-language/core-language/importing-modules it instead is described here, which is not really where it should be as it's an imhex behavior (otherwise just delete the repo and merge the two.)

paxcut:
I seem unable to locate it but I remember working with somebody on the description of how the search paths are structured because originally it was worded in a way that didn't make it clear you needed to create additional folders inside the entries added to the folder tab in settings. I incorrectly thought you were quoting the documentation i just mentioned but now i see you were quoting the settings path tab description. sorry about that.

It is often stated and I agree with it  that documentation could be improved. I find writing documentation to be much harder than fixing bugs.

PR #2750, opened 2026-05-30
fix: bad selections on delete.

### LINKED ISSUE #2748: [Bug] Pattern Editor: Selection skips over lines/characters when done immediately after deleting text
### Operating System

Windows

When using shift-arrows to select text immediately after having deleted some text and without moving the cursor in-between, the selection skips over an equivalent number of characters as were just deleted.

It appears to me that the while the cursor position is updated when deleting, the previous selection end position is not cleared and is restored when selecting again.

A related quirk (which is more unexpected than buggy per-se) is that selecting text to the end of a line and then pressing the End key (in such a way that the cursor does not move) does not clear the selection like in most text editors.

- Start with the cursor in the middle of some text like `// |foobarbaz` (`|` represents the cursor position)
- Use Shift-right to select the text `foo`, then press delete to get `// |barbaz`.
- Without moving the cursor in between, press Shift-right again once.
- Instead of selecting `[b]|arbaz`, the previous selection position is remembered, and you get `b|[ar]baz` instead.

The issue also happens over multiple lines (e.g., selecting a few lines of text, pressing delete, and then using shift-up/down will instantly select the same number of lines as were deleted) and pressing either delete, backspace, ctrl-x to cut, etc.

### ImHex Version

1.39.0.WIP merge@eb1bae1

### ImHex Build Type

### Installation type

MSI

### Additional context?

_No response_

paxcut:
This attempts to fix the problems described in issue #2748. When a selection is deleted, the selection ends are changed correctly, but the interactive selection is not so attempting to use it after the deletion produces unexpected results.

The fix is to update the interactive selection as well. I also made some code more efficient by removing the creation of temporary objects that were not necessary. Some code was repeated several times, so I move the code into their own function. Cleaned the code handling mouse clicks and added some `using` typedefs to shorten code.

### DIFF
--- a/plugins/builtin/source/content/views/view_pattern_editor.cpp
+++ b/plugins/builtin/source/content/views/view_pattern_editor.cpp
@@ -1500,13 +1500,13 @@ namespace hex::plugin::builtin {
                         }
                         return false;
                     });
+                    m_runningParsers += 1;
                     TaskManager::createBackgroundTask("hex.builtin.task.parsing_pattern", [this, code = std::move(code), provider](auto &){
                         this->parsePattern(code, provider);
 
                         if (m_runAutomatically)
                             m_triggerAutoEvaluate = true;
                     });
-                    m_runningParsers += 1;
                     m_hasUnevaluatedChanges.get(provider) = false;
             }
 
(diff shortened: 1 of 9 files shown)


yuriks:
Tested the PR build and can confirm it fixes the selection shifting bug for me!

However... it seems to break text editing in other ways. For example, opening a new file, then pressing Ctrl-A to select the default comment text, then pressing backspace does not successfully delete it, just clearing the selection instead. However, I can then press Ctrl-Z and imhex inserts back a copy of the file, so it seems like it did think it deleted that text although it failed to do so. I'm going to install the previous build I was using just to confirm it's actually a regression.

yuriks:
Can confirm the issue I mentioned above doesn't happen on the latest CI build of the master branch.

I appreciate your patience with these. I feel kinda bad that even after you quickly fix one issue, I end up immediately finding a different one. These are in no way urgent for me so feel free to prioritize other work if you want to.

paxcut:
I really have nothing but appreciation for all the time you are spending on making ImHex a better hex editor  regardless of how much you are or are not enjoying yourself while doing it. From your perspective, finding these bugs is easier than fixing them but from my perspective it is the opposite. As evidence to that I think I have fixed the problem with select all which in turn made a bug in the undo system become visible so I fixed that too. 

I do test the fixes to the best of my ability but as everybody knows, bugs are unavoidable, so please keep trying to find more problems if you are inclined to do so and if you want to hear my opinion, the faster they are found the earlier they are fixed..

PR #2759, opened 2026-06-13
fix: avoid spawning temporary console for utf-8 setup

0x5abe:
### Problem description

To set up utf-8 console output `system("chcp 65001 >nul")` was being used. This was spawning a temporary console which became really annoying when running any sort of script that ran ImHex repeatedly.

Replaced the `system("chcp 65001 >nul")` calls with the WinAPI functions `SetConsoleOutputCP(CP_UTF8)` and `SetConsoleCP(CP_UTF8)` to avoid spawning a new console window.

### DIFF
--- a/main/gui/source/window/platform/windows.cpp
+++ b/main/gui/source/window/platform/windows.cpp
@@ -420,14 +420,16 @@ namespace hex {
             // If the application is running in debug mode, ImHex runs under the CONSOLE subsystem,
             // so we don't need to do anything besides enabling ANSI colors and UTF-8 encoding
             log::impl::enableColorPrinting();
-            system("chcp 65001 >nul");        //codepage 65001 = UTF-8.
+            ::SetConsoleOutputCP(CP_UTF8);
+            ::SetConsoleCP(CP_UTF8);
         } else if (hex::getEnvironmentVariable("__IMHEX_FORWARD_CONSOLE__") == "1") {
             // Check for the __IMHEX_FORWARD_CONSOLE__ environment variable that was set by the forwarder application
 
             // Enable ANSI colors in the console
             log::impl::enableColorPrinting();
             // Enable UTF-8 encoding
-            system("chcp 65001 >nul");
+            ::SetConsoleOutputCP(CP_UTF8);
+            ::SetConsoleCP(CP_UTF8);
         } else {
             log::impl::redirectToFile();
         }


shewitt-au:
LOL. I suggested this but was ignored.

shewitt-au:
The call to `SetConsoleCP` isn't really needed. It sets the input code page. Doesn't hurt though. Calling `SetConsoleOutputCP`, as both you and I have suggested, is the correct way to address utf-8 in the logging window.

widberg:
I'm getting the same incessant terminal popups using `--mcp`. Very annoying. Please merge this.

shewitt-au:
> I'm getting the same incessant terminal popups using `--mcp`. Very annoying. Please merge this.

In the past I've had to @ either WerWolv or Paxcut to garner any attention, and even then, with mixed results. I risk encouraging you to be a nag here.

Frankly, for whatever reason (lack of time perhaps?), one could be forgiven for thinking there in little interest in contributions apart from from the blessed.

I finally tapped out and left the Discord group after suggesting exactly what you are suggesting here. Just got sick of working for nothing.

paxcut:
Please, have a little patience. You may not have noticed that the workflows were approved. I'm working on fixing the archlinux builds atm.

shewitt-au:
@paxcut 

And how patient would you suggest I have to be to get a single comment on my work on [this](https://github.com/WerWolv/ImHex/pull/2677)?

paxcut:
some clarifications are in order considering the content of the posts made by @shewitt-au on this thread.
* You were never ignored once in any of your posts on discord.  Your posts were all answered even when you just said that the code was shit which failed to provide a reason for having to switch the current implementation.
*  Despite the many posts here and on discord about needing to change this not a single one provides with some easy to test case that shows the problem. Only allusions to it using some extra fractions of a second, running under scripts, using --mcp or outdated descriptions on chcp. I went out of my way to look for reasons and found that you can't even tell a terminal is spawned when running imhex from the command line, using a script or using explorer. The problem of terminal popups is something i can't reproduce and i fave tried to.
* The only valid reason that I have found so far is that we still support windows 7 which is known to have problems with chcp.
* The reason I can't merge that PR you mention is that it is more complex than something I can review.

shewitt-au:
@paxcut

"You were never ignored once in any of your posts on discord. Your posts were all answered even when you just said that the code was shit which failed to provide a reason for having to switch the current implementation."

Well, now you know... I tried to tell you. But this isn't the place for this debate I guess, and that is my fault. So, I'll lay off.

0x5abe:
"Despite the many posts here and on discord about needing to change this not a single one provides with some easy to test case that shows the problem."

Here's a small clip recreating it, so there's evidence of it happening

https://github.com/user-attachments/assets/2310c223-31eb-4f07-a3db-9b21c7ca89ed

shewitt-au:
I do not know the exact reason, but nor am I surprised. Getting `system` involved simply to launch an app to call `SetConsoleOutputCP`, when we could just call it directly, was asking for trouble at worst, and inefficient at best.

shewitt-au:
I would guess this:

1. You are launching a script from VSCode.
2. VSCode has no console (in the Win32 API sense).
3. VSCode launches the script interpreter. There is no console to be inherited.
4. The script launches ImHex. Again, no console.
5. ImHex calls `system`. There is no console so it creates one.
6. Repeat from 4 (the console would be inherited from step 4).

If there is no console a call to `SetConsoleOutputCP` is a NOP.

Hard to be sure from the video alone.

paxcut:
@shewitt-au You said you were ignored and two people reacted as surprised and I assume it means they were surprised at you being ignored. Now you admit to not being ignored although only implicitly. You try to make it sound as if it was me not trying to listen but that's also not correct. I can prove that by posting our entire conversation but I don't need to. I'll just post this as sufficient evidence that you couldn't find a real reason
image
 
About it being inefficient, that would only be a valid argument if the inefficiency could be observed. Asking for trouble is not a valid reason either. 

@0x5abe There is no point in trying to find more reasons when one is enough. I am not saying they don't exist, just that I couldn't find one when I tried my best.  I wrote a windows script (not a python) and there was not problems I could find, but my script called the imhex gui. The only reason the code was added was to  format the log correctly, so when no log is echoed to the terminal, like when running the imhex cl, the change of code page is not necessary. The fix for that would not involve the changes made here though.  You would simply not change the code page when the command line imhex was invoked thus eliminating the problems with your script and running --mcp.

There were several posts in discord that expressed the same idea but none provided a reason until a regular pointed out that imhex runs on windows 7  which I have read had problems with chcp having side effects when used. That is a valid reason because the side effects would extend to anything run after the imhex gui, but it is well documented that windows 10 and 11 eliminated those problems and chcp is now safe to use when a gui app is just using the terminal to print data. I think it is just unpopular because of what it used to do.

shewitt-au:
Just call `SetConsoleOutputCP` and fix the issue, will you? Calling `system` is asking for trouble, as this issue would seem to verify. Merge [0x5abe](https://github.com/0x5abe)'s changes.

paxcut:
> Just call SetConsoleOutputCP and fix the issue, will you? Calling system is asking for trouble, as this issue would seem to verify.

1) issue is not cause by calling chcp. it is caused by spawning a terminal when the command line version of imhex is used. The fix for that is to not change the code page when the command line is used. It does not involve the changes made here.
2) I already stated that asking fort trouble is not a valid reason yet you repeat the same non-reason.
3) this issue does not verify that spawning a terminal when calling the gui of imhex causes any problems.
4) I have provided a real reason to change it so there is no need to invent an imaginary one.

paxcut:
@0x5abe @shewitt-au 
I think there is a major misunderstanding here. When you call the cl imhex (aka forwarder) you don't use the `imhex-gui.exe` but you use `imhex.exe`. The former will launch the gui so it isn't what is being shown on the video. What is also true is that if the command line is not empty then the forwarder never calls the `initNative()` function so that it is not possible that the terminals being created in your video are from calls to `system()`. I ran the forwarder under the debugger using `--pl run -v` and `initNative()` and therefore `system()` were never called. 

If as @0x5abe suggest, the popups were created by the call to system() and popups do not occur after applying the changes  then we have some rather bizarre bug that is running imhex-gui code when imgui.exe is invoked with cl arguments. The other possibility is that you are running imhex-gui with command line parameters in a debug build which makes little sense so we probably issue some waning or add code that will redirect the log to a file like we do in the no debug builds.

0x5abe:
> @0x5abe @shewitt-au I think there is a major misunderstanding here. When you call the cl imhex (aka forwarder) you don't use the `imhex-gui.exe` but you use `imhex.exe`. The former will launch the gui so it isn't what is being shown on the video. What is also true is that if the command line is not empty then the forwarder never calls the `initNative()` function so that it is not possible that the terminals being created in your video are from calls to `system()`. I ran the forwarder under the debugger using `--pl run -v` and `initNative()` and therefore `system()` were never called.
> 
> If as @0x5abe suggest, the popups were created by the call to system() and popups do not occur after applying the changes then we have some rather bizarre bug that is running imhex-gui code when imgui.exe is invoked with cl arguments. The other possibility is that you are running imhex-gui with command line parameters in a debug build which makes little sense so we probably issue some waning or add code that will redirect the log to a file like we do in the no debug builds.

I don't know what to tell you, the changes I made stops the popups from happening

paxcut:
But are you calling imhex-gui.exe or imhex.exe? your video only showes imhex_path but not itss value.Are you running a debug build?

0x5abe:
> But are you calling imhex-gui.exe or imhex.exe? your video only showes imhex_path but not itss value.Are you running a debug build?

Calling `imhex.exe`, using the nightly download from commit `296e5c1` Windows-Portable-x86_64.

paxcut:
That wouldn't allow you to make changes to test if it fixed tings. never mind, thats your commit

paxcut:
ok, that means that i need to run a complete working command line to see why it is not using the else since there is no log sent to the console. it is trycky because i can't use the debugger to check what it is running.

main/gui/source/window/platform/windows.cpp
```diff
@@ -420,14 +420,16 @@ namespace hex {
             // If the application is running in debug mode, ImHex runs under the CONSOLE subsystem,
             // so we don't need to do anything besides enabling ANSI colors and UTF-8 encoding
             log::impl::enableColorPrinting();
-            system("chcp 65001 >nul");        //codepage 65001 = UTF-8.
+            ::SetConsoleOutputCP(CP_UTF8);
+            ::SetConsoleCP(CP_UTF8);
```
paxcut, line 424: I don't think this one is needed

main/gui/source/window/platform/windows.cpp
```diff
@@ -420,14 +420,16 @@ namespace hex {
...
         } else if (hex::getEnvironmentVariable("__IMHEX_FORWARD_CONSOLE__") == "1") {
             // Check for the __IMHEX_FORWARD_CONSOLE__ environment variable that was set by the forwarder application
 
             // Enable ANSI colors in the console
             log::impl::enableColorPrinting();
             // Enable UTF-8 encoding
-            system("chcp 65001 >nul");
+            ::SetConsoleOutputCP(CP_UTF8);
+            ::SetConsoleCP(CP_UTF8);
```
paxcut, line 432: Both of these need to be removed and also delete the comment above

main/gui/source/window/platform/windows.cpp
```diff
@@ -420,14 +420,16 @@ namespace hex {
...
             log::impl::enableColorPrinting();
-            system("chcp 65001 >nul");        //codepage 65001 = UTF-8.
+            ::SetConsoleOutputCP(CP_UTF8);
+            ::SetConsoleCP(CP_UTF8);
         } else if (hex::getEnvironmentVariable("__IMHEX_FORWARD_CONSOLE__") == "1") {
             // Check for the __IMHEX_FORWARD_CONSOLE__ environment variable that was set by the forwarder application
 
             // Enable ANSI colors in the console
             log::impl::enableColorPrinting();
             // Enable UTF-8 encoding
```
paxcut, line 430: remove this too

paxcut (changes requested):
The reasons behind these changes proposed are explained in my last comment in the discussion.

paxcut:
I think I finally understand what the real problem is in the video you showed thanks to the information you provided. the relevant code is
```cpp
void Window::initNative() {
        if (ImHexApi::System::isDebugBuild()) {
            // If the application is running in debug mode, ImHex runs under the CONSOLE subsystem,
            // so we don't need to do anything besides enabling ANSI colors and UTF-8 encoding
            log::impl::enableColorPrinting();
            system("chcp 65001 >nul");        //codepage 65001 = UTF-8.
        } else if (hex::getEnvironmentVariable("__IMHEX_FORWARD_CONSOLE__") == "1") {
            // Check for the __IMHEX_FORWARD_CONSOLE__ environment variable that was set by the forwarder application

            // Enable ANSI colors in the console
            log::impl::enableColorPrinting();
            // Enable UTF-8 encoding
            system("chcp 65001 >nul");
        } else {
            log::impl::redirectToFile();
        }

         :
         :
    }
    ```
Three choices. At first glance the first and the second seem to be similar and the third is different. I incorrectly used that to assume the first two were used for the gui and the last for the cl. in fact it is the middle one that is for the cl because the environment variable is set by the forwarder.

The top one will never create terminals because it always has one which is where the log is written. The log has alignment requirements that need to set the code page to utf-8 for the console, but the console in this case is guaranteed to exist (we actually discard the one that is spawned for all guis if the gui is not run from explorer). It will be changed to keep supporting windows 7 builds and may be reverted back when we stop win 7 support as it may provide some advantages to spawn a terminal of we ever run into a case when it doesn't have one.

The middle one uses the console to write error messages and console output generated by the pattern language when it runs patterns. These messages don't have alignment requirements so they don't need to set code pages. If not output is created by the PL then the console may not even be created when running scripts but if they do create any output they will pop open and close as each instance of the command line is run. I traced the problem with using --mcp to the same exact issue.  Adding a code page here is a mistake regardless of the method used to add it.

The last one is used when running release guis. even if you run them from a command line the logs will be redirected to a file. This needs no further explanation.

Before merging this PR we should discuss these findings and some code changes I propose based on them that are part of the review to make sure we all agree they are the best changes we can make at this point.

WerWolv:
What on earth are you guys even arguing about here. `system()` is never the right thing to use for this. The PR looks reasonable and if it fixes the issue at hand, I say we merge it. Even if it ends up not solving the issue fully, it's strictly superior to calling `system()`

@paxcut and @shewitt-au you two, take your differences elsewhere. I literally cannot deal with the constant bickering and arguing anymore. You're making me and everybody else miserable with this shit.

0x5abe:
Is this ever gonna get merged? 💀

paxcut:
The obvious question to ask before considering if this pr should be merged as  is is why the code page is being changed on a console that's not writing any text to its output. As explained above the code page change to utf-8 was added only to align the log columns that are shown in the console when running a debug build because the log can contain utf-8 characters depending on the language chosen.

The addition of the code page change when the console is not displaying the log was a mistake that I made that doesn't need to be repeated by this PR. The only change needed to align the logs is the output and input is never requested on the console when running the gui,  so setting the input can be avoided although this is a minor point.

0x5abe:
Wow ur so fast when ur not ignoring WerWolv telling you to merge it

0x5abe:
Since it you who made the original mistake I don't see why I have to be the one to remove the code. I just made it behave the same way it was before, you have had write access this whole time.

WerWolv:
No more arguing here, please

PR #2761, opened 2026-06-16
Replace NFD Dependency With DialogModule

samuelvenable:
### Problem description

File dialogs, (as provided with the NFD dep), do not support a file filter dropdown box on macOS. This fixes that, by replacing the NFD dependency with `libdlgmod`. More details here: https://github.com/samuelvenable/libdlgmod

I replaced the NFD dependency with my own file dialogs which are improved ones compared to NFD. They also match the current desktop environment: Qt on KDE, GTK on non-KDE, WinAPI on Windows, and AppKit on macOS.

### Screenshots

Screenshot 2026-07-04 at 1 20 15 AM

Screenshot 2026-07-04 at 1 19 56 AM

Screenshot 2026-07-04 at 1 19 15 AM

Screenshot 2026-07-04 at 1 18 13 AM


paxcut:
Originally ImHex used this same library but it was very unpopular. Another imgui based app (shadered) uses that same file dialog and I noticed in 2023 that the author looked into using it in ImHex. this is what he said when I asked about about it.

image

image

image

If you join us in discord, you will be able to read more comments about it and about the one used to replace it which you say has problems but  have never noticed any.

WerWolv:
Hey! Latest commit properly sets the parent window on the file dialogs. I hope this solves some of the issues.

About the rest of the PR, having to pull in the entirety of SDL is pretty much a show stopper for me. I've been thinking about replacing GLFW with SDL3 eventually but that will require a ton of work elsewhere. I'd rather not have both in at the same time

PR #2789, opened 2026-07-19
feat: merge close confirmation popups

iTrooz:
### Problem description

Users often get 2 confirmation popups when closing ImHex: one to save the provider, and one to save the project. Even worst: they often get popups when they just saved their provider and have no project data, meaning ImHex should not prompt the user at all !

This PR fixes this path in order to streamline the usage of projectless ImHex (and it probably also improves projects)

- Implemented two dirty flags per provider: data dirty, and metadata (project data) dirty. This separation allows us to not prompt the user at all for a project if they don't need one
- In the case they do need one, the two popups are merged, and what needs to be saved is clarified

### Screenshots
image

### Additional things
This PR also fixes the popups showing when you have already saved your provider, and had absolutely nothing to save (Let's not gaslight the users into thinking they forgot to save something lol)


paxcut:
ImHex does not prompt to save changes that don't exist anymore. Making changes to the provider generates changes that need to be saved to the project even if you have already saved the provider to disk. If this pr is discarding all the changes that go into the project when the input file is saved then that would be equivalent to not prompting the user about unsaved changes when they exit imhex.

iTrooz:
Don't worry, this PR doesn't discard changes, it just merges both popups into a single one, with more choices

paxcut:
In that case it is fine. It is just that you keep using an example where project data is being generated and saying that imhex is asking to save nonexistent changes when the changes do in fact exist. Prior to this PR there is currently no popups at all created when they are not needed.

iTrooz:
Could you clarify ? My example is the following:
- Open binary file
- Edit bytes
- Save file
- Try to exit ImHex

No project data should be generated if I haven't touched bookmarks, the pattern, highlights..

paxcut:
open View>Patches before making the3 edits and you'll see what i mean

paxcut:
Imhex wants to save the patches data generated by your edits into the project so when you load the project you can undo changes if you want to.

WerWolv:
Maybe we should change that for the patches specifically. Have them only be saved to the project if there actually is one open. And otherwise simply ask if the changes made to the file should be saved to the file when closing

paxcut:
thats a good idea.Some changes are not meant to be undone so it is good to be able to discard them without effort.

iTrooz:
I agree

To still allow good experience, I should add a small text to explain patches in the save popup, and always keep the "save project" button available (currently it is grayed out when there are just patches, which as AxCut said, is problematic)

iTrooz:
Are you sure this works ? I just tried the latest release (and the latest nightly), and I can't save patches to the project and restore them.
And in the source code, saving patches seems to be a no-op: https://github.com/WerWolv/ImHex/blob/58d3203878fabe470b34c0d132d75976dcb523ac/plugins/builtin/source/content/views/view_patches.cpp#L41

paxcut:
you are right. patches used to be stored in projects but now they can be exported and i dont see them in the project anymore now im confused about why is imhex asking to save projects when the file itself is not even saved as part of the project unless it is an unsaved new file with edits.

paxcut:
here is the current content on a provider

image

And an old project i have:
image

Issue #1270, opened 2023-08-28
[Bug] Copy/Paste of a larger amount of data doesn't work


PlasmaHH:
### Operating System

Linux

### What's the issue you encountered?

Copy/Pasting of larger amounts of data from one file into another doesn't work.

What I do is, I select some amount of data in one file, ctrl-c to copy it, then go into another file and the place where I want to overwrite the data, then ctrl-v there. For small amounts of data ( a couple of 16 byte lines maybe, haven't found the actual number yet ) this works fine, data is overwritten, displayed in red. When doing it for more data ( like, several kB ) it doesn't to anything.

### How can the issue be reproduced?

Open a sufficiently big file ( maybe a few 100k? Not sure here ). CTRL-A to select everything, CTRL-C to copy. Now open a new file. Resize to something big. CTRL-A to select everything. CTRL-V to paste. Nothing happens.

Doing the same with a smaller selection of the origin file however does work.

### ImHex Version

1.30.1

### ImHex Build Type

- [ ] Nightly or built from sources

### Installation type

OpenSUSE Tumbleweed  update

manswad:
Same in windows

PR #2793, opened 2026-07-21
feat: paste from source popup for pasting data from another provider

jiishiin:
### Feature
Paste from Source popup for pasting data from another provider

### Reason for implementation
Pasting very large amount of bytes (like 2.5 GB) from clipboard either freezes ImHex or doesn't paste anything. This feature lets you to paste data from source to destination provider (if selected from hex editor, it will be set as the default destination).

### Screenshots
imhex_context_menu

feature_info

### Feature Description

I have added a feature to paste data directly from another file since pasting large size of data from the clipboard doesn't work as expected.

It has following features - 

1. ProviderSselector: Allows to select the files opened in Hex Editor

2. Region Selection Toggle: It allows to select the region of the provider directly from Hex Editor.

3. Select the entire provider region

4. Shows the Begin area of the current selection

5. Shows the End area of the current selection

6. Clear currently selected region

7. Swap source and destination provider and selection region

8. Input the selection region manually

9. Paste Modes: Has multiple modes of paste operation
   - Paste Over Selection :- Paste the selected source region over the selected destination region. If the source region is smaller than the destination region, the remaining bytes in the destination region are left unchanged.
   
   - Paste Everything :- Paste the entire selected source region starting at the begin address of the selected destination region. The destination provider may be resized to fit the source region data..
   
   - Replace Selection :- Replace the selected destination region with the selected source region. The selected destination region will be resized to fit the selected source region data.
   
   - Insert Everything :- Insert the entire selected source region starting at the begin address of the selected destination region, shifting the existing data forward.
   
10. Info Box: Shows information regarding paste operation.

11. Recommendation Toggle: When enabled automatically shows the recommended paste modes based on the destination selection size.
    - It checks whether the destination selection is single byte, multi byte or zero byte.
    
12. Input Base Toggle: Switch between hexadecimal and decimal view of the input region/size.

13. Apply paste operation

14. Cancel 

## Video

https://github.com/user-attachments/assets/6d538e54-077b-4e87-b71d-179bbf80921c

## Known issues

- Implementation uses raw read, write, insert and remove provider APIs. So undo/redo will not work for modification done using this feature.

- This feature does not allows source and destination provider to be the same.


jiishiin:
@WerWolv Please have a look at this feature pull request.

jiishiin:
Hello @paxcut, I have implemented this feature to allow pasting of larger data from another file since pasting large data from clipboard makes ImHex not-responding. If ImHex is accepting feature PRs, can you review this?

paxcut:
I saw this yesterday and I was very impressed with the level of detail and the massive amount of options available. I work mostly on the pattern editor and can fix simple bugs and review small PRs outside of that, so this is a bit too big for me to handle. OTOH the code is self contained and there are not many changes to existing systems so I may get permission to review it but I need to ask first.

I couldn't find any mention of an undo feature either as a proper feature or as a limitation. Would the builtin undo/redo system be able to recover data in case of an accidental paste?

The last thing I wanted to mention is the reason posted for the creation of this feature. Is the pasting of large files using the clipboard breaking a know issue? How big are we talking about? Is it just slow or does it hang ImHex?
I see this feature as an alternative to using the clipboard, not as a substitute, so us developers should look into what is happening on that end too.

jiishiin:
I have mentioned the limitation of not being able to undo/redo in the known issues (1st point) section due to the uses of RAW provider api's for write/insert/remove.  When I had used the normal provider APIs, its was causing the same not responding issue. 
Since the write operation is done in a chunk of 4KB, the amount of undo/redo will be very large for a very large data if normal write/insert/remove provider api is used. So I had to use the raw apis.

The reason for this feature was, I was trying to fix a corrupted video file, so I had to replace some of the atoms in the corrupted video file highlighted by the mp4 pattern. While copy/pasting data ( i think it was over 2.5 GB) the ImHex won't either respond, or doesn't paste anything. I think similar behavior was reported a long back (Issue #1270).

Now the only way to revert an accidental paste is to close the file without saving. I will add a warning regarding the undo/redo limitation for now.

jiishiin:
@paxcut I have added a warning regarding the undo/redo limitation.
image

jiishiin:
Added info icon.
image

jiishiin:
Added file path in the tool tip of the file list. Useful if files with same name are open.
image

jiishiin:
@WerWolv I have submitted this PR to allow user to also paste data from another file without using clipboard. Please review it if ImHex is accepting feature PRs.

WerWolv:
Hey @jiishiin
Thanks for the PR! I looked over the code real quick but there's quite a few styling inconsistencies everywhere. Please make sure to follow the style the rest of the project uses (most notably, we don't really use snake_case, defines or /* */ style comments)

As for the feature, I really like the idea and wouldn't mind merging it eventually.

About the UI, please use `ImGuiExt::DimmedButton()` for the buttons as that's what's used in all other UIs. Please also don't hardcode colors but instead use the existing ones through the style system (or add new ones to the styles if you really must), otherwise theming these will not be possible.

Once these things are fixed I'm happy to do a proper code review

jiishiin:
> Hey @jiishiin Thanks for the PR! I looked over the code real quick but there's quite a few styling inconsistencies everywhere. Please make sure to follow the style the rest of the project uses (most notably, we don't really use snake_case, defines or /* */ style comments)
> 
> As for the feature, I really like the idea and wouldn't mind merging it eventually.
> 
> About the UI, please use `ImGuiExt::DimmedButton()` for the buttons as that's what's used in all other UIs. Please also don't hardcode colors but instead use the existing ones through the style system (or add new ones to the styles if you really must), otherwise theming these will not be possible.
> 
> Once these things are fixed I'm happy to do a proper code review

Thanks for the quick review. I will make the necessary changes and let you know.

jiishiin:
@WerWolv I have made the changes as per the initial review.

image

jiishiin:
Added Select everything button to select entire region.

image

jiishiin:
Added paste confirmation popup.

image

jiishiin:
Added icons to show which file is opened in hex editor (current provider). Also clicking the header willl set it as the current provider.

image

plugins/builtin/include/content/popups/hex_editor/popup_hex_editor_paste_from_file.hpp
```diff
@@ -0,0 +1,82 @@
...
+            void drawPasteModeButton(const char *buttonLabel, PasteModeType buttonMode, PasteHintType buttonHint, bool buttonDisable, float buttonWidth);
+            void drawIntegerInputField(const char *inputLabel, void *inputValue, float inputWidth);
+            void drawPasteConfirmationPopup(ViewHexEditor *editor);
+
+            void setSelection(u64 start, u64 end);
+            bool executePasteOperation(void) const;
+            std::string elapsedTimeFormatted(void) const;
+
+        private:
+            const float m_tableWidth = 600_scaled; // Set the same width for all tables which decides the popup width
```
WerWolv: This needs to be a private member function that gets called from the other functions. Otherwise when scaling changes while the popup is still open, the table size will not update until it is reopened

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+#include <hex/api/imhex_api/hex_editor.hpp>
+#include <hex/helpers/fmt.hpp>
+#include <imgui_internal.h>
+#include <fonts/vscode_icons.hpp>
+
+namespace hex::plugin::builtin {
+
+    PopupPasteFromFile::PopupPasteFromFile(const ImHexApi::HexEditor::ProviderRegion &selection) {
+        // Pin the popup window
+        this->setPinned(true);
```
WerWolv, line 16: This should be a deliberate operation by the user like for the other popups

plugins/builtin/include/content/popups/hex_editor/popup_hex_editor_paste_from_file.hpp
```diff
@@ -0,0 +1,82 @@
...
+            std::string elapsedTimeFormatted(void) const;
+
+        private:
+            const float m_tableWidth = 600_scaled; // Set the same width for all tables which decides the popup width 
+            bool m_modeRecommend = true;
+            bool m_inputBaseHex = true;
+            PasteModeType m_pasteMode = PasteModeType::ModeNotSelected;
+            PasteHintType m_pasteHint = PasteHintType::HintDefaultDescription;
+            std::chrono::nanoseconds m_elapsedTime{0};
+            std::array<ProviderInfo, 2> m_fileProvider;
```
WerWolv: `m_providers`. FileProvider has the specific meaning of the provider that can open files from the filesystem but ImHex supports many more data sources.

I'd argue this should also be a struct as to be more explicit with which of the two entries is the source and which one the destination

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+
+        auto &[src, dst] = m_fileProvider;
```
WerWolv: I'd use `source` and `dest` as the names

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+        drawPasteInfoBox();
+        ImGui::Spacing();
+        drawPasteDecisionBox(editor);
+    }
+
+    void PopupPasteFromFile::drawFileSelectionBox(void) {
+        auto &[src, dst] = m_fileProvider;
+
+        // Table for selecting source and destination file
+        if (ImGui::BeginTable("##table_1", 2, (ImGuiTableFlags_BordersOuter | ImGuiTableFlags_BordersInnerV | ImGuiTableFlags_SizingStretchSame), ImVec2(m_tableWidth, 0.0f))) {
```
WerWolv: Give the tables a more descriptive name. That helps a lot when debugging ImGui issues

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+            drawRegionProvider(dst, editor);
+            ImGui::Spacing();
+
+            ImGui::EndTable();
+        }
+    }
+
+    void PopupPasteFromFile::drawPasteModeBox(void) {
+        auto &[src, dst] = m_fileProvider;
+        auto isDstSelectionMultiByte = (dst.providerSelection.size > 1);
```
WerWolv: These can all be `const`

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+            // Second column
+            ImGui::TableSetColumnIndex(1);
+            auto decisionButtonWidth = ((ImGui::GetContentRegionAvail().x - ImGui::GetStyle().ItemSpacing.x) / 2.0f);
+
+            ImGui::Spacing();
+            // Apply button
+            auto &[src, dst] = m_fileProvider;
+            auto &srcSelection = src.providerSelection;
+            auto &dstSelection = dst.providerSelection;
+            auto isSrcRegionOffsetsValid = (src.providerValidity &&
```
WerWolv: `const` too

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+
+                ImGui::EndTabBar();
+            }
+            ImGui::PopID();
+        }
+        ImGui::EndDisabled();
+    }
+
+    void PopupPasteFromFile::drawFilePathToolTip(const prv::Provider *provider) const {
+        auto *dataDescriptionProvider = dynamic_cast<const prv::IProviderDataDescription*>(provider);
```
WerWolv: This code here is wrong. DataDescription can return literally anything in any order. Additionally, Providers do not need to have a path associated with them. For example the ProcessMemoryProvider doesn't open any file, it opens a live process. 

Don't try to draw information like this that does not make sense in all cases

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+                ImGui::PopStyleColor();
+            }
+        }
+        ImGui::EndDisabled();
+    }
+
+    void PopupPasteFromFile::drawIntegerInputField(const char *inputLabel, void *inputValue, float inputWidth) {
+        ImGui::SetNextItemWidth(inputWidth);
+        ImGuiExt::InputIntegerPrefix(inputLabel,
+                                     (m_inputBaseHex ? "0x" : "0d"),
```
WerWolv: I'd not use any prefix for Decimal. We don't anywhere else either

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+        m_setSelectionStart = start;
+        m_setSelectionEnd = end;
+        m_setSelectionTrigger = true;
+    }
+
+    bool PopupPasteFromFile::executePasteOperation(void) const {
+        auto &[src, dst] = m_fileProvider;
+        u64 pasteSize = 0;
+
+        if ( (src.providerSelection.provider == nullptr) || (dst.providerSelection.provider == nullptr)) {
```
WerWolv: Extra space

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+                break;
+            }
+
+            default: {
+                return false; // Invalid paste mode
+            }
+        }
+
+        // Paste the source region data over the respective (selected/resized/inserted) region of destination
+        constexpr size_t chunkSize = 0x1000; // 4 KB : 4096 Bytes
```
WerWolv: `constexpr static size_t ChunkSize = 4_KiB;`

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+            bytesLeft -= chunk;
+        }
+
+        dst.providerSelection.provider->markDataDirty();
+
+        return true; // Paste successfull
+    }
+
+    std::string PopupPasteFromFile::elapsedTimeFormatted(void) const {
+        using namespace std::chrono;
```
WerWolv: `using namespace std::literals::chrono_literals`;

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+
+            // Second column: Swap action button (rendering the swap column first)
+            ImGui::TableSetColumnIndex(1);
+            // Read the column width (_SizingStretchSame: all columns will have same width)
+            auto availColumnWidth = (ImGui::GetContentRegionAvail().x);
+            // Calculate swap button width
+            auto swapButtonWidth = (ImGui::CalcTextSize(ICON_VS_ARROW_SWAP, nullptr, true).x + ImGui::GetStyle().FramePadding.x * 2.0f);
+            // Center the swap button
+            ImGui::SetCursorPosX(ImGui::GetCursorPosX() + (availColumnWidth/2) - (swapButtonWidth/2));
+            ImGui::PushID("button_swap_files");
```
WerWolv: I don't think that PushID/PopID is necessary

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
```diff
@@ -0,0 +1,968 @@
...
+        ImGui::Spacing();
+        drawPasteDecisionBox(editor);
+    }
+
+    void PopupPasteFromFile::drawFileSelectionBox(void) {
+        auto &[src, dst] = m_fileProvider;
+
+        // Table for selecting source and destination file
+        if (ImGui::BeginTable("##table_1", 2, (ImGuiTableFlags_BordersOuter | ImGuiTableFlags_BordersInnerV | ImGuiTableFlags_SizingStretchSame), ImVec2(m_tableWidth, 0.0f))) {
+            auto srcColumnHeader = fmt::format("{} {}", ((ImHexApi::Provider::getCurrentProviderIndex() == src.providerIndex) ? ICON_VS_FILE_BINARY : ICON_VS_FILE),
```
WerWolv: `const`

WerWolv (changes requested):
Overall the code looks really clean, good job!

plugins/builtin/include/content/popups/hex_editor/popup_hex_editor_paste_from_file.hpp
jiishiin: I have made it a struct.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: I made it pin at the start to avoid accidentally closing the popup. Also when the selection toggle is active, if its not pinned user will not be able to select the region from hex editor selection.

Now I have made it to only pin the popup when selection toggle is active and restoring the original pin status when selection toggle is turned off.

My only concern is that once the user sets the provider and region, they might accidentally close the popup and have to redo all the steps again. Let me know which is the better approach.

plugins/builtin/include/content/popups/hex_editor/popup_hex_editor_paste_from_file.hpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: I have removed this now.

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
jiishiin: Done.

jiishiin:
Hello @WerWolv , Thanks for the detailed review. I have made the requested changes. I would like to hear your final opinion on this point.

image

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_paste_from_file.cpp
WerWolv: You're probably right, given that it's a lot of settings. Let's leave it as is

jiishiin: Yeah, i have kept it pinned at the start.

WerWolv (approved):

WerWolv:
Thanks a lot for the effort!

Issue #568, opened 2022-07-05
[Bug] can't open file with spaces in name


bsdf:
### Operating System

Linux

### What's the issue you encountered?

when opening a file with a space in the name, the NFD library returns a url-encoded filename which std::fs doesn't seem to like. (`/home/bsdf/hello world.txt` becomes `/home/bsdf/hello%20world.txt`)

### How can the issue be reproduced?

open a file with a space in the name on Linux

### ImHex Version

master and 1.19.0

### ImHex Build Type

- [X] Nightly or built from sources

iTrooz:
Hey, could you test if this problem was present in the last release ?

bsdf:
appears to work as expected in v1.18.2

PR #2800, opened 2026-07-25
fix: Percent-encoded paths when opening files on macOS

n0kovo:
### Problem description

Trying to open `/tmp/my file.bin` by passing the path to the ImHex executable, results in `Failed to open data source: Failed to open file /tmp/my%20file.bin: No such file or directory`

`-[HexDocument readFromURL:ofType:error:]` in `lib/libimhex/source/helpers/utils_macos.m` pulled the path out of `-[NSURL absoluteString]`, so `open()` got `/tmp/my%20file.bin` (so this goes for all characters encoded by NSURL, not just spaces).

The argv route breaks for the same reason. AppKit turns launch arguments into `readFromURL:` calls for document-based apps, so the CLI path never reaches the `--open` subcommand handler.

This affects:
- `/Applications/ImHex.app/Contents/MacOS/ImHex "/tmp/my file.bin"`
- `open -a ImHex "/tmp/my file.bin"`
- Dragging and dropping `/tmp/my file.bin` onto the app icon

It’s not an issue when using the UI file open dialog or dragging and dropping into the UI.

Use `-filePathURL` plus `-fileSystemRepresentation`. That also resolves `file:///.file/id=…` reference URLs, and the new early `return NO` sets `outError` per the `NSDocument` contract.

Verified against real files with a path containing a space, and through AppKit's actual delivery path for both invocations.Tested on macOS Tahoe 26.5.1.

Related issue (maybe): #568

### DIFF
--- a/lib/libimhex/source/helpers/utils_macos.m
+++ b/lib/libimhex/source/helpers/utils_macos.m
@@ -277,14 +277,24 @@
     @implementation HexDocument
 
     - (BOOL) readFromURL:(NSURL *)url ofType:(NSString *)typeName error:(NSError **)outError {
-        NSString* urlString = [url absoluteString];
-        const char* utf8String = [urlString UTF8String];
+        (void)typeName;
 
-        const char *prefix = "file://";
-        if (strncmp(utf8String, prefix, strlen(prefix)) == 0)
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #2807, opened 2026-07-29
Fix toolbar create and open icons disappearing

### LINKED ISSUE #2803: [Bug] Toolbar icons are missing
### Operating System

Windows

I noticed that two icons from the toolbar, _Create new File_ and _Open File_, are missing after I added the _Theme Manager_ and _Settings_ icon to the toolbar (see screenshot 1).

I tried to fix it by removing the _Create new File_ icon and the adding it to the toolbar again. But then even more icons disappeared (see screenshot 2).

**Screenshot 1:**

Image

&nbsp;
**Screenshot 2:**

Image

See above.

I don't know how to reproduce this.

### ImHex Version

1.39.0.WIP

### ImHex Build Type

### Installation type

Portable

### Additional context?

_No response_

MinhakaDev:
Fixes #2803

### Problem description

When customizing the toolbar (e.g. adding the Theme Manager or Settings icons), the **Create new File** and **Open File** icons disappear from the toolbar. Removing and re-adding them makes the icons vanish.

These two are the only default toolbar items that are also registered via `addTaskBarMenuItem`. That call inserts a *second* entry into `s_menuItems` with the same leaf name as the real menu item, but with an empty glyph.

The toolbar setting's `load()` matched menu items by leaf name only (`unlocalizedNames.back()`). Because `s_menuItems` is iterated in priority order, the lower-priority empty-glyph taskbar duplicate matched first and received the `toolbarIndex`. The toolbar draw loop then culls any item whose `icon.glyph` is empty (resetting its `toolbarIndex` to `-1`), so the icon disappeared.

The fix skips glyph-less menu items when restoring the toolbar in `load()`, so the real icon-carrying item gets the toolbar slot instead of its taskbar duplicate. An item with no glyph can never be a valid toolbar button, so this is consistent with the existing invariant already enforced in the draw loop.

### Screenshots
Before:
image

After:
image

### DIFF
--- a/plugins/builtin/source/content/settings_entries.cpp
+++ b/plugins/builtin/source/content/settings_entries.cpp
@@ -679,9 +679,9 @@ for (const auto &path : m_paths) {
                     return;
 
                 for (auto &[priority, menuItem] : ContentRegistry::UserInterface::impl::getMenuItemsMutable()) {
-		    if (menuItem.icon.glyph.empty()) {
-			    continue;
-		    }
+                    if (menuItem.icon.glyph.empty()) {
+                        continue;
+                    }
                     for (const auto &[index, value] : toolbarItems) {
                         const auto &[name, color] = value;
                         if (menuItem.unlocalizedNames.back().get() == name) {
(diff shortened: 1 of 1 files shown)


WerWolv (commented):
Fix indenting

WerWolv:
Thank you!

PR #2808, opened 2026-07-29
fix(builtin/process_memory): fix ANSI encoding crash and scan boundaries on Windows

ChronoBit:
### Description of Changes

1. **Fix UTF-8 Encoding Crash on Non-English Locales:**
   * Replaced `GetModuleFileNameExA` with `GetModuleFileNameExW` using extended buffer length to support long paths and avoid ANSI/CP1251 system error crashes in `std::filesystem::path`.

2. **Fix `VirtualQueryEx` Loop Boundary & Performance:**
   * Bound memory region scanning to `sysInfo.lpMaximumApplicationAddress` using `GetSystemInfo()`.
   * Correctly advance the memory address cursor using `memoryInfo.BaseAddress + regionSize` to avoid getting stuck in loops at user-space boundaries.

3. **Fix `EnumProcessModules` Buffer Resizing:**
   * Fixed vector resizing logic when process module count exceeds the initial allocation.

### DIFF
--- a/plugins/builtin/source/content/providers/process_memory_provider.cpp
+++ b/plugins/builtin/source/content/providers/process_memory_provider.cpp
@@ -473,44 +473,69 @@ namespace hex::plugin::builtin {
 
         #if defined(OS_WINDOWS)
             DWORD numModules = 0;
-            std::vector<HMODULE> modules;
+            std::vector<HMODULE> modules(1024);
 
-            do {
-                modules.resize(modules.size() + 1024);
-                if (EnumProcessModules(m_processHandle, modules.data(), modules.size() * sizeof(HMODULE), &numModules) == FALSE) {
-                    modules.clear();
-                    break;
+            if (EnumProcessModules(m_processHandle, modules.data(), modules.size() * sizeof(HMODULE), &numModules)) {
+                if (numModules > modules.size() * sizeof(HMODULE)) {
+                    modules.resize(numModules / sizeof(HMODULE));
+                    EnumProcessModules(m_processHandle, modules.data(), modules.size() * sizeof(HMODULE), &numModules);
                 }
-            } while (numModules == modules.size() * sizeof(HMODULE));
-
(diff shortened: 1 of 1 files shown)


Nemoumbra:
How can I repro these bugs on master?

ChronoBit:
Hi @Nemoumbra! Here are the steps to reproduce both issues on Windows:

### 1. Crash on non-ASCII paths (CP1251 / Localized Windows)
* **Pre-conditions:** Run Windows on a non-English locale (e.g., Russian, Japanese, etc.) or attach to a process located in a directory path containing non-ASCII/Cyrillic characters.
* **Steps:**
  1. Open ImHex.
  2. Select the **Process Memory** provider and attach to a process whose executable path contains non-ASCII characters (or has loaded DLLs from such paths).
* **Behavior on `master`:** `GetModuleFileNameExA` returns an ANSI-encoded string (e.g., CP1251). Passing this to `std::filesystem::path` triggers a `std::system_error` / `throw_conversion_error` exception inside MSVC/GCC runtime because it expects valid UTF-8, causing ImHex to crash.
* **Fix:** Using `GetModuleFileNameExW` handles Wide/UTF-16 paths safely.

---

### 2. VirtualQueryEx Scan Boundaries
* **Steps:** Open the **Process Memory** provider on Windows for any 64-bit target process.
* **Behavior on `master`:** 
  The loop condition `address < this->getActualSize()` uses a hardcoded maximum size (`0xFFFF'FFFF'FFFF`) rather than querying the OS for the actual user-space boundary (`0x7FFFFFFF0000`). 
* **Fix:** 
  Uses `GetSystemInfo(&sysInfo).lpMaximumApplicationAddress` to properly bound the upper memory limit for user-mode applications, and ensures the memory address cursor correctly advances via `memoryInfo.BaseAddress + regionSize`.

WerWolv:
Thank you!

PR #2810, opened 2026-08-02
impr: don't draw an empty settings box for disassemblers

neptuwunium:
### Problem description
If a disassembler provider has no settings, ImHex draws an empty box. This is a very minor opinionated aesthetics thing.

Add a hasSettings method that returns true if settings should be drawn, this is a breaking change.

### Screenshots
Left: Before
Right: After

image


WerWolv:
Thank you!

PR #2821, opened 2026-08-08
fix: Updated paste validation check and provider relative offset address for paste from source operation

jiishiin:
### Problem description
1. Provider base address was not considered during paste validation check.
2. Provider relative offset was not passed to raw provider APIs.

1. Updated region validation before paste operation
2. Used provider relative offset address for raw provider APIs.


jiishiin:
@WerWolv Please review this.

WerWolv:
Thank you!

Issue #2825, opened 2026-08-09
How to make ImHex easier to package for distributions ?


iTrooz:

neptuwunium:
-Werror is still a mandatory compile option for the disassembler subproject, aside from that i've basically upstreamed most of the workarounds my personal gentoo overlay had to build it without consuming most of the vendored dependencies.

mbedtls3 is the only issue left, and that just requires the `-3` variants to be added to the FIND_LIBRARY calls in `cmake/modules/FindmbedTLS.cmake`.

n.b. I don't maintain the Gentoo GURU package for ImHex, but I do occasionally fix it up. There's a separate ImHex package in my own distro repo.

WerWolv:
The last two commits removed the explicit warning flags from libdisassembler (using the same one as for the rest of ImHex instead) and added the -3 suffix to the mbedTLS search. Hope that helps

PR #2831, opened 2026-08-11
Fix AES modes

connor-cihula:
### Problem description

#2811 Revealed issues with AES decryption.

- ECB: Breaks on input that is anything other than exactly one 16 byte block of input.
- CBC: Didn't set a padding mode. 
- CFB128: Correct.
- CTR: Correct.
- GCM: Decrypts without authenticating.
- CCM: Broken.
- OFB: Correct  
- XTS: Broken and hidden from UI.

-  GCM/CCM: API change to allow usage of AAD and tag.
- CBC: Use `MBEDTLS_PADDING_NONE`.
- XTS: Was ignored before, I continue to ignore.

### Screenshots

Before:
image
After:
image

### Additional things

@WerWolv once approved, I'll work on a PR for GNUTLS support as an alternative to MbedTLS, which should help with #2825 and conclude the discussion on #2811


WerWolv:
Hey! Thank you very much! I didn't realize there was so much wrong with these nodes

Code quality wise it looks good to me but I'm obviously lacking the expertise to properly review this. Gonna see if Copilot has something to say but otherwise we can merge this if the CI passes

lib/libimhex/source/helpers/crypto.cpp
```diff
@@ -512,39 +509,78 @@ namespace hex::crypt {
...
+            }
+        }
+
+        if (mode == MBEDTLS_MODE_GCM || mode == MBEDTLS_MODE_CCM) {
+            const auto &aeadIv = mode == MBEDTLS_MODE_GCM ? iv : nonce;
+            std::vector<u8> authenticatedInput;
+            authenticatedInput.reserve(input.size() + tag.size());
+            authenticatedInput.insert(authenticatedInput.end(), input.begin(), input.end());
+            authenticatedInput.insert(authenticatedInput.end(), tag.begin(), tag.end());
+
```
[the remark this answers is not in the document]
WerWolv: That can be safely ignored I think since the usecase is to allow people to decrypt any, even badly encrypted data.

WerWolv:
CI failures are unrelated and the unit tests pass so from my side this is okay 👍

WerWolv (approved):

lib/libimhex/source/helpers/crypto.cpp
```diff
@@ -512,39 +509,78 @@ namespace hex::crypt {
...
+            output.resize(outputSize);
+            return output;
         }
 
+        if (input.empty())
+            return {};
+
+        std::array<u8, 16> nonceCounter = { 0 };
+        std::ranges::copy_n(nonce.begin(), std::min(nonce.size(), size_t(8)), nonceCounter.begin());
+        std::ranges::copy_n(iv.begin(), std::min(iv.size(), size_t(8)), nonceCounter.begin() + 8);
```
neptuwunium: I think this silently truncates and moves the first 8 bytes of the IV to the second half of the nonceCounter which is used as the IV later. Pin inputs would end up needing to be nonce = iv[0...8] iv = iv[8...16]. Would make more sense to just use iv if the mode does not consume a nonce value.

plugins/builtin/source/content/data_processor_nodes/decode_nodes.cpp
```diff
@@ -41,20 +41,29 @@ namespace hex::plugin::builtin {
             if (input.empty())
                 throwNodeError("Input cannot be empty");
 
-            std::array<u8, 8> ivData = { 0 }, nonceData = { 0 };
+            const bool isAuthenticatedMode = mode == crypt::AESMode::GCM || mode == crypt::AESMode::CCM;
+            const bool requiresIv = mode != crypt::AESMode::ECB && mode != crypt::AESMode::CCM;
+            const bool requiresNonce = mode != crypt::AESMode::ECB && mode != crypt::AESMode::GCM;
```
neptuwunium: this evaluates to true on CBC, which relates to the aforementioned IV footgun

neptuwunium:
looked it over, there's some surprises with CBC, while mathematically correct nobody uses two separate values for CBC IV in reality

lib/libimhex/source/helpers/crypto.cpp
connor-cihula: Good call out. The strange IV + nonce interaction was existing behavior, my changes just fed into that. The recent commit should solve that issue.

neptuwunium:
Perfect

WerWolv:
Thanks a lot everyone!

PR #2832, opened 2026-08-11
Allow renaming in-memory provider tabs

### LINKED ISSUE #2669: [Feature] Rename section opened as a separate view
### What feature would you like to see?

When you open a `Section` from a parsed file into its "Raw Data" view, you are unable to rename the tab.

### How will this feature be useful to you and others?

When you are working on multiple files with the same structure, for example to figure out what changed between version A, B or C of file, you end up with having a bunch of tabs that all have the same name.

It should work the same way as a Selection View works, which can be renamed

### Request Type

- [ ] I can provide a PoC for this feature or am willing to work on it myself and submit a PR

### Additional context?

_No response_

tianrking:
## Summary

- add the existing rename action to in-memory provider context menus
- allow raw section views to be renamed like other provider tabs
- refresh the window title after a provider is renamed

Fixes #2669

## Validation

- `git diff --check`
- CMake Debug configure with GCC 15.2 and Ninja
- `ninja -C build/codex plugins/builtin/CMakeFiles/builtin.dir/source/content/ui_items.cpp.o -j2`
- `ninja -C build/codex lib/libimhex/CMakeFiles/libimhex.dir/source/providers/memory_provider.cpp.o -j2`

### DIFF
--- a/plugins/builtin/source/content/ui_items.cpp
+++ b/plugins/builtin/source/content/ui_items.cpp
@@ -16,6 +16,7 @@
 #include <hex/helpers/fmt.hpp>
 #include <hex/helpers/logger.hpp>
 #include <hex/helpers/debugging.hpp>
+#include <hex/providers/memory_provider.hpp>
 #include <hex/providers/provider.hpp>
 
 #include <fonts/vscode_icons.hpp>
@@ -27,6 +28,7 @@
 
 #include <toasts/toast_notification.hpp>
 #include <popups/popup_question.hpp>
+#include <popups/popup_text_input.hpp>
 
 #include <csignal>
 #include <fonts/tabler_icons.hpp>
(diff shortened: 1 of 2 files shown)


plugins/builtin/source/content/ui_items.cpp
```diff
@@ -423,6 +425,15 @@ namespace hex::plugin::builtin {
     }
 
     static void drawProviderContextMenu(prv::Provider *provider) {
+        if (auto *memoryProvider = dynamic_cast<prv::MemoryProvider*>(provider); memoryProvider != nullptr) {
```
WerWolv, line 428: Instead of hardcoding the menu items here, use the existing `IProviderMenuItems` facility

WerWolv (changes requested):

PR #2833, opened 2026-08-12
fix: Set cursor position explicitly before jumping to the selection in paste from source feature

jiishiin:
### Problem description
Due the cursor position change in hex editor setSelection function in commit af154270bf3baa92a602206138f3018d8c170fb0, jumping to begin or end using action buttons is not working as expected in the paste from source popup.

- Explicitly set the cursor position before jumping to the selection.
- Also fixed the popup border highlight getting clipped

### DIFF
--- a/plugins/builtin/include/content/views/view_hex_editor.hpp
+++ b/plugins/builtin/include/content/views/view_hex_editor.hpp
@@ -38,6 +38,10 @@ namespace hex::plugin::builtin {
             m_hexEditor.setSelection(start, end);
         }
 
+        void setCursorPosition(u64 cursorPosition) {
+            m_hexEditor.setCursorPosition(cursorPosition);
+        }
+
         void jumpToSelection() {
             m_hexEditor.jumpToSelection();
         }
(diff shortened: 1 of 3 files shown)


paxcut:
Can you explain what you mean by it not working as expected?
The way I understand the feature is that pressing shift allows you to resume the selection and pressing ctrl-t allows you to switch from one end to the other. That appears to work correctly and you can see  where the cursor is when it jumps to the other end.

jiishiin:
> Can you explain what you mean by it not working as expected? The way I understand the feature is that pressing shift allows you to resume the selection and pressing ctrl-t allows you to switch from one end to the other. That appears to work correctly and you can see where the cursor is when it jumps to the other end.

This is specific to the paste from source feature I implemented. There are action buttons which can be clicked to show the begin and end of current selection using the setSelection() and jumpToSelection() api. With the commit af154270bf3baa92a602206138f3018d8c170fb0 the setSelection() will set the cursor postion to selection end. So in the feature pressing the goto begin/end was not working as expected. This change explicitly sets the cursor postion before jumping to the selection.

paxcut:
Thanks for the explanation and changing the title to clarify.

jiishiin:
@WerWolv Please review this.

WerWolv:
Thank you!

PR #2835, opened 2026-08-13
add IMHEX_ENABLE_TELEMETRY (default ON) and fix telemetry config defaults

laineu:
### Problem
Some Linux distros (for example Gentoo) have stricter rules about telemetry being in software and prefers giving users the option of opting out at compile-time (https://wiki.gentoo.org/wiki/Software_Telemetry).

### Implementation
Added IMHEX_ENABLE_TELEMETRY as an option (default ON) and fixed a couple of telemetry config defaults being left enabled because the first-time popup asking for telemetry never appeared.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(IMHEX_ENABLE_CPPCHECK            "Enable cppcheck static analysis"       
 option(IMHEX_BUNDLE_PLUGIN_SDK          "Enable bundling of Plugin SDK into install package"                                            ON )
 set   (IMHEX_RENDERING_BACKEND          "GLFW" CACHE STRING "Set the rendering backend ImHex should use (GLFW or SDL)"                     )
 option(IMHEX_ENABLE_UPDATER             "Enable automatic updater"                                                                      ON )
+option(IMHEX_ENABLE_TELEMETRY           "Build with telemetry support"                                                                  ON )
 ## Testing
 option(IMHEX_ENABLE_UNIT_TESTS          "Enable building unit tests"                                                                    OFF)
 option(IMHEX_ENABLE_IMGUI_TEST_ENGINE   "Enable the ImGui Test Engine"                                                                  OFF)
--- a/cmake/build_helpers.cmake
+++ b/cmake/build_helpers.cmake
@@ -139,6 +139,10 @@ macro(addDefines)
     if (IMHEX_ENABLE_UPDATER)
         add_compile_definitions(IMHEX_ENABLE_UPDATER)
     endif()
+
+    if (IMHEX_ENABLE_TELEMETRY)
+        add_compile_definitions(IMHEX_ENABLE_TELEMETRY)
+    endif()
 endmacro()
 
 function(addDefineToSource SOURCE DEFINE)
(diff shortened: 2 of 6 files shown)


WerWolv:
Hey! Thanks for the PR
However, I'm not open to merging this right now as this will basically render the Telemetry entirely useless. Every packager will immediately enable it on their builds and make the whole thing useless.

I strongly believe that this should not be the decision of the person packaging ImHex but of the users themselves whose (fully anonymous) data is being collected. They get asked on the first start if they want to allow or reject it and we honor the `DO_NOT_TRACK` environment variables since a few days. That has to be enough.

laineu:
>I'm not open to merging this right now as this will basically render the Telemetry entirely useless. Every packager will immediately enable it on their builds and make the whole thing useless.
> 
> I strongly believe that this should not be the decision of the person packaging ImHex but of the users themselves whose (fully anonymous) data is being collected. They get asked on the first start if they want to allow or reject it

This doesn't touch the consent part at all, it only changes whether it's even allowed to ask in the first place. The config changes were made because some options were left enabled even when telemetry wasn't allowed (ex: for OS_WEB).

First-time users will still be asked for consent.

WerWolv:
If there's issues with when the telemetry is enabled, then that should be a separate PR. My stance is purely because of the `IMHEX_ENABLE_TELEMETRY` compile time option. With that enabled, there would be no point to ask for consent anyways since you, the packager, would have already made that decision for the user

laineu:
I think there's been a huge misunderstanding, having this enabled does NOT opt the user into telemetry (the flag can be renamed if its confusing), it only gives a compile-time option of if telemetry CAN be enabled.

with IMHEX_ENABLE_TELEMETRY=ON, you can still disable it and it will work EXACTLY the same as before:
Screenshot_20260813_164921
Screenshot_20260813_164938

with IMHEX_ENABLE_TELEMETRY=OFF, the telemetry options are grayed out and are not allowed to be enabled:
Screenshot_20260813_165206

WerWolv:
No, that's exactly how I understood the PR. 
I'd like the end user to make the final decision if they want to have telemetry enabled or not, not somebody further up the chain disabling it entirely for the build.

neptuwunium:
As for Gentoo, ImHex satisfies

> Ideally the user is informed about the connection and the transferred data

with an opt in behavior on first launch.

ImHex also satisfies

> Even an automatic check for updates on the upstream website enables upstream to collect more data than some users want to spread.

via IMHEX_ENABLE_UPDATER which is set to OFF on the GURU ebuild, and typically is standard on most distro repo builds due to updates being managed by the distro instead.

From my understanding, Gentoo's preposition of the compile time toggle is presented as an option for if the design time consent controls are too much effort to implement. This would also only really give consent controls to source build consumers on Gentoo, Nix and Guix while leaving everyone else to use git/livebuilds or use whatever the distro decides is good for them.

neptuwunium:
An alternative would be to add IMHEX_DISABLE_TELEMETRY_ONBOARDING, which just disables the first launch splash and implies IMHEX_ENABLE_UPDATER=OFF. Leaving the toggle functional in the settings for if they choose to re-enable it later.

Additionally, under this patch having the controls present but greyed out is a design anti-pattern. If disabled with no way to enable it at runtime it should just not be rendered.

PR #2838, opened 2026-08-14
fix: only exit on startup if an error happened

neptuwunium:
### Problem description

At the moment running `imhex path/to/file` without a main instance will just exit the program rather than showing the main interface. 

This is caused by this commit: https://github.com/WerWolv/ImHex/commit/ced975989b052cb6937716615d7f41e76a29bf50

also EXIT_CONTINUE was introduced, which changes the exit code behavior.

this changes the subcommand processing loop to check for EXIT_CONTINUE, and only exits the full program if the final exit code is not EXIT_CONTINUE


WerWolv:
Thank you!

PR #2846, opened 2026-08-16
fix: preserve custom encoding row cache boundaries

### LINKED ISSUE #2805: [Bug] Custom encoding view misaligns multi-byte characters (e.g. GBK) across row boundaries
### Operating System

Windows

When using the custom encoding view with a multi-byte encoding (GB2312/GBK), the decoder occasionally misaligns on row boundaries.

Example: `D3 A6 D3 C3` should decode to `应` + `用`, but sometimes it decodes as `应` + `τ` (`D3 A6` + `A6 D3`). This suggests an off-by-one byte offset after a row boundary.

1. Open a **large file** containing some multi-byte encoded text that **spans across a row boundary** (part at the end of one row, rest at the beginning of the next).
2. Enable custom encoding view and select a multi-byte encoding.
3. Scroll to the region containing the text. The decoding may not appear immediately; scroll around a bit until it shows. When it does, observe that some characters near the row boundary may be decoded incorrectly.

Not 100% reproducible. Even with the same file and the same location, the result can vary.

### ImHex Version

1.38.1

### ImHex Build Type

### Installation type

imhex-1.38.1-Windows-x86_64.msi

### Additional context?

In `plugins/builtin/source/views/hex_editor.cpp` around line 983, `m_encodingLineStartAddresses` uses `push_back` on every render without being resized or cleared. Possibly related.

atirna:
### Problem description

Fixes #2805. The custom encoding view could use continuation offsets written by earlier repaint passes at the wrong row index, causing scroll-dependent decoding errors around row boundaries for multi-byte encodings.

Store a computed continuation only when rendering the current cache tail. The cache therefore keeps one decode-start offset per file row instead of appending stale offsets on every repaint. Adds a focused `EncodingLineStartAddressCache` helpers test that covers both already-cached rows and tail extension.

### Screenshots

Not applicable; this corrects decoding state rather than layout.

### Additional things

The PR adds focused regression coverage for the cache-tail behavior.


lib/libimhex/include/hex/helpers/encoding_line_cache.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <hex/helpers/types.hpp>
+
+#include <vector>
+
+namespace hex::impl {
+
+    inline void appendEncodingLineStartAddress(std::vector<u64> &lineStartAddresses, size_t line, u64 nextLineStartAddress) {
```
WerWolv: Move this to the encoding_file.cpp instead of implementing it in a new header

WerWolv (changes requested):

lib/libimhex/include/hex/helpers/encoding_line_cache.hpp
atirna: i moved the helper definition into `lib/libimhex/source/helpers/encoding_file.cpp`, declared it from the existing encoding header, and removed `encoding_line_cache.hpp`. the tail-only condition is still `line + 1 == lineStartAddresses.size()`; `EncodingFile::getEncodingFor` is unchanged. `clang++ -std=c++23 -fsyntax-only` passed for the implementation and the `EncodingLineStartAddressCache` test source. head: `e8f4e3dac8b16df0fd9160a5cfa2b8be269df81e`. thanks for the placement note, nothing needed from you.

WerWolv:
Simply for transparency sake, this entire PR and interaction feels **extremely** like I'm talking to an LLM. 
Did you use AI for this change?

atirna:
Hi @WerWolv , I did use AI for review/and quick fix to your comment using a review bot. But initial code change/and submissions were done by me.

WerWolv:
Thank you!

atirna:
sorry for trouble, thank you for merge

PR #2849, opened 2026-08-17
Add Nettle support

connor-cihula:
### Problem description

Resolves issues discussed in #2811 (MbedTLS instability) and #2825 (ease of packaging).

Nettle can be chosen as the crypto backend with -DCRYPTO_BACKEND=nettle. MbedTLS remains the default, but can be explicitly set with -DCRYPTO_BACKEND=mbedtls.

### Additional things

In #2811 we decided on GnuTLS, but turns out GnuTLS doesn't support the AES modes that ImHex uses. Nettle does, and GnuTLS depends on Nettle, so I went with Nettle.


tests/algorithms/source/aes.cpp
```diff
@@ -1,6 +1,13 @@
 #include <hex/helpers/crypto.hpp>
 #include <hex/test/tests.hpp>
 
+#if defined(IMHEX_CRYPTO_BACKEND_MBEDTLS)
+    #include <mbedtls/ccm.h>
+    #include <mbedtls/cipher.h>
+    #include <mbedtls/error.h>
+    #include <mbedtls/gcm.h>
+#endif
```
billatarm: We shouldn't spill backend details across a generic interface. Ill need to change this.

billatarm (changes requested):

billatarm:
bump @WerWolv any update on this patch?

PR #2859, opened 2026-08-21
feat: Added option to set the address of the selected byte

jiishiin:

Added option to set the address of the selected byte.

### Screenshots

Icon Disabled:
image

Set Base Address:
image

Set Byte Address:
image


WerWolv:
Thank you!

I like the idea but I think a simple toggle in the Base Address popup would be better for this. Since the two options are basically identical besides the offset.
Additionally, right now you're using the start address of the selected region to set the address. Would using the current cursor position make more sense maybe? Or maybe instead only allowing the option to be used if the selection size is 1?

jiishiin:
> Thank you!
> 
> I like the idea but I think a simple toggle in the Base Address popup would be better for this. Since the two options are basically identical besides the offset. Additionally, right now you're using the start address of the selected region to set the address. Would using the current cursor position make more sense maybe? Or maybe instead only allowing the option to be used if the selection size is 1?

The reason I used a selection size of 1 is that, if a selection is active with size > 2, the cursor position can be at either the start or the end, depending on the direction of the selection. So, if the cursor position is used as the target byte while multiple bytes are selected, the user might end up setting the address at the wrong byte position. To avoid this scenario, I chose to enable this popup only when single byte is selected.

I think the 1 byte selection approach also makes it difficult to use this within the Set Base Address popup, since the selection shouldn't matter when opening the Set Base Address popup.

WerWolv:
Okay I kinda misunderstood how this feature worked, sorry. I built the branch locally now and played around with it a bit. 

For me, the UX doesn't feel great. Both the Base Address and Byte Address setting ultimately change the same value but using different semantics. Both options are right next to each other, called almost the same and open a popup that looks almost the same. I feel like these two will be confused with each other quite a lot.

Instead what I'd suggest is maybe going the same route as the CTRL + F Find dialog. We have a single `Set Base Address...` option in the right-click menu which opens a popup that has tabs for the two options. Base Address is selected by default and Byte Address (or maybe better Current Byte Address or something) is secondary and grayed out unless only one byte is selected.
That way we could also give the tabs some tooltips that explain what exactly the difference is. I think that would make it easier to understand in the end

neptuwunium:
What if it used different phrasing? Like "Set Address Anchor" or something, though ideally that would just be a button you press without input, so you'd still need to use Set Base Address and THEN press Set Address Anchor

WerWolv:
@neptuwunium This could maybe be implemented with just a toggle in the Base Address popup. By default, have it set the base address of the entire file but if some "Anchor" toggle button is enabled, have it set the address for the byte that's currently selected. Again with a tooltip there what the toggle does. That might be the cleanest solution actually

jiishiin:
@WerWolv I have updated the Set Base Address UI as per your suggestion. Should the checkbox be placed above or below the input?

When single byte is selected:
image

When multiple bytes are selected:
image

WerWolv:
What does the `Set base address` checkbox do? Wouldn't disabling that mean to just not do anything?

What I meant was basically this (conceptually):
```cpp
ImGui::TextInput();
ImGui::SameLine();
ImGuiExt::DimmedIconToggle(ICON_TA_ANCHOR, ICON_TA_ANCHOR_OFF);
ImGui::SetItemTooltip("Some Description here about what the this option does");
```

Sorry for not being clear with the last message

jiishiin:
> What does the `Set base address` checkbox do? Wouldn't disabling that mean to just not do anything?
> 
> What I meant was basically this (conceptually):
> 
> ```c++
> ImGui::TextInput();
> ImGui::SameLine();
> ImGuiExt::DimmedIconToggle(ICON_TA_ANCHOR, ICON_TA_ANCHOR_OFF);
> ImGui::SetItemTooltip("Some Description here about what the this option does");
> ```
> 
> Sorry for not being clear with the last message

Here the check box behaves like a toggle. So one of the option will always be selected.

https://github.com/user-attachments/assets/1f96f5f5-80b9-4009-8be2-aa6ddae7d14f

WerWolv:
You basically manually implemented `ImGui::RadioButton()` there :D
So the logic is fine, you're just using the incorrect widget right now. ImGui offers `ImGui::RadioButton()` but ImHex _generally_ uses `ImGuiExt::DimmedIconToggle()` with two different icons to do this.

jiishiin:
> You basically manually implemented `ImGui::RadioButton()` there :D So the logic is fine, you're just using the incorrect widget right now. ImGui offers `ImGui::RadioButton()` but ImHex _generally_ uses `ImGuiExt::DimmedIconToggle()` with two different icons to do this.

If I use the ImGuiExt::DimmedIconToggle(), whats the best way to show which mode is currently active, since tooltip only tells what it does.

WerWolv:
By using the icons. You can provide two icons to `ImGuiExt::DimmedIconToggle` like in the example I did. One with an Anchor with a disabled line through it, one with an Anchor without the line. That plus the blue outline of the button when it's enabled should be enough to show when the setting is on or off. The tooltip should only say what the option does then, not what the current setting is

jiishiin:
> By using the icons. You can provide two icons to `ImGuiExt::DimmedIconToggle` like in the example I did. One with an Anchor with a disabled line through it, one with an Anchor without the line. That plus the blue outline of the button when it's enabled should be enough to show when the setting is on or off. The tooltip should only say what the option does then, not what the current setting is

Sorry for the confusion, what i meant was which icon is for which mode (base or byte) from user perspective.

WerWolv:
I figured the anchor with the disabled line would be the normal one and the Anchor without the line would be the byte mode. But feel free to look through the list of available icons here, maybe you can find some that fit better:

`ICON_VS_XXX`: https://microsoft.github.io/vscode-codicons/dist/codicon.html
`ICON_TA_XXX`: https://tabler.io/icons

jiishiin:
> I figured the anchor with the disabled line would be the normal one and the Anchor without the line would be the byte mode. But feel free to look through the list of available icons here, maybe you can find some that fit better:
> 
> `ICON_VS_XXX`: https://microsoft.github.io/vscode-codicons/dist/codicon.html `ICON_TA_XXX`: https://tabler.io/icons

Sure. Thank you!

jiishiin:
Icon Disabled:
image

Set Base Address:
image

Set Byte Address:
image

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_base_address.cpp
```diff
@@ -1,57 +1,114 @@
...
 namespace hex::plugin::builtin {
 
-    PopupBaseAddress::PopupBaseAddress(u64 baseAddress) : m_baseAddress(baseAddress) {}
+    PopupBaseAddress::PopupBaseAddress(const ImHexApi::HexEditor::ProviderRegion &selection) 
+            : m_selection(selection), m_baseAddress(0), m_byteAddress(0), m_byteOffset(0), 
+              m_byteAddressMode(false), m_byteAddressDisable(true) {
+        if (m_selection.getProvider() != nullptr) {
+            m_baseAddress = m_selection.getProvider()->getBaseAddress();
+            m_byteAddressDisable = ((m_selection.getProvider()->getActualSize() == 0) || 
+                                    (m_selection.getSize() != 1)) ? true : false;
```
WerWolv: `true : false` is redundant

plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_base_address.cpp
```diff
@@ -1,57 +1,114 @@
...
         const auto width = ImGui::GetWindowWidth();
-        ImGuiExt::InputHexadecimal("##base_address", &m_baseAddress);
-        if (ImGui::IsItemFocused() && (ImGui::IsKeyPressed(ImGuiKey_Enter) || ImGui::IsKeyPressed(ImGuiKey_KeypadEnter))) {
-            if (this->isValid()) {
-                setBaseAddress(m_baseAddress);
-                editor->closePopup();
-            }
+
+        if (!m_byteAddressMode){ // Set base address
+            ImGuiExt::InputHexadecimal("##base_address_input", &m_baseAddress, ImGuiInputTextFlags_AutoSelectAll);
```
WerWolv: Do the condition on the variable here, otherwise the ID changes of the input which can cause some problems with tabbing. 

`m_byteAddressMode ? &m_byteAddress : &m_baseAddress`

WerWolv:
Besides the two things I commented on, this looks good for merging. Thanks a lot!

jiishiin:
> Besides the two things I commented on, this looks good for merging. Thanks a lot!

I have fixed the review comments. Thank you!

WerWolv:
Thanks!

Issue #2847, opened 2026-08-17
[Bug] Bytes are filled at wrong offset on a provider with non-zero base address


jiishiin:
### Operating System

Windows

### What's the issue you encountered?

Bytes are filled at wrong offset on a provider with non-zero base address.

The addition of the provider base address to the input address might be the root cause since the input address is already relative to the base address.

<img width="1082" height="513" alt="Image" src="https://github.com/user-attachments/assets/dede48da-a988-4f7d-8bf2-70897a2f0263" />

### How can the issue be reproduced?

Set base address other than zero. Fill bytes using Fill.. popup.

### ImHex Version

242416bc55024cbc9d34f2b914713131d6bfb15f

### ImHex Build Type

- [x] Nightly or built from sources

### Installation type

built from source

paxcut:
Thanks for the report. I was able to reproduce the problem, but \rather than adding the base address twice as suggested it appears that the base address is being completely ignored by fill. Setting the starting fill address at 0x10 always fills bytes starting 10 bytes from the start of the file for any base address. My guess is that the base address is not being set at all but I need to debug it to tell for sure.

jiishiin:
> Thanks for the report. I was able to reproduce the problem, but \rather than adding the base address twice as suggested it appears that the base address is being completely ignored by fill. Setting the starting fill address at 0x10 always fills bytes starting 10 bytes from the start of the file for any base address. My guess is that the base address is not being set at all but I need to debug it to tell for sure.

After removing the highlighted base address and simply assigning "startAddress = address", the issue was not reproduced. This might be the fix.

paxcut:
Yes, of course. 

Contrary to what I first thought, adding the base address twice has the net effect of ignoring the base address altogether. For example if base address is zero then using 5 as the fill address fills the fifth byte. With a base address of 2 using a 5 as fill address should fill the 3rd byte. Adding the base address to the fill address, which is already relative to base is adding the base address twice as in the bug. In that case  you get 8 which is the fifth byte from the start, the same as when base address was zero.
My mistake was not realizing we both were saying the same thing in different ways.

paxcut:
An easier way to say it is that the fill address is already relative to base so to find its position in the byte array you need to subtract the base address from the fill address. With BA of 5 fill address of 7 is  7-5= the 2nd byte. When you add the base address to the fill address you make it as if the base address was zero because 7-5+5 = 7 and that is independent of the numbers chosen.

PR #2863, opened 2026-08-24
fix: Incorrect start address used while filling bytes

jiishiin:
### Problem description
For Fill feature, when a region is selected, the start address is already relative to base address. So adding the base address again to the selected region start address fills the bytes at incorrect offset.

1. Fixed incorrect start address used while filling bytes
2. Fixed random bytes button alignment

### Screenshots
Before:
image

Now:
image


jiishiin:
Fixes [#2847](https://github.com/WerWolv/ImHex/issues/2847)

jiishiin:
I think this feature also needs region validity check before applying fill.

WerWolv:
Thank you! Checking region validity is a good point but that can be addressed out of this PR

PR #2865, opened 2026-08-25
Add support for LZF decompression to the pattern language

ddmitv:
Adds `builtin::hex::dec::lzf_decompress` function to the pattern language for data decompression using LZF algorithm.
It is a modified version of the original liblzf implementation that can be found here: http://software.schmorp.de/pkg/liblzf.html (use Wayback Machine to access if unavailable).

Since this decompression algorithm is extremely simple, it doesn't introduce any new dependencies. I tested this function on modified [`dod-save.hexpat`](https://github.com/WerWolv/ImHex-Patterns/blob/master/patterns/dod-save.hexpat) that uses `builtin::hex::dec::lzf_decompress` to decompress save data and parse the result in another section. Additionally, I fuzz tested `lzf_decompress` for about an hour with libFuzzer + ASAN + UBSAN ([fuzzer source code](https://godbolt.org/z/s9aYYxGnr)), so I'm pretty sure it doesn't contain any vulnerabilities or bugs.

If this PR is merged, I think I will next make a PR to the ImHex-Patterns repo to actually add `lzf_decompress` to the hex library.

### DIFF
--- a/plugins/decompress/source/content/pl_functions.cpp
+++ b/plugins/decompress/source/content/pl_functions.cpp
@@ -2,6 +2,7 @@
 #include <hex/plugin.hpp>
 
 #include <hex/api/content_registry/pattern_language.hpp>
+#include <hex/helpers/literals.hpp>
 #include <pl/core/evaluator.hpp>
 #include <pl/patterns/pattern.hpp>
 
@@ -9,6 +10,8 @@
 
 #include <vector>
 #include <optional>
+#include <cstring>
+#include <algorithm>
 
 #if IMHEX_FEATURE_ENABLED(ZLIB)
     #include <zlib.h>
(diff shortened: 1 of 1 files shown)


plugins/decompress/source/content/pl_functions.cpp
```diff
@@ -345,6 +347,71 @@ namespace hex::plugin::decompress {
...
         });
+
+        /* lzf_decompress(compressed_pattern, section_id) */
+        ContentRegistry::PatternLanguage::addFunction(nsHexDec, "lzf_decompress", FunctionParameterCount::exactly(2), [](Evaluator *evaluator, auto params) -> std::optional<Token::Literal> {
+            // original liblzf implementation: http://software.schmorp.de/pkg/liblzf.html (use Wayback Machine to access it)
+            const auto input = getCompressedData(evaluator, params[0]);
+            auto &output = evaluator->getSection(u64(params[1].toUnsigned()));
+
+            // decompressed size heuristic with a limit
+            constexpr size_t InitialSizeLimit = 0x40000000; /* 1 GiB */
```
WerWolv: You can use ImHex's size UDLs for this. `1_GiB`

plugins/decompress/source/content/pl_functions.cpp
```diff
@@ -345,6 +347,71 @@ namespace hex::plugin::decompress {
...
+            output.resize(std::min(input.size() * 12, InitialSizeLimit));
+
+            size_t iidx = 0;
```
WerWolv: These could use more descriptive names unless if they refer to a name used in some specification

WerWolv (changes requested):

ddmitv:
Should be fixed. The reason I used `0x40000000` instead of `1_GiB` is that `lzma_decompress` actually uses it too, but I don't think I should change it in this PR

WerWolv (approved):

WerWolv:
Thank you!

PR #2867, opened 2026-08-27
Add copy start/end addresses to the context menu for the diffing view table

ddmitv:
Adds a context menu to both "Data Source A" and "Data Source B" columns in the `diffing` plugin's differences table with "Copy Start Address" and "Copy End Address" buttons.

image

I have also moved and slightly changed the call to `ImGui::Selectable`, since `ImGui::BeginPopupContextItem` would apply to the entire row due to the `ImGuiSelectableFlags_SpanAllColumns` flag. This caused right clicking anywhere on the row to trigger the context menu which was not the intended behavior.


ddmitv:
Hi @WerWolv, could you please take a look at this PR and https://github.com/WerWolv/libwolv/pull/46 when you have time?

WerWolv:
Thank you!

PR #2868, opened 2026-08-27
fix: Update highlighting on base address change

jiishiin:
### Problem description
When base address is changed, the modified region/byte highlighting from the previous address is added at the same address relative to the new base address.

Post EventHighlightingChanged event after setting new base address.

### DIFF
--- a/plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_base_address.cpp
+++ b/plugins/builtin/source/content/popups/hex_editor/popup_hex_editor_base_address.cpp
@@ -1,6 +1,8 @@
 #include "content/popups/hex_editor/popup_hex_editor_base_address.hpp"
 #include "content/views/view_hex_editor.hpp"
 
+#include <hex/api/events/events_interaction.hpp>
+
 #include <fonts/tabler_icons.hpp>
 
 namespace hex::plugin::builtin {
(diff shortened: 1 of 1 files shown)


jiishiin:
Although this fixed the highlighting of modified byte/region, the pattern highlights still gets shifted on base address change. May be its not a valid use case since changing base address does not run pattern script (begin address is always mapped @ 0x00).

@WerWolv Is it possible for the pattern scripts to map the start address to base address of the current provider instead of hardcoding @ 0x00?

paxcut:
I just checked on my local build and patterns detect the base address as they should. For example try
```cpp
#pragma example 11 22 33 44
#pragma base_address 5
u32 a @5;
```
Note that the base address can also be changed using the menu with the exact same effect

jiishiin:
> I just checked on my local build and patterns detect the base address as they should. For example try
> 
> ```c++
> #pragma example 11 22 33 44
> #pragma base_address 5
> u32 a @5;
> ```

Sorry I was speaking with respect to the existing scripts, For example running mp4 pattern on mp4 file after changing the base address to some non-zero value.

jiishiin:
@paxcut What I meant was, instead of modifying the script every time, is it possible for the script to dynamically use the base address of the current provider as the start address?

paxcut:
For that case use `std::mem::base_address()` to place the pattern instead of zero. There is no way to change existing patterns that use 0 to use something else, but you can use import * to place the pattern at a different location

paxcut:
It just occurred to me that another option is ti place the pattern at $ which will always be the first address available.

PR #2875, opened 2026-09-03
feat: Add Unifont Upper and per-script Noto fallback fonts

scottanderson:

Add unifont_upper.otf. It covers the supplementary Unicode plane, such as historic scripts and most emoji, that the existing Unifont file does not.

Add Noto Sans Egyptian Hieroglyphs, Noto Sans Cuneiform, and Noto Sans Anatolian Hieroglyphs. Register all three, and Noto Color Emoji below, ahead of Unifont Upper, so they take priority for the codepoints they cover. All four are the unmodified files from their upstream releases.

Add Noto Color Emoji, and a `colorGlyphs` flag on `registerMergeFont()`. FreeType needs `ImGuiFreeTypeLoaderFlags_LoadColor` to load a font's color glyphs instead of converting them to grayscale, and `ImGuiFreeTypeLoaderFlags_Bitmap` to load them at all. Only Noto Color Emoji sets this flag.

Noto Color Emoji and Noto Sans Egyptian Hieroglyphs both map `U+200D` (zero width joiner) to a blank glyph for their own internal ligature use, and both are registered ahead of Unifont, so `U+200D` renders blank instead of with Unifont's own visible glyph for it. Left as upstream ships it rather than hand-patching the binary, so the bundled font files stay byte-for-byte verifiable against their release archives.

Select the closest available strike with `FT_Select_Size()` on a non-scalable face, such as this bitmap-only color emoji font, instead of `FT_Request_Size()`. Scale the glyph's advance width and on-screen geometry to the size actually requested. The bitmap itself stays at the strike's native resolution and is scaled on display, the same as any other font glyph. This is a local patch to vendored Dear ImGui code, following the `IMHEX PATCH` convention already used elsewhere in this file.


WerWolv:
Still unsure here if I really want to add that big of a font just to draw some hieroglyphs. I'm completely fine with the emojis though

PR #2879, opened 2026-09-05
fix: multiple pattern related bugs

neptuwunium:
1. Fixes a `-Werror` violation in debug builds for Clang 23.1.0, where clang's analyzer finds that `s_debugLoggingEnabled` is only written to, not read and throws an error.
2. Fixes viable patterns silently being dropped if the filename matches, even if they are in sub-directories and are different patterns.
3. Fixes `#pragma once` silently dropping the next `#pragma` if it's not followed by a blank newline (WerWolv/PatternLanguage#249) (this commit has been dropped in this PR because it would make merging more complicated, but the fix is still valid on the PatternLanguage PR)
4. Fixes an ImHex empty ID assertion that is hit when temporary project files are removed from the project view while in a new project. This somehow clears the displayName field of the recent entry, which crashes on the assertion. 
5. Fixes an out of bounds access when a malformed pattern is pasted ontop of another malformed pattern
6. Fixes a json exception when saving the pattern
7. Fixes a regression causing ImHex to append the pattern to the current pattern file.

### DIFF
--- a/lib/libimhex/include/hex/helpers/file_attached_data.hpp
+++ b/lib/libimhex/include/hex/helpers/file_attached_data.hpp
@@ -32,6 +32,9 @@ namespace hex {
             }
 
             void erase() {
+                if (m_json->type() == nlohmann::json::value_t::null) {
+                    return;
+                }
                 m_json->erase(getPathKey());
                 save();
             }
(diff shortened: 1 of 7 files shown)


neptuwunium:
Force pushes are rebasing to tip and dropping the submodule commit for fix 3.

neptuwunium:
implicitly merged in #2888

PR #2882, opened 2026-09-06
feat: Toggle comments on the selected lines with Ctrl+/

scottanderson:
### Problem description
Comments or uncomments every line the selection touches, or the cursor's own line when nothing is selected.

Uncomments only when every line with text on it is already commented. A partly commented run becomes wholly commented, which one more press then clears. A blank line takes no token and does not decide the direction. The token goes at the shallowest indent in the run, so the block keeps its shape.

### Additional things
The token comes from the language definition, not from a hard coded "//", so the shortcut does nothing in a language that declares none.

Put the text rule in a new header, ui/line_comment.hpp, so it can be tested on its own. TextEditor::toggleLineComment() only reads the lines, calls it, and writes the result back as one undo record, the same shape doPaste() builds. A single Ctrl+Z therefore takes the comment back off every line at once.


WerWolv:
This one is for you @paxcut

PR #2888, opened 2026-09-08
feat: pattern formatting on save

neptuwunium:
Depends on #2879 and WerWolv/libwolv#47

Adds 3 new settings to:

 - convert spaces to tabs on save
 - trim end-of-line whitespace (with tabs, needs support for spaces)
 - add final newline

This is mostly for people that have the same text formatting paradigm as me.

While it would be better to support tabs in the editor at all,  ImHex's text editor has a lot of assumptions that spaces are always used, and ImGui's lack of tab support in the text glyph rendering system system makes it akin to rewriting the entire text editor line system.

Also makes pasted/loaded code use the correct tab width.


plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -1776,20 +1802,17 @@ namespace hex::plugin::builtin {
...
-        file.writeString(code);
-        file.flush();
-
+        std::string code;
         if (trackFile) {
             if (!m_sourceCode.bind(provider, path))
                 return;
             code = m_sourceCode.get(provider);
         } else {
+            wolv::io::File file(path, wolv::io::File::Mode::Write);
```
paxcut: Shouldn't this be Read mode?

plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -259,6 +260,31 @@ namespace hex::plugin::builtin {
...
+        const auto trimWhitespace = m_formattingTrimWhitespace.get();
+        const auto insertFinalNewline = m_formattingFinalNewline.get();
+        if (!shouldConvertToTabs && !trimWhitespace && !insertFinalNewline) {
+            return code;
+        }
+
+        auto formattedCode = code;
+
+        if (shouldConvertToTabs) { // spaces -> tabs
+            formattedCode = wolv::util::replaceSpacesWithTabs(code, 4, trimWhitespace);
```
paxcut: Wouldn't it be better if we used the value on the settings for tab size instead of hard coding it to 4?

neptuwunium: yes but the value is still hardcoded to 4 in every call site of preprocessText, since prior to the linked PR it would always default to 4.

plugins/builtin/source/content/views/view_pattern_editor.cpp
neptuwunium: yes, sorry. my mistake copy and pasting code around 😨

plugins/builtin/source/content/views/view_pattern_editor.cpp
paxcut: PreprocessText is only used if the file read had tabs in it. most of the pattern read are created in imhex so thats never used. Writing tabs to the file on output generates patterns with tabs that preprocessText needs to handle. So this feature requires the previously low level bug become a must fix.

paxcut:
I haven't had time to develop an idea that occurred to me after our debate on tabs pm discord. you are correct in saying that adding tab support the straight forward way requires a lot of work but what if we were to enable the use of virtual tabs while keeping the internal representation as spaces? That would require no changes in the way we calculate lengths and positions and as far as the user can tell there will be no difference in behavior form actually using tabs.

plugins/builtin/source/content/views/view_pattern_editor.cpp
neptuwunium: https://github.com/WerWolv/ImHex/pull/2888/commits/c32f43ec639fa28e9ba849e20110b65a3b4535ce

should be normalized now

neptuwunium:
That would work

plugins/builtin/source/content/views/view_pattern_editor.cpp
```diff
@@ -259,6 +260,37 @@ namespace hex::plugin::builtin {
...
+        if (insertFinalNewline && !formattedCode.ends_with('\n')) {
+            formattedCode += '\n';
+        }
+
+        return formattedCode;
+    }
+
+    [[nodiscard]] std::string PatternSourceCode::preprocessPattern(const std::string &code) {
+        // todo: trim trailing whitespace, move this somewhere ui can use it as well
+        return wolv::util::preprocessText(code, m_tabSize.get());
```
paxcut: Adding tabSize as a parameter to preprocess text makes you add tabSize as argument to functions that are not related to  formatting like appendLine. PatternSourceCode was created to allow providers to share patterns. With the new project it is used to handle the file attachment and adding variables and functions unrelated to that role makes the code more complex and it is unnecessary. 
Instead do to utils what we are doing to pattern editor to pass the value of tabSize to the functions in there.

neptuwunium: 1. It always had assumed tabSize was 4, it was always calling `wolv::util::preprocessText` on appendLine which was silently calling replaceTabsWithSpaces with an indent level of 4; the old code was:
```cxx
    std::string preprocessText(const std::string& code) {
        std::string result = replaceStrings(code, "\r\n", "\n");
        result = replaceStrings(result, "\r", "\n");
        result = replaceTabsWithSpaces(result, 4);

        return result;
    }
```
All that was changed is that 4 was no longer hard coded and used as an arg instead.
Arguably the call site can be removed entirely from some places, but that's out of scope for this PR.
2. It's a static, not an instance method. It was just a convenient place to put it but it shares no code access to the rest of `PatternSourceCode`. It can be moved, which is what the todo implied.
3. The primary call site is from `PatternSourceCode.m_perProviderSource.decode`, which would mean to pass `m_tabSize` from that lambda we would have to bind `this` to the lambda and synchronize the `m_tabSize` value from the `m_textEditor`. This also means that formatting and preprocessing is still within scope of the class as well, since it is responsible for encoding and decoding data.
Realistically encode/decode should be moved out of `PatternSourceCode`, but requires more design changes which are out of scope for this PR.

paxcut: From your answers specially from 3) it seems I wasn't clear as to what I was proposing. If you reread it you will see that it isn't needed to have tabSize inside any of the functions of PatterSourceCode. Also preprocessing, formatting and encoding/decoding are unrelated operations that need to be kept separated as they currently are.  

I used the example of the TextEditor handling of tabsize but that wasn't enough. this time  i'll write the code explicitly so there are no misunderstandings. First the code in libwolv string utilities cpp file. Add a variable in the `wolv::util` namespace and a setter to modify it from anywhere in imhex that includes `wolv/util/string.hpp` where the getter's signature is added. Then substitute the literal with the variable

```cpp
namespace wolv::util {
   u32 n_tabSize;
   u32 setTabSize(u32 tabSize) { n_tabSize = tabSize; }
    .
    .
    .
    std::string preprocessText(const std::string& code) {
        std::string result = replaceStrings(code, "\r\n", "\n");
        result = replaceStrings(result, "\r", "\n");
        result = replaceTabsWithSpaces(result, n_tabSize);

        return result;
    }
    .
    .
    .
}
```

The changes to view_pattern_editor.cpp are simpler. Only one line is added and no further changes are necessary to any of the existing code.
    
 ```cpp
ContentRegistry::Settings::onChange("hex.builtin.setting.pattern_editor"_unlocalized, "hex.builtin.setting.pattern_editor.tab_size"_unlocalized, [this](const ContentRegistry::Settings::SettingsValue &value) {
    if (ImHexApi::Provider::isValid())
        m_textEditor.get(ImHexApi::Provider::get()).setTabSize(value.get<u32>(4));
    wolv::util::setTabSize(value.get<u32>(4));
});
```
That's it. nothing else needs to change. You can add the rest of the features without modifying existing code.
I can't imagine any argument that can be made against this when you take into account that it is a minimal implementation only so don't object to the global variable and embed it in a container instead.  This is the way settings are propagated to places that can't access the settings api everywhere in imhex.

Comment (neptuwunium) [reply plugins/builtin/source/content/views/view_pattern_editor.cpp]: > Also preprocessing, formatting and encoding/decoding are unrelated operations that need to be kept separated as they currently are.

*It already did this before the PR.*

Decode *would preprocess the text*

Comment (neptuwunium) [reply plugins/builtin/source/content/views/view_pattern_editor.cpp]: >  Add a variable in the wolv::util namespace and a setter to modify it from anywhere in imhex that includes wolv/util/string.hpp where the getter's signature is added. Then substitute the literal with the variable

this is an antipattern because now you're introducing a global in a shared library that can be used by multiple different projects that have different tab sizes.

Comment (neptuwunium) [reply plugins/builtin/source/content/views/view_pattern_editor.cpp]: What you're asking is changing fundamental behaviors within both libwolv that makes it act in global space (which is bad) and has implications across multiple consumers, for example a plugin that loads libwolv and calls that setTabSize method *will also end up controlling ImHex's tab size*. Tab Size is a variable owned and controlled by ImHex, it should remain in ImHex's code and memory space.

I will not do this. This is a bad design.

Comment (neptuwunium) [reply plugins/builtin/source/content/views/view_pattern_editor.cpp]: 90335859b404d747deb1ff91925069409386b6a8 moves things around, i could remove m_tabSize from PatternSourceCode, but you would end up with 8 more calls to preprocessPattern and formatPattern

along with silly stuff like:

```
code = preprocessPattern(file.readString(), m_textEditor.get(provider).getTabSize());
m_sourceCode.set(provider, formatPattern(code, m_textEditor.get(provider).getTabSize()));
```

paxcut: It is much better now. Not perfect as in the call to doPaste still needing to handle tab sizes but as I am further developing the reintroduction of tabs to the pattern editor I'll probably need to deal with that internally. As for respect of this PR I am now ready to merge. What follows are thoughts about moving forward considering that tabs may be supported soon and don't apply to this PR directly so it may be skipped.

The way I see it now is that the function to convert ends of line and the functions to convert spaces/tabs need to be separated because one is always applied but the other will depend on what settings are chosen. 

If you choose not to convert tabs to spaces in the settings then you wont need to call either of tabs to spaces or spaces to tabs. If you choose to convert then you probably don't want to save as tabs because you can simply use tabs always. Conversion to and from tabs to and from spaces also need shortcuts that can work on selections or on the entire file.Also show white spaces needs to properly handle tabs display.

PR #2889, opened 2026-09-08
feat: Add an encoding pipeline for the hex editor and pattern language

scottanderson:
Adds support for PL encoding to ImHex.

### Description

A string pattern decodes through its own `[[encoding]]` attribute, the document encoding, or UTF-8, in that order. Add `hex::dec::encode()`, `hex::dec::decode()`, and a declared-encoding column in the hex editor.

Add `EncodingFile` support for `.tbl` tables, plus UTF-8, UTF-16 and UTF-32. `decodeUtf8Bounded()`, `decodeUtf16Bounded()` and `decodeUtf32Bounded()` decode up to a code point limit and report why they stopped (`DecodeStop`). `PatternLanguageStringCodec::decode()` needs that for the tree view display budget.

Add `Codepage` (`hex/helpers/codepage.hpp`), a single-byte character set. The hex editor text column reads the data with it. A multi-byte encoding cannot drive a one byte cell, so the column stays on ASCII for those.

Name only "UTF-16LE"/"UTF-16BE"/"UTF-32LE"/"UTF-32BE", never a bare "UTF-16"/"UTF-32". Those name the BOM-carrying encoding schemes of the Unicode Standard, and nothing here reads or writes a BOM, so accepting them would silently guess the byte order.

Add `escapeCodepoint()`, escaping one decoded codepoint for display: a non-printable ASCII byte becomes `\xNN`; a codepoint above ASCII with no glyph becomes `\uNNNN`, or `\UNNNNNNNN` past U+FFFF; all else passes through. Add `escapeControlCharacters()`, applying this to a whole decoded string.  It returns `std::nullopt` on invalid UTF-8, since a bad byte cannot round-trip as an escape; the caller shows "Invalid" instead.

Teach `hex::decodeByteString()` the `\0`, `\u` and `\U` escapes, so everything `escapeCodepoint()` writes can be read back.

Uses the pluggable text codec `pattern_language` already provides.

### Table files

A `.tbl` value can now name a code point with `\uXXXX`. A table can then map a byte to a character with no glyph of its own, such as a control code. `\\` is one literal backslash. Any other backslash stands for itself, so a table that already holds one keeps working.

A table means what it says. A byte with no line of its own decodes to ".". There is no automatic ASCII for a byte in 0x00-0x7F.

Three changes are necessary in ImHex-Patterns:

- 27 of the 38 tables give a control code a name, such as `00=NUL`. That byte now decodes to the text "NUL". The line must become `00=\u0000`.
- `big5.tbl` and `jis_x_0213.tbl` name no byte below 0x80. They need their ASCII lines to work as a document encoding.
- A name resolves to a file stem only. There is no alias table, so each file needs its official name.

### Dependencies

- https://github.com/WerWolv/PatternLanguage/pull/246
- https://github.com/WerWolv/PatternLanguage/pull/250
- #2883

### DIFF
--- a/lib/libimhex/include/hex/helpers/encoding_file.hpp
+++ b/lib/libimhex/include/hex/helpers/encoding_file.hpp
@@ -2,12 +2,19 @@
 
 #include <hex.hpp>
 
+#include <array>
+#include <bit>
+#include <functional>
 #include <map>
+#include <optional>
+#include <string>
 #include <string_view>
 #include <vector>
 #include <span>
 
 #include <wolv/io/fs.hpp>
+
+#include <pl/core/string_encode_decode.hpp>
 
 namespace hex {
 
@@ -31,11 +38,39 @@ namespace hex {
         EncodingFile& operator=(const EncodingFile &other);
         EncodingFile& operator=(EncodingFile &&other) noexcept;
 
+        // Works like getEncodingFor(). Returns std::nullopt when the encoding has
+        // no entry for these bytes.
+        [[nodiscard]] std::optional<std::pair<std::string_view, size_t>> lookup(std::span<const u8> buffer) const;
+
         [[nodiscard]] std::pair<std::string_view, size_t> getEncodingFor(std::span<const u8> buffer) const;
         [[nodiscard]] u64 getEncodingLengthFor(std::span<u8> buffer) const;
         [[nodiscard]] u64 getShortestSequence() const { return m_shortestSequence; }
         [[nodiscard]] u64 getLongestSequence()  const { return m_longestSequence;  }
         [[nodiscard]] std::string decodeAll(std::span<const u8> buffer) const;
+
+        // Whether every byte in `buffer` decodes to a known value in this
+        // encoding, with no leftover unmapped byte. False for a buffer
+        // getEncodingFor() would otherwise paper over with a "." placeholder.
+        [[nodiscard]] bool isFullyMapped(std::span<const u8> buffer) const;
+
+        // Decodes `buffer` one entry at a time, stopping at the first byte
+        // sequence with no entry, at the end of `buffer`, or after maxCodepoints
+        // entries - whichever comes first. Unlike isFullyMapped() + decodeAll(),
+        // this tells apart a buffer that is simply too short for even the
+        // shortest mapped sequence (DecodeStop::EndOfInput) from one with a byte
+        // sequence this encoding does not know (DecodeStop::MalformedBytes).
+        [[nodiscard]] pl::core::DecodeResult decodeBounded(std::span<const u8> buffer, std::optional<size_t> maxCodepoints = std::nullopt) const;
+
+        // Returns std::nullopt if `sequence` does not start with a known encoded value.
+        [[nodiscard]] std::optional<std::pair<std::vector<u8>, size_t>> getBytesFor(std::string_view sequence) const;
+
+        // False when one decoded value maps to more than one byte sequence, or one
+        // decoded value is a prefix of another.
+        [[nodiscard]] bool canEncode() const { return !m_ambiguousEncoding; }
+
+        // Returns std::nullopt when the encoding is ambiguous (see canEncode()), or
+        // when `sequence` has a character with no byte value in this encoding.
+        [[nodiscard]] std::optional<std::vector<u8>> encodeAll(std::string_view sequence) const;
 
         [[nodiscard]] bool valid() const { return m_valid; }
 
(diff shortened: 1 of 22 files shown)


lib/libimhex/include/hex/helpers/encoding_file.hpp
```diff
@@ -31,12 +38,40 @@ namespace hex {
         EncodingFile& operator=(const EncodingFile &other);
         EncodingFile& operator=(EncodingFile &&other) noexcept;
 
+        // Works like getEncodingFor(). Returns std::nullopt when the encoding has
```
WerWolv: Use doxygen style comments

lib/libimhex/include/hex/helpers/encoding_file.hpp
```diff
@@ -51,9 +86,98 @@ namespace hex {
...
         std::unique_ptr<std::map<size_t, std::map<std::vector<u8>, std::string>>> m_mapping;
+        std::unique_ptr<std::map<size_t, std::map<std::string, std::vector<u8>, std::less<>>>> m_reverseMapping;
 
         u64 m_shortestSequence = std::numeric_limits<u64>::max();
         u64 m_longestSequence  = std::numeric_limits<u64>::min();
+
+        bool m_ambiguousEncoding = false;
+    };
+
+    // A single-byte character set: one byte maps to at most one character, for
```
WerWolv: I think this stuff should rather be in a separate file. encoding_file is primarily the abstraction above thingy table files

lib/libimhex/include/hex/helpers/string_codec.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace hex {
+
+    // Decodes and encodes a PatternString's bytes for the pattern language runtime.
+    // This is the value a pattern script sees, through ==, std::print, and other
+    // operations.
+    class ImHexStringCodec : public pl::core::StringEncodeDecode {
```
WerWolv: Would call this PatternLanguageStringCodec

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -1,13 +1,222 @@
...
+#include <mutex>
 #include <ranges>
 #include <wolv/io/file.hpp>
 #include <wolv/utils/string.hpp>
 
 namespace hex {
 
+    namespace {
+
+        // Bytes 0x00-0x7F always mean standard ASCII. Bytes 0x00-0x1F and 0x7F get
```
WerWolv: Same here, this should be in its own file I think

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -1,13 +1,222 @@
...
+                return { Utf8CodepointStatus::Invalid, 0, 0 };
+
+            return { Utf8CodepointStatus::Complete, length, codepoint };
+        }
+
+        // Maps a standard encoding name, an IANA name or a name a pattern author
+        // reaches for, to the ImHex table file that implements it. ImHex names
+        // its tables after their purpose, not the encoding. Remove an entry once
+        // its file is renamed to the standard name in ImHex-Patterns.
+        constexpr static auto EncodingNameAliases = std::to_array<std::pair<std::string_view, std::string_view>>({
```
WerWolv: Remove

I'm happy to rename the existing thingy files once this PR is merged using the official names

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -167,11 +471,449 @@ namespace hex {
                 m_mapping->insert({ fromBytes.size(), {} });
 
             u64 keySize = fromBytes.size();
+
+            bool isStandardAsciiEntry = keySize == 1 && fromBytes[0] <= 0x7F;
```
WerWolv: I don't think we should special-case certain text sequences. Instead, as discussed, we should add an extension to table files that allows using raw codepoints in the form of `\uXXXX` on the right hand side and use those if a non-printable character is desired

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -167,11 +471,449 @@ namespace hex {
...
+                // "Invalid".
+                return std::nullopt;
+            }
+
+            const bool isLast = offset + length == text.size();
+
+            if (codepoint == 0) {
+                // A lone trailing NUL reads as a normal C-string end. Show
+                // "\0", not the general \x00 escape.
+                result += (!seenNull && isLast) ? "\\0" : "\\x00";
```
WerWolv: We use \x00 everywhere else. I don't think we should special case the end of a string

plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -647,7 +654,105 @@ namespace hex::plugin::builtin {
...
+            this->applyEncoding(provider);
+        });
+
+        // A commented-out pragma fires nothing. Reads the code directly instead
+        // of a flag from the pragma's own handler; see declaredEncodingPragmaValue().
+        EventPatternExecuted::subscribe(this, [this](const std::string &code) {
+            this->resetEncodingIfNotDeclared(code);
+        });
+
+        // The pragma applies on every keystroke via the editor's syntax
```
WerWolv: We don't want the encoding to change by just modifying the source code and we definitely do not want the preprocessor to run on every keystroke

plugins/builtin/source/content/views/view_hex_editor.cpp
```diff
@@ -647,7 +654,105 @@ namespace hex::plugin::builtin {
...
+            if (encodingName == name)
+                return;
+
+            encodingName = name;
+            this->applyEncoding(provider);
+        });
+
+        // A commented-out pragma fires nothing. Reads the code directly instead
+        // of a flag from the pragma's own handler; see declaredEncodingPragmaValue().
+        EventPatternExecuted::subscribe(this, [this](const std::string &code) {
```
WerWolv: This should be handled through the main runtime's encoding #pragma handler instead of re-parsing the entire source file _again_ just to get the encoding name

plugins/ui/source/ui/hex_editor.cpp
```diff
@@ -579,6 +610,10 @@ namespace hex::ui {
 
                 } else {
                     asciiVisualizer.enableExtendedAscii(m_showExtendedAscii);
+                    // A multi-byte encoding, such as Shift-JIS or UTF-8, cannot fit in a
+                    // one-byte-wide ASCII cell. The separate custom encoding column still
+                    // handles those.
+                    asciiVisualizer.setCodepage(&m_codepage, m_codepageDeclared);
```
WerWolv, line 609: The ASCII visualizer is meant to do exactly what's on the tin: Visualize ASCII. There shouldn't be any other encodings in there. Anything special goes in the Custom Encoding column

WerWolv (changes requested):

plugins/ui/source/ui/hex_editor.cpp
scottanderson: If the document is EBCDIC, an ASCII representation of those bytes will misrepresent the contents of the file. That is what makes having an ASCII representation on screen when the encoding is not set to ASCII incorrect. Fixing this is the central feature of this PR.

I can update the variable name and comment to reflect the current behavior, but my understanding was that we had already agreed to add codepage support to the visualizer. Since this is the central feature of the PR, it's not something I'm going to remove.

WerWolv (approved):

PR #2893, opened 2026-09-10
fix: segfault with error markers at obscenely high framerates

neptuwunium:
line gets deleted mid render, apparently making the iterator increment to an invalid out of bounds value.

this is just a hotfix, it should be pinned down where the error lines get mutated during a render.

### DIFF
--- a/plugins/ui/source/ui/text_editor/render.cpp
+++ b/plugins/ui/source/ui/text_editor/render.cpp
@@ -1682,9 +1682,10 @@ namespace hex::ui {
 
         ErrorMarkers::iterator errorIt;
         auto errorHoverBoxKey = lineStart + Coordinates(1, 1);
-        if (errorIt = std::find_if(m_lines.m_errorMarkers.begin(), m_lines.m_errorMarkers.end(), [&](const auto &item) {
+        auto errorMarkers = m_lines.m_errorMarkers; // copy so lines can't be deleted mid flight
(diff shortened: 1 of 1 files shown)


plugins/ui/source/ui/text_editor/render.cpp
```diff
@@ -1682,9 +1682,10 @@ namespace hex::ui {
 
         ErrorMarkers::iterator errorIt;
         auto errorHoverBoxKey = lineStart + Coordinates(1, 1);
-        if (errorIt = std::find_if(m_lines.m_errorMarkers.begin(), m_lines.m_errorMarkers.end(), [&](const auto &item) {
+        auto errorMarkers = m_lines.m_errorMarkers; // copy so lines can't be deleted mid flight
```
WerWolv, line 1685: Doesn't this just move the problem further up? Like, the modification can still happen during the copy so if another thread modifies the error markers, this would fail the same way

neptuwunium: Not sure, this resolved the issue for me so if what you say is true it just made the race condition window smaller.

WerWolv: Yeah, I believe this just makes the race less likely to happen but the race is still there

neptuwunium:
closing since #2900 seems to attempt to fix it

PR #2895, opened 2026-09-10
refactor: Store one Codepoint per byte in a Codepage

scottanderson:
A Codepage gives one byte one cell. A `std::string` entry could hold more than one Codepoint, which does not fit. Store a `char32_t`, where -1 means the byte has no character.

`Codepage::fromEncoding()` now rejects a table with a byte that draws as more than one Codepoint. Before, it dropped only that byte. This covers a control code named as text, such as `"00=NUL"`. A table writes the escape `"00=\u0000"` instead. The table stays valid for the custom encoding column, which decodes with the `EncodingFile` itself.

### DIFF
--- a/lib/libimhex/source/helpers/codepage.cpp
+++ b/lib/libimhex/source/helpers/codepage.cpp
@@ -11,9 +11,8 @@ namespace hex {
         static const Codepage asciiCodepage = [] {
             Codepage result;
 
-            // A control code has no character to draw, so its entry stays empty.
-            for (size_t byte = 0x20; byte < 0x7F; byte += 1)
-                result.m_characters[byte] = std::string(1, char(byte));
+            for (size_t byte = 0; byte <= 0x7F; byte += 1)
+                result.m_characters[byte] = Codepoint(byte);
 
             return result;
         }();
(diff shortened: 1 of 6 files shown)


plugins/ui/source/ui/hex_editor.cpp
```diff
@@ -36,57 +36,32 @@ namespace hex::ui {
...
                         "\u2410", "\u2411", "\u2412", "\u2413", "\u2414", "\u2415", "\u2416", "\u2417",
                         "\u2418", "\u2419", "\u241A", "\u241B", "\u241C", "\u241D", "\u241E", "\u241F",
                     };
 
-                    ImGuiExt::TextFormattedDisabled(c == 0x7F ? "\u2421" : ControlCharacters[c]);
+                    ImGuiExt::TextFormattedDisabled(character == 0x7F ? "\u2421" : ControlCharacters[character]);
                     return;
                 }
 
-                // CP1252 fills undefined high bytes, but never over a declared table.
```
WerWolv, line 60: Doesn't this break the Extended ASCII toggle option? Or where does it get the codepoints from now?

scottanderson: This option controlled two features: (1) replacing control codes (for ASCII, `0x00-0x19` and `0x7F`) with their corresponding control pictures (`␀`, `␁`, etc.), while still keeping them grayed out, and (2) toggling between ASCII and CP1252 rendering of 0x80-0xFF.

The latter behavior is now handled by codepage selection. If you want `0x80-0xFF` interpreted as CP1252, you can switch the codepage from ASCII to CP1252, which in turn will affect how strings render throughout all of ImHex.

WerWolv: I think we should rename that option (and variables) then to something else, not sure what though exactly.
Displaying extended ASCII was the primary purpose of the setting which isn't there anymore now

scottanderson: I agree, but on reflection I think this should be split into its own PR. I'll remove the commit that changes this from this PR, and create a new PR that incorporates this feedback.

scottanderson: Split into #2902

lib/libimhex/source/helpers/codepage.cpp
```diff
@@ -11,9 +11,8 @@ namespace hex {
         static const Codepage asciiCodepage = [] {
```
WerWolv: This can be `constexpr`

PR #2896, opened 2026-09-12
feat: Table file dialog widget and directives

scottanderson:
## Directives

A header of directives now comes before the first entry.

```properties
# Generated by ImHex-Patterns/scripts/generate_encodings.py
-name Windows 1252
-description Windows encoding for Western European languages (English, French, German, and others)
-include includes/windows_latin.tbl
8A=Š
8E=Ž
98=˜
...
```

`-include path` brings in another table's entries. `path` is relative to the table that names it, and names its own extension, such as `includes/box_drawing.tbl`. It can go up a folder with `..`, but never past the encodings search paths.

`-alias name` makes a table another name for another table. Such a table holds nothing else.
```properties
# Generated by ImHex-Patterns/scripts/generate_encodings.py
-alias cp1252
```

The header ends at the first entry. A line that starts with `-` below it makes the table invalid.

## Widget

The custom encoding dialog showed only the file's name. A new widget draws the description above the name, smaller and dimmer, with every other name the table answers to. Each table gets one row, since an alias gives its name to the table it links to.

The header rule keeps the dialog fast. The widget reads only the start of each file, so a table with many thousands of entries costs no more than a small one. `PopupNamedFileChooserBase` draws a row through a new `drawEntry()`, which any chooser can replace.

## Screenshot

image

## Dependencies

- #2895
- https://github.com/WerWolv/ImHex-Patterns/pull/583

### DIFF
--- a/lib/libimhex/source/helpers/encoding_file.cpp
+++ b/lib/libimhex/source/helpers/encoding_file.cpp
@@ -85,6 +85,32 @@ namespace hex {
             return std::nullopt;
         }
 
+        /**
+         * @brief Maps the lower case of every table file's name to that file
+         *
+         * Read once, on the first name that no file answers to by its own spelling. A base path
+         * earlier in the list wins, the same way findEncodingFile() takes the first it finds.
+         */
+        const std::map<std::string, std::fs::path, std::less<>>& encodingFilesByLowerCaseName() {
+            static const auto files = [] {
+                std::map<std::string, std::fs::path, std::less<>> result;
+
+                for (const auto &basePath : paths::Encodings.read()) {
+                    std::error_code error;
+                    for (const auto &entry : std::fs::directory_iterator(basePath, error)) {
+                        if (entry.path().extension() != ".tbl")
+                            continue;
+
+                        result.emplace(toLower(entry.path().stem().string()), entry.path());
+                    }
+                }
+
+                return result;
+            }();
+
+            return files;
+        }
+
     }
 
     namespace impl {
@@ -380,16 +406,22 @@ namespace hex {
 
         std::scoped_lock lock(mutex);
 
-        if (const auto entry = encodings.find(name); entry != encodings.end())
-            return entry->second.valid() ? &entry->second : nullptr;
-
-        // An encoding name is conventionally case-insensitive.
+        // An encoding name is conventionally case-insensitive, so one entry serves every case.
         const auto lowerCaseName = toLower(name);
+
+        if (const auto entry = encodings.find(lowerCaseName); entry != encodings.end())
+            return entry->second.valid() ? &entry->second : nullptr;
 
         auto path = findEncodingFile(name);
 
-        if (!path.has_value()) {
+        if (!path.has_value())
             path = findEncodingFile(lowerCaseName);
+
+        // A file named in any other case answers too, which costs one read of the directory.
+        if (!path.has_value()) {
+            const auto &files = encodingFilesByLowerCaseName();
+            if (const auto file = files.find(lowerCaseName); file != files.end())
+                path = file->second;
         }
 
         EncodingFile encoding;
@@ -397,7 +429,7 @@ namespace hex {
             encoding = EncodingFile(EncodingFile::Type::Thingy, *path);
 
         // A failed lookup is cached too, so a bad name hits the file system once.
-        const auto &result = encodings.emplace(name, std::move(encoding)).first->second;
+        const auto &result = encodings.emplace(lowerCaseName, std::move(encoding)).first->second;
         return result.valid() ? &result : nullptr;
     }
 
--- a/lib/libimhex/source/helpers/encoding_file.cpp
+++ b/lib/libimhex/source/helpers/encoding_file.cpp
@@ -7,9 +7,11 @@
 #include <hex/helpers/utils.hpp>
 
 #include <algorithm>
+#include <cctype>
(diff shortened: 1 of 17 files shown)


WerWolv:
Will post-pone review of that until #2895 is merged since it contains both changes

WerWolv:
This needs to be rebased now as it conflicts with the changes made by the other PR

plugins/builtin/include/content/helpers/popup_encoding_chooser.hpp
```diff
@@ -0,0 +1,82 @@
...
+            float height = ImGui::GetTextLineHeight();
+            if (!subtitle.empty())
+                height += ImGui::GetTextLineHeight() * SubtitleScale + ImGui::GetStyle().ItemSpacing.y;
+
+            const bool clicked = ImGui::Selectable("##entry", selected, ImGuiSelectableFlags_NoAutoClosePopups, ImVec2(0, height));
+            const auto rowSize = ImGui::GetItemRectSize();
+
+            ImGui::SetCursorPos(startPos);
+
+            fonts::Default().pushBold();
```
WerWolv: Using bold here feels a bit out of place. I'd just use normal text for the title and a slightly smaller font for the description below

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -85,6 +87,295 @@ namespace hex {
...
+                return result;
+            }();
+
+            return files;
+        }
+
+        /**
+         * @brief Reads what follows a directive, when `line` is that directive
+         */
+        std::optional<std::string_view> directiveArgument(std::string_view line, std::string_view directive) {
```
WerWolv, line 241: API-wise, I think it would be nicer if `directive` was just the name of the directive without the `-` in front. If you don't have that `-` at the start, this function will not do the right thing

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -85,6 +87,295 @@ namespace hex {
...
+
+        /**
+         * @brief Maps the encodingFileName() of every table file's name to that file
+         *
+         * Read once, on the first name that no file answers to by its own spelling. A base path
+         * earlier in the list wins, the same way findEncodingFile() takes the first it finds.
+         */
+        const std::map<std::string, std::fs::path, std::less<>>& encodingFilesByName() {
+            static const auto files = [] {
+                std::map<std::string, std::fs::path, std::less<>> result;
```
WerWolv: The `std::less<>` shouldn't be needed for all these.  That's the default

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -85,6 +87,295 @@ namespace hex {
...
+        /**
+         * @brief Replaces every -include line with the entries of the table it names
+         *
+         * An included table's entries go last, so the table keeps every byte it gives a value
+         * of its own. A table a cycle reaches again brings nothing a second time.
+         *
+         * Keeps the other directives of `content` and drops those of an included table, since
+         * a table's name and description are its own.
+         */
+        std::string expandIncludes(std::string_view content, const std::fs::path &context, const IncludeResolver &resolveInclude, std::set<std::string, std::less<>> &included, bool topLevel) {
```
WerWolv: Instead of combining the content of the files, wouldn't it make more sense to call the EncodingFile parser recursively here and combine the resulting mapping at the end?

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -371,34 +696,98 @@ namespace hex {
...
+
+        return true;
     }
 
 
+    std::string encodingFileName(std::string_view name) {
+        std::string result(name);
+
+        for (auto &character : result) {
+            const auto byte = static_cast<unsigned char>(character);
```
WerWolv: `u8`

lib/libimhex/source/helpers/encoding_file.cpp
```diff
@@ -371,34 +696,98 @@ namespace hex {
                 }
             }
         }
+
+        return true;
     }
 
 
+    std::string encodingFileName(std::string_view name) {
```
WerWolv, line 726: As discussed in DMs

plugins/builtin/include/content/helpers/encoding_chooser.hpp
```diff
@@ -0,0 +1,132 @@
...
+     * @brief Reads what the chooser shows for every table among `paths`
+     *
+     * Reads only each table's header, so this costs the same however many entries the tables
+     * hold. A table that is only another name for another one gets no entry of its own. It
+     * gives its name to the table it links to instead, so the chooser lists each table once.
+     *
+     * @param paths The files the chooser could list
+     * @return The files to list, and what to show for each table among them
+     */
+    inline EncodingChoices readEncodingChoices(const std::vector<std::fs::path> &paths) {
```
WerWolv: Implement these in a cpp file. Keeps compile times down a bit

WerWolv (changes requested):

lib/libimhex/source/helpers/encoding_file.cpp
scottanderson: This would require reading the file twice or creating a serializer because `view_hex_editor.cpp` writes `m_tableContent` to the project.

PR #2899, opened 2026-09-12
fix: Guard the map of unlocalized names with a mutex

scottanderson:
A `Lang` writes its unlocalized name into a global map, and `Lang::get()` reads that map. Any thread can build a `Lang`. A background service does it to name its own thread.

Two threads could write to the map together, which loses an entry or breaks the tree. Take a mutex in all three places that reach it.

`get()` gives back a pointer into the map. That stays good after the lock goes, since a map keeps each string in place.

### DIFF
--- a/lib/libimhex/source/api/localization_manager.cpp
+++ b/lib/libimhex/source/api/localization_manager.cpp
@@ -185,7 +185,12 @@ namespace hex {
 
     static AutoReset<std::map<std::size_t, std::string>> s_unlocalizedNames;
 
+    // Any thread can name itself, so any thread can reach the map below.
+    static std::mutex s_unlocalizedNamesMutex;
+
     Lang::Lang(std::string_view unlocalizedString) : m_entryHash(LangConst::hash(unlocalizedString)) {
+        std::scoped_lock lock(s_unlocalizedNamesMutex);
+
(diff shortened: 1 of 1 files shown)


lib/libimhex/source/api/localization_manager.cpp
```diff
@@ -185,14 +185,21 @@ namespace hex {
 
     static AutoReset<std::map<std::size_t, std::string>> s_unlocalizedNames;
 
+    // Any thread can name itself, so any thread can reach the map below.
+    static std::mutex s_unlocalizedNamesMutex;
+
     Lang::Lang(std::string_view unlocalizedString) : m_entryHash(LangConst::hash(unlocalizedString)) {
+        std::scoped_lock lock(s_unlocalizedNamesMutex);
```
WerWolv: Use `std::lock_guard` for these. `std::scoped_lock` does a whole lot more like deadlock-avoidance and as such is a whole lot slower and generates more code

WerWolv (changes requested):

lib/libimhex/source/api/localization_manager.cpp
Nemoumbra: Character development

PR #2900, opened 2026-09-13
fix: eliminate thread safety data races in pattern highlighting and pattern editor

AkiSakurai:
highlightSourceCode() ran on a background thread but directly called clearErrorMarkers(), setErrorMarkers(), and wrote to m_showConsole, m_lastEvaluationError, and m_lastCompileError, all of which are also accessed by the main thread during rendering.

evaluatePattern() similarly wrote to m_showConsole, m_lastEvaluationError, m_lastCompileError, and m_callStack from a background thread.

- Wrap clearErrorMarkers() calls in TaskManager::doLater
- Move setErrorMarkers() to main thread via doLater with move semantics
- Move m_showConsole/m_lastEvaluationError/m_lastCompileError writes from the background thread to the main thread via doLater
- Change m_callStack from a raw pointer into a copied vector of pl::core::Location to avoid dangling pointer into destroyed runtime


paxcut:
It is not5 possible for both background tasks to run concurrently. syntax highlighter waits for parse pattern to finish if it is running and if a new parsing task is started before highlighting ends it interrupts the highlighting task. All the variables that are accessed from threads should be atomic. no data races can occur.  What makes these changes necessary?

AkiSakurai:
Empirically, it crashes randomly when accessing these variables, but it no longer does so after moving the assignment to the main thread. Yes, I missed the atomic m_lastEvaluationProcessed, and on further review, the fix is incomplete.

neptuwunium:
> no data races can occur. What makes these changes necessary?

there is a data race somewhere though, the m_errorMarkers vector gets changed mid-render, mid-loop causing an out of bounds read and segfault. it also seems to happen more frequently when frame time is sub millisecond (>1000 FPS) basically got it to hit almost every single time an error marker was changed at a framerate over 2000.

paxcut:
ok, i was talking about the variables mentioned in the description only. I recently fixed (not yet committed) a bug where the colors were updated  in the thread while the number of lines changes in the main thread. i'll have to look at the errormarkers but i seem to remember making them copies to avoid using the ones that can be updated in a thread. maybe i missed something or some changed afterwards.

AkiSakurai:
[tsan_stderr.14452.log](https://github.com/user-attachments/files/32158961/tsan_stderr.14452.log)
Here is the TSAN report from my machine, in case it helps.

AkiSakurai:
[tsan_stderr.42710.log](https://github.com/user-attachments/files/32159525/tsan_stderr.42710.log)
Here is the TSAN report from my machine, in case it helps.
build from WerWolv/ImHex@2ff18a5e2f2d8e1f9aceb1fc284d16d3e6a8838f

paxcut:
it does help thank you. For what I can see all errors are on the error markers as stated by @neptuwunium  above. Error markers are generated in the parser thread, the higlighter thread which cannot be concurrent then should make a copy of them so that the main thread can use them to display. The copies are made in a variable called requiredInputs that the text editor should be using instead of using the variables that the parser thread uses but it seems that i neglected to implement the last step and that makes the race condition possible.

paxcut:
This turned out to be a very big mess of problems that are for the most part totally unnecessary. Upon closer inspection the errors come from calling textEditor::clearErrorMarkers from the highlighter thread and  the main thread. Originally the thread was doing the rendering of the errors but when that was moved to the editor the code in the highligther should have been removed. Error markers are generated on the parser or the evaluator thread, then copied over the texteditor when either are finished. As for the variables m_showConsole, m_lastEvaluationError, and m_lastCompileError,  they are never accessed from the highlighter and i don't see them mentioned in the asan log.

PR #2902, opened 2026-09-14
feat: Rename the extended-ASCII toggle to control pictures

scottanderson:
## Dependencies

- #2895

## Summary

The toggle named "extended ASCII" no longer toggles between ASCII and CP1252, as Codepage selection does that job now. The toggle now only swaps a dot for a control-code picture on control codes (0x00-0x1F, 0x7F), a behavior which was previously not explained. This PR renames the toggle to match its remaining job.

- Renamed the variables, functions, the settings key, and the lang key.  "extendedAscii"/"extended_ascii" becomes "controlPictures"/ "control_pictures". The new name matches the Unicode block the glyphs come from.
- Reworded the English tooltip. It said "Display unprintable characters in ASCII column." It now says "Display control-code pictures instead of a dot." This matches what the toggle does today.
- Dropped the hu_HU, ko_KR, uk_UA, and zh_CN strings for this tooltip. Each one answered the old question. Localization falls back to English for a missing key, until a translator gives it a new string.

## Compatibility note

The settings key change resets this toggle to its default for existing users. It holds one boolean and no other state, so this PR adds no migration for it.

### DIFF
--- a/lib/libimhex/include/hex/helpers/unicode.hpp
+++ b/lib/libimhex/include/hex/helpers/unicode.hpp
@@ -21,11 +21,19 @@ namespace hex {
     bool isSingleCharacter(std::string_view text);
 
     /**
-     * @brief Checks whether `byte` is a standard-ASCII control code
-     * @param byte The byte to check
-     * @return Whether the byte is a control code, which has no glyph
+     * @brief Reads `text` as exactly one whole, valid UTF-8 code point
+     * @param text The text to read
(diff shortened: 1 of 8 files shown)


lib/libimhex/include/hex/helpers/unicode.hpp
pankajsartape07-crypto: For the control code is a working from the access control panel

PR #2903, opened 2026-09-15
feat: Add export differences button to diffing view

ddmitv:
### Problem description
It's currently impossible to export difference data from the diffing view.

This PR reworks the export formatter system (the `hex::ContentRegistry::DataFormatter::addFindExportFormatter` and other related functions) and makes it more generalized (no longer tied specifically to Find view).
Used the newly reworked system to add an "Export differences" button in the diffing view.

It also contains a small regression in the Find view's CSV/TSV export. Previously, it output the `offset` column in `0x{:08X}` format and the `size` column in `0x{}` format, now it outputs them as plain decimal numbers. I think this is a good change if they are intended to be parsed.

Other plugins can add new export formatters (like YAML, TOML or whatever) and these will be added to the export options of both the Find and Diffing views.

### Screenshots

image
image


WerWolv:
Thank you!

PR #2904, opened 2026-09-16
impr: Refuse invalid Data Inspector edits and guard against decode crashes

scottanderson:
## Problem

A Data Inspector row could not differentiate "no bytes" from "invalid input", so bad input committed an empty value instead of being refused.

A row's display or edit function can also throw: a custom inspector runs pattern language code, and a pattern-backed row can decode invalid bytes. The throw crossed ImGui's style and table stacks, leaving them unbalanced and crashing ImHex.

## Implementation

Return `std::optional` from `EditWidget::Widget::Function` and `hex::decodeByteString()`. Invalid text keeps edit mode open with a red border, checked on every keystroke. Trim the edit buffer to the typed length each frame, so it cannot grow without bound.

Catch at each site that can throw, log, and show "Invalid". The pattern tree's "Copy Value" also calls `toString()` inside a table, so guard that too.


plugins/builtin/source/content/views/view_data_inspector.cpp
```diff
@@ -486,8 +487,20 @@ namespace hex::plugin::builtin {
...
 
-            // Render inspector row value
-            const auto &copyValue = entry.displayFunction();
+            // A custom inspector's display function runs pattern language code, which can throw.
+            std::string copyValue;
+            try {
+                copyValue = entry.displayFunction();
+                entry.displayErrorLogged = false;
+            } catch (const std::exception &e) {
+                // Log once; a failing row redraws every frame.
```
WerWolv, line 496: Handling this here sounds like a bad idea. Instead, errors from custom inspector rows should be handled at the place where the pattern language library is being invoked. And then that value + potential error value should be cached there and re-used until the row is invalidated. 

I guess having a try-catch block here as well is okay but that shouldn't be the thing catching pattern language errors.

lib/libimhex/source/api/content_registry.cpp
```diff
@@ -907,13 +907,45 @@ namespace hex {
 
         namespace EditWidget {
             std::optional<std::vector<u8>> TextInput::draw(std::string &value, std::endian endian) {
-                if (ImGui::InputText("##InspectorLineEditing", value,
-                                 ImGuiInputTextFlags_EnterReturnsTrue |
-                                 ImGuiInputTextFlags_AutoSelectAll)) {
-                    return getBytes(value, endian);
+                // Sized generously; one typed character can take several UTF-8 bytes.
+                const auto bufferSize = std::max<size_t>(value.size() * 4, 256) + 1;
```
WerWolv: Probably shouldn't try to resize this on every frame but only when new data is actually entered. You should be able to do that in the callback, there's an example in the ImGui Demo for that I think

WerWolv (changes requested):

plugins/builtin/source/content/views/view_data_inspector.cpp
scottanderson: Fixed the misleading comment.

PR #2906, opened 2026-09-16
feat: Add a lossy write path for string pattern values

scottanderson:
## Problem

Committing an edited string pattern value always wrote it exactly as encoded, or failed the write outright when a typed character had no byte under the field's encoding. There was no way to force a write when the exact text could not be represented, and no live signal while typing that a character would not survive the round trip.

## Implementation

Enter commits a string edit only when the encoding represents every character exactly. Shift+Enter force-writes through `PatternString::setValueLossy()`, putting the encoding's replacement character, or `?`, in place of anything it cannot represent.

`PatternLanguageStringCodec::encodeLossy()` builds those bytes one code point at a time rather than through `EncodingFile::encodeAll()`. `encodeAll()` refuses an ambiguous table outright, but a lossy write already accepts approximation, so ambiguity should not block it.

While a field is open, the editor checks every keystroke and marks a character the encoding cannot hold, so the choice between the two commits is visible before either is pressed.

## Dependencies

- #2905


WerWolv:
Needs to be rebased

plugins/builtin/include/content/helpers/decoded_string.hpp
```diff
@@ -0,0 +1,146 @@
...
+     * @brief Extends a selection forward to the end of the code point it lands in
+     *
+     * A click-select size must never end inside a code point cell.
+     *
+     * @param buffer The bytes to decode from
+     * @param targetSize The selection size to extend, in bytes
+     * @param decodeOne Decodes one code point from the front of a span
+     * @return The extended size; past `targetSize` only when a code point straddles it
+     */
+    inline size_t extendToWholeCodePoints(const std::vector<u8> &buffer, size_t targetSize,
```
WerWolv: These functions should all be implemented in cpp files

plugins/builtin/source/content/views/view_data_inspector.cpp
```diff
@@ -130,23 +139,110 @@ namespace hex::plugin::builtin {
...
+                longestSequence = table->getLongestSequence();
+        }
+        const size_t maxSize = DisplayBudget * longestSequence;
+
+        std::vector<u8> buffer(std::min<size_t>(m_validBytes, maxSize));
+        m_selectedRegion.getProvider()->read(m_selectedRegion.getStartAddress(), buffer.data(), buffer.size());
+        preprocessBytes(buffer);
+
+        // Stateless; one shared instance instead of one allocation per row build.
+        static const PatternLanguageStringCodec codec;
```
WerWolv: Making this `static` seems bad. This function runs in the update thread while e.g the editingFunction further below will run in the  main thread

plugins/ui/include/ui/control_byte_picture.hpp
```diff
@@ -0,0 +1,53 @@
...
+     * Not reversible for text that already holds a real U+2400: nothing in
+     * a null-terminated buffer can tell the two apart, so editing such a
+     * string changes it.
+     */
+    constexpr inline char NulPicture[] = { '\xE2', '\x90', '\x80' };
+
+    /**
+     * @brief Replaces each NUL byte with NulPicture, so an edit box can hold the whole value
+     */
+    inline std::string nulToPicture(std::string_view text) {
```
WerWolv, line 26: Would it make sense to replace all control characters that can't be displayed properly in the edit box with their "picture" counterparts?

plugins/ui/source/ui/pattern_value_editor.cpp
```diff
@@ -23,6 +26,35 @@
...
+     * @brief Sets a pattern's value, catching an exception that would otherwise crash the application
+     *
+     * A write formatter can reject input, or a string can fail to encode a
+     * character its encoding has no byte for. An uncaught exception here
+     * would cross a BeginTable()/EndTable() pair and corrupt ImGui's state.
+     *
+     * @param setValue Sets the pattern's value; called once
+     */
+    template<typename F>
+    static void trySetValue(F &&setValue) {
```
WerWolv: Is there a reason to take in a lambda here instead of just a pattern reference plus the value to set it to?

WerWolv (changes requested):

plugins/ui/include/ui/control_byte_picture.hpp
scottanderson: Users can't actually write a string containing the real `␀` codepoint, so this isn't cost-free, and I wouldn't extend this approach to the other control pictures.

plugins/ui/source/ui/pattern_value_editor.cpp
```diff
@@ -161,58 +214,145 @@ namespace hex::ui {
...
+                // toString() can throw on bytes not valid under the field's encoding.
+                m_editingValue = nulToPicture(pattern.toString());
+            } catch (const std::exception &e) {
+                log::error("Failed to decode pattern value: {}", e.what());
+            }
+        }
+        std::string &value = m_editingValue;
+
+        // Sized for UTF-8 bytes of the typed text, not the target encoding; getBytesOf() caps the actual write.
+        const auto bufferSize = std::max(value.size(), pattern.getSize() * 4) + 1;
```
WerWolv: This can also be moved into the resize callback again

plugins/ui/source/ui/pattern_value_editor.cpp
```diff
@@ -161,58 +214,145 @@ namespace hex::ui {
...
             m_onEditCallback();
         }
+        cancelIfDeactivated(submitted);
     }
 
     void PatternValueEditor::visit(pl::ptrn::PatternUnion& pattern) {
-        auto value = pattern.toString();
-        if (ImGui::InputText("##Value", value, ImGuiInputTextFlags_AutoSelectAll | ImGuiInputTextFlags_EnterReturnsTrue)) {
-            pattern.setValue(value);
+        // See the note in visit(PatternStruct&) above.
```
WerWolv: Not a fan of referencing other comments like this. This will be forgotten if we ever change the code above

PR #2907, opened 2026-09-16
fix: Respect NumLock when navigating pattern editor search history

### LINKED ISSUE #2866: [Bug] Using numpad in text search dialog will prefer to behave as arrow keys regardless of NumLock state.
### Operating System

Windows

When typing, for example "u8" in the search box using the numpad, the "8" is interpreted as an up arrow regardless of numlock state.

Hit Ctrl+F,
Type "u" and then press "8" on the numpad.

Observe the search has been cleared OR replaced with your previous searches.

### ImHex Version

Nightly

### ImHex Build Type

### Installation type

MSI

### Additional context?

_No response_

lyn-un:
### Problem description

Fixes #2866.

The Pattern editor's find and replace fields treat keypad 8/2 as history navigation keys even when NumLock is enabled. Entering a number can therefore replace the current input with a history entry.

Cache the NumLock state reported by GLFW key press events and expose it through ImHexApi::System as Unknown, On, or Off. Keypad 8/2 navigate history only when the state is Off; dedicated arrow keys and the existing input activation/history checks remain unchanged.

On Windows and Linux, read lock modifiers only when GLFW_LOCK_KEY_MODS is enabled. Enable this reporting for detached windows as well, using the existing callback chaining. Reset the cached state when the main window is initialized or shut down.

On macOS and Web, the state remains Unknown, disabling keypad history navigation while preserving character input and dedicated arrow key navigation.

### Additional things

- Built and manually tested on Windows; NumLock on/off behavior was confirmed.
- Linux, macOS, and Web have not been tested.
- git diff --check passed.
- The cache reflects the latest key press reported by GLFW; it is not a live OS query and does not match lock state individually to queued ImGui events.


paxcut:
The issue is not with the glfw key handling as Imgui correctly converts the key presses to either just detect the key press when num-lock is off or to also send the number to `io.InputQueueCharacters`  when num-lock is on. The following single change alone seems to fix the problem.
```cpp
bool upArrow     = ImGui::IsKeyPressed(ImGuiKey_UpArrow, false)       || 
                             (ImGui::IsKeyPressed(ImGuiKey_Keypad8, false)    &&   io.InputQueueCharacters.empty());

bool downArrow = ImGui::IsKeyPressed(ImGuiKey_DownArrow, false)  || 
                             (ImGui::IsKeyPressed(ImGuiKey_Keypad2, false)   &&   io.InputQueueCharacters.empty());
```
I added the fix to a feature branch that Im getting ready  to make public in the near future so it can be tested.

lyn-un:
Thanks for the suggestion! I tested this change on Windows, and it fixes the issue. This is much simpler than caching the NumLock state.

PR #2911, opened 2026-09-21
fix: Only highlight a byte red if its value actually differs from the original

scottanderson:
The Patches view highlighted an address if any write operation touched it, regardless of whether the write changed the byte's value. A no-op write, a multi-byte write that left some bytes unchanged, or two edits that netted out to the original value all painted the byte red.

Track the true original value per byte instead, recorded the first time it changes. The highlight now compares the live byte value against that original, so only actually-modified bytes turn red.


WerWolv:
Looking at my own code there, I think this should be rewritten entirely. This will cause a large amount of memory to be allocated when the user e.g does a big Fill operation. Currently with my code it's >8 bytes per modified byte, with yours it will probably be >8+16 bytes. 
I thing that needs to be improved in some way before we can consider improvements like this

PR #2912, opened 2026-09-22
Use a 64 bit counter in IntelHexProvider::readRaw

arpitjain099:
### Problem description

`IntelHexProvider::readRaw` counts with a `u32` while `interval.start`, `interval.end` and `offset` are all `u64`.

An ExtendedLinearAddress record of `0x0001` puts a region at 4 GiB, and `std::max(interval.start, offset)` truncates to 0. The `(i - offset) < size` guard still passes, so the body runs and reads `(*data)[0 - 0x100000000]`.

Three call sites read the whole provider in one go at offset 0 and would reach it: `tcp_client_server.cpp:87`, and `view_find.cpp:129` and `view_hex_editor.cpp:1325` after Select All. I read those rather than clicking through them.

The counter is now `u64`. Motorola SREC derives from this provider so it is covered by the same change.

I could not build the full GUI, so I took the loop as it stands into a small program against the real `wolv::container::IntervalTree`, with one region based at 4 GiB. It faults under ASan before the change. After it, offset 0 reads back as a hole, the region's bytes show up at `0x100000000`, and a region below 4 GiB is unaffected.


WerWolv:
Thank you!

PR #2916, opened 2026-09-23
fix: Sort Find results by value without decoding them

scottanderson:
A search sorts its results again when it completes. With the Value column as the sort key, each comparison decoded and escaped both whole results. A result that sorts late can be compared about N times, so one large string could freeze the UI for many seconds.

Read each result one time, up to its first 256 bytes, then sort on these bytes. A shorter result gets a shorter key. The table shows no more than 256 bytes, so the visible order does not change. Constant results still sort by their name.

Remove the default limit from `decodeValue()`. Each caller must now give its own limit.


plugins/builtin/source/content/views/view_find.cpp
```diff
@@ -798,6 +799,37 @@ namespace hex::plugin::builtin {
...
     }
 
+    void ViewFind::sortOccurrencesByValue(prv::Provider *provider, std::vector<FindOccurrence> &occurrences, bool ascending) const {
+        // A shorter value gets a shorter key. The table shows no more, so a longer key cannot change the visible order.
+        constexpr static size_t MaxKeySize = 256;
+
+        // Read each key once. A comparison then touches no provider and allocates nothing.
+        std::vector<std::vector<u8>> keys;
+        keys.reserve(occurrences.size());
+        for (const auto &occurrence : occurrences) {
```
WerWolv: Since there can be quite a lot of occurrences, moving the condition out of the for loop might make sense

plugins/builtin/source/content/views/view_find.cpp
```diff
@@ -798,6 +799,37 @@ namespace hex::plugin::builtin {
...
+            } else {
+                auto &key = keys.emplace_back(std::min<size_t>(occurrence.region.getSize(), MaxKeySize));
+                provider->read(occurrence.region.getStartAddress(), key.data(), key.size());
+            }
+        }
+
+        std::vector<size_t> order(occurrences.size());
+        std::iota(order.begin(), order.end(), 0);
+        std::ranges::stable_sort(order, [&keys, ascending](size_t left, size_t right) {
+            return ascending ? keys[left] < keys[right] : keys[right] < keys[left];
```
WerWolv: Same thing here, that's even more of a hot loop

neptuwunium:
This fixes the freeze I encountered I described in the discord, and overall makes it much faster than current non-nightly release. Sorting is still accurate.

neptuwunium:
There is one bug though that I don't know is related to the PR but I don't really consider it a blocker more a mild inconvenience. When finding strings and sorting them by Size, filtering them resets the sort order back to offset despite it still showing the sorting chevron under Size.

neptuwunium:
I just tested the behavior on nightly without this PR, it does seem to be a regression introduced by these changes.

plugins/builtin/source/content/views/view_find.cpp
```diff
@@ -798,6 +799,37 @@ namespace hex::plugin::builtin {
...
+        for (const auto &occurrence : occurrences) {
+            if (m_decodeSettings.mode == SearchSettings::Mode::Constants) {
+                keys.emplace_back(occurrence.string.begin(), occurrence.string.end());
+            } else {
+                auto &key = keys.emplace_back(std::min<size_t>(occurrence.region.getSize(), MaxKeySize));
+                provider->read(occurrence.region.getStartAddress(), key.data(), key.size());
+            }
+        }
+
+        std::vector<size_t> order(occurrences.size());
```
WerWolv: This whole thing feels like an extremely expensive way of writing `for(int i = 0; i < occurences.size(); i += 1)`

plugins/builtin/source/content/views/view_find.cpp
```diff
@@ -1233,8 +1265,13 @@ namespace hex::plugin::builtin {
                 sortSpecs->SpecsDirty = true;
             }
 
+            if (sortSpecs->SpecsDirty && sortSpecs->Specs->ColumnUserID == ImGui::GetID("value")) {
```
WerWolv: Checking SpecsDirty twice and then manually resetting it feels really weird, that should just be moved into the if below

WerWolv (changes requested):

neptuwunium:
> There is one bug though that I don't know is related to the PR but I don't really consider it a blocker more a mild inconvenience. When finding strings and sorting them by Size, filtering them resets the sort order back to offset despite it still showing the sorting chevron under Size.

This issue has been fixed in the recent commits.

